PR #21, opened 2024-06-08
Replace DirectSound with miniaudio

foxtacles:
Functional.

I've noticed a "crackling" sound artifact in a few cases when spatialization is enabled. Easiest way to reproduce this is to spawn at the hospital and wait for the ambulance animation to start. Not sure where that comes from - when 3D sound is disabled, the sound renders just fine (can test by returning `FALSE` from `MxOmni::IsSound3D`)

Here's a video with the sound artifact:

https://github.com/isledecomp/isle-portable/assets/1135351/4f2ba446-8a41-494e-af7e-52c94b73833f

@mackron do you have any idea what could cause this artifact? It's only present when spatialization is enabled on some of the sounds attached to the engine.

Tests to find any possible inconsistencies with the retail version are welcome.

The code that plays MIDI files (`MxMusicManager`, `MxMIDIPresenter`, `MxLoopingMIDIPresenter`, `MxMusicPresenter`) has been deleted since it is not used in the original game. There's no MIDI files in any of the `.SI` scripts as far as I know.


mackron:
I'm not entirely sure what's going on with that one. I noticed in the video that it happens when either the sound is moving or when the player is moving. It's also a high frequency glitch, almost as if it's happening every frame. I wonder if an update to sound or listener setting is causing it? As an experiment, what happens if you comment out the `ma_sound_set_pitch()` call?

foxtacles:
> I wonder if an update to sound or listener setting is causing it? As an experiment, what happens if you comment out the `ma_sound_set_pitch()` call?

Disabling `ma_sound_set_pitch` was actually the first thing I tried. It didn't make any difference, and turned out it wasn't even used for the sounds in this case in the first place.

I've also tried the following:

* Making the listener stationary (not using the `listener_set_*` functions, except in initial setup before the sounds start playing)
* Using miniaudio's default engine device (`ma_engine_init(NULL, &engine)`) instead of SDL for output
* Using miniaudio's high level `ma_sound_init_from_file` instead of the streaming mechanism with a ring buffer as seen in this PR

None of this seemed to make a difference. I've then went on and set up a small app to be able to reproduce the issue outside the game, which didn't work at first until I've set the volume of the siren sound to the same high value that is used in the game. Then I started hearing the artifact - sometimes, it doesn't seem to be consistent.

I've attached the app that reproduces the issue (for me) in this comment. Could you try and see if you can reproduce it on your end? I start hearing the glitch once the siren sound is relatively close to the listener near the end. (you may have to run it a few times - the intensity of the glitch seems to vary). Thank you!

[hospital.zip](https://github.com/user-attachments/files/15753007/hospital.zip)

mackron:
I haven't had a chance to run your test yet, but from your description one possibility is that it could be a clipping issue where some samples in the mix exceed the -1..1 range. In the mean time, what happens if you do `ma_engine_set_volume(&engine, 0.5f)` (and then just turn your speakers up to compensate)? If that works, my guess is that there's not enough headroom in one or both of your sounds. (When sounds are mixed each sample is summed together - if there's not enough headroom it's possible for that sum to exceed the -1..1 range and glitch.)

foxtacles:
Thank you for the suggestion! I've tried that and even though the overall volume goes down, if I turn up the speakers the glitch is still present/audible.

foxtacles:
Merging since everything seems to be functional in terms of audio except for the glitch; I've created a ticket for it here: https://github.com/isledecomp/isle-portable/issues/24

PR #25, opened 2024-06-22
Fix mingw build + copy d3drm from wine

madebr:
- Fixes the build so it builds and links with mingw again
- Copied the d3drm sources from wine's gitlab, using [this git revision](https://gitlab.winehq.org/wine/wine/-/tree/6c5d17af07a318d754c0c21023b2d162a0d3725d/dlls/d3drm) When using a 32-bit mingw, it builds and links. (I didn't try to run as I don't have the data setup at the moment)


foxtacles:
Nice. Could you add a MinGW build using the wine d3drm to the GitHub build action? Thank you!

tools/ncc/skip.yml
```diff
@@ -30,3 +30,4 @@ i_activity: "Allow original naming from beta"
 i_actor: "Allow original naming from beta"
 score: "Allow original naming from beta"
 c_LOCATIONS_NUM: "Allow original naming from beta"
+SDLCALL: "SDL function call convention"
```
foxtacles, line 33: This should go to the --definition parameter of the ncc invocation instead like the others such as WINAPI

ISLE/isleapp.h
```diff
@@ -1,15 +1,14 @@
 #ifndef ISLEAPP_H
 #define ISLEAPP_H
 
+#include "lego/legoomni/include/legoutils.h"
```
foxtacles, line 4: I think it should be possible to directly include legoutils.h without the path presented?

madebr: Yes, it's not needed.
I don't know why it did not work when I first tried to fix this.

PR #26, opened 2024-06-25
Fix pointer issues and build with mingw64

madebr:


foxtacles:
Looks good! As far as I see it should be possible to apply many of the changes to `LEGO1` to the decomp repository as well without affecting accuracy. Can you create a PR with those changes over there so we can merge them? Then I will sync/copy it to `isle-portable`.  Since the codebase of this repository is based on the decomp I'd like to keep the divergence to a minimum. Thanks!

madebr:
I'll look into it.

Looks like this repo has no common history with the isle repo.
Wouldn't it make more sense to make it a proper fork with full history?
I think it will make tracking changes with the original repo much easier.

foxtacles:
Yeah, maybe it would make more sense. I wasn't sure if it would be "the right" thing to do since no changes are meant to be merged back into "upstream" from here, and not all commits are meant to be synced into `isle-portable` (for instance, everything related to the decomp tools has been removed). So the histories will naturally diverge to a certain extent.

What I've been doing so far is use `git format-patch --stdout` and `git am` to extract and then selectively apply the commits, but I'm open to changing this.

foxtacles:
@madebr updated master. There was one minor compilation error with regards to the app data type in the MinGW build, but we can fix it here

3rdparty/d3drm/CMakeLists.txt
```diff
@@ -2,7 +2,33 @@ project(wine_d3drm LANGUAGES C)
...
-add_library(d3drm-wine SHARED EXCLUDE_FROM_ALL
+if(MSVC)
+    # FIXME: this is wrong.
+    if(CMAKE_SIZEOF_VOID_P EQUAL 4)
+        set(def "${CMAKE_CURRENT_SOURCE_DIR}/d3dxof32.def")
+    else()
+        set(def "${CMAKE_CURRENT_SOURCE_DIR}/d3dxof64.def")
+    endif()
+    set(imp "${CMAKE_CURRENT_BINARY_DIR}/d3dxof.lib")
+    set(exp "${CMAKE_CURRENT_BINARY_DIR}/d3dxof.exp")
```
madebr, line 13: I was able to get everything building and linking with MSVC targeting x64.
There is now an issue with d3drm's dependency on d3dxof: does a 64-bit d3dxof.dll even exist?

[DirectXFileCreate](https://learn.microsoft.com/en-us/windows/win32/direct3d9/directxfilecreate)

Perhaps we need to copy its code from wine as well.

madebr: I think the (initial) easiest solution would be to load `DirectXFileCreate` dynamically (using `LoadLibrary` + `GetProcAddress`).
We might not even need `d3dxof.dll` library.

madebr: Never mind, looks like it is used to deserialize data.

madebr: I made this an option so you can try by configuring with `-DWINE_D3DRM_DYNAMIC_D3DXOF=ON/OFF`.

PR #30, opened 2024-06-26
Parse cli arguments + log with SDL_Log

madebr:
- copy DLL dependencies to binary dir. This will also copy a system SDL3.dll to the build directory (when configuring cmake with `-DDOWNLOAD_DEPENDENCIES=OFF`)
- Only include `<SDL3/SDL.h>`. Including every header for each feature is a bit overkill (imho). All SDL symbols have a `SDL_` prefix, so there's little change of introducing a conflict.
- Parse arguments: this allows a custom ini path, instead of the global one:`--ini <ini-path>`
- Create a SDL_window using SDL properties. Imho not having if-branch makes it more clearer what is happening.
- Log errors/warnings with SDL_Log, and check for `E_INTERFACE` as well (I always get this error when trying to get a `IDirect3D2` object)
- Simplify d3drm3d.def, it should work with mingw32/mingw64/msvc x32/msvc x64
- Ignore a few assertions that are always failing with wine's d3drm (they always return `E_NOTIMPL`)
- Fix iniparser's conflict between static and shared lib names (with MSVC) (https://gitlab.com/iniparser/iniparser/-/issues/168)


madebr:
Btw, wine d3rm does not result in a working game.

I think a way forward is to document how d3rm is used and what features are enabled/disabled.
Lego Island's assets are not that complicated (not much vertexs), so I think we don't need to worry much about "advanced" techniques such quadtrees, efficient backface culling, ... etc. Those things can come later if really needed.

foxtacles:
> Btw, wine d3rm does not result in a working game.
> 
> I think a way forward is to document how d3rm is used and what features are enabled/disabled. Lego Island's assets are not that complicated (not much vertexs), so I think we don't need to worry much about "advanced" techniques such quadtrees, efficient backface culling, ... etc. Those things can come later if really needed.

Fully agreed.

Unfortunately some core functionality like the Update/Render methods are not implemented in wine d3drm, so naturally it's not going to work for now, but I think what they have is a good starting point.

PR #41, opened 2024-12-19
replace directinput joystick APIs with SDL

Ramen2X:
Functional.

I've renamed `GetJoystickId()` to `GetJoystick()` since we don't always globally grab the handle to the joystick from its id anymore like we did with DirectInput. Instead, we use the joystick id once to grab and store the handle to the joystick in an `SDL_Joystick` object, which I have made a member of LegoInputManager named `m_joystick`.

I've removed `buttonState` since it is effectively unused. LEGO Island does not support any joystick buttons *except* for hat button input for POV switching which gets handled in a separate manner.

The reason for the switch statement in `GetJoystickState()` is because SDL's definitions for joystick hat values do not correspond to those in DirectInput, so conversion is necessary to keep the POV switching identical to how it was in the original game. If you have a better or cleaner way of doing this, by all means let me know.

Since SDL's API is vastly superior to DirectInput, this enables hotswapping of joysticks:

https://github.com/user-attachments/assets/82e9c1c8-197a-45b0-95b2-78a806457e7d

Just like the original, the game does not accept any joystick input if "Use Joystick" is set to `FALSE`, and the keyboard remains available even if "Use Joystick" is set to `TRUE`.

### DIFF
--- a/LEGO1/lego/legoomni/include/legoinputmanager.h
+++ b/LEGO1/lego/legoomni/include/legoinputmanager.h
@@ -9,7 +9,7 @@
 
 #include <SDL3/SDL_joystick.h>
 #include <SDL3/SDL_keyboard.h>
-#include <dinput.h>
+#include <windows.h>
 
 class LegoCameraController;
 class LegoControlManager;
--- a/LEGO1/lego/legoomni/include/legoinputmanager.h
+++ b/LEGO1/lego/legoomni/include/legoinputmanager.h
@@ -7,6 +7,7 @@
 #include "mxpresenter.h"
 #include "mxqueue.h"
 
+#include <SDL3/SDL_joystick.h>
 #include <SDL3/SDL_keyboard.h>
 #include <dinput.h>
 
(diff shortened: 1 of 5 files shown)


LEGO1/lego/legoomni/include/legoinputmanager.h
```diff
@@ -7,6 +7,7 @@
 #include "mxpresenter.h"
 #include "mxqueue.h"
 
+#include <SDL3/SDL_joystick.h>
 #include <SDL3/SDL_keyboard.h>
 #include <dinput.h>
```
foxtacles, line 12: Maybe we can remove this Windows header now?

foxtacles (commented):
Looks good, thank you! Just one comment

LEGO1/lego/legoomni/include/legoinputmanager.h
Ramen2X: I had the same thought, but it unfortunately seems to fail due to `MxResult Create()` taking an `HWND` as a parameter. Assuming this is inherited recursively, we may be able to replace `dinput.h` with `windows.h` here since we aren't using any DirectInput-specific code anymore.

foxtacles: Yes. Let's replace it with `windows.h` then!

Ramen2X: Hadn't seen your edit until now for some reason, glad we had the same idea!

foxtacles (approved):

PR #44, opened 2024-12-23
Use SDL3 functions in more locations

madebr:
This slightly reduces the number of Windows-only apis.

Also untested.

I refactored the CMake script a bit because of [this Windows behavior](https://devblogs.microsoft.com/oldnewthing/20140321-00/?p=1433). Some symbols were not exported from the dll.


foxtacles:
Looks good! I'll test it tomorrow but since all the SDL functions are API-compatible it should be fine.

madebr:
> Looks good! I'll test it tomorrow but since all the SDL functions are API-compatible it should be fine.

I think the replacement of `timeGetTIme` with `SDL_Delay` is the biggest change. Both return times in ms.

LEGO1/lego/legoomni/src/build/legocarbuild.cpp
```diff
@@ -559,7 +561,7 @@ MxResult LegoCarBuild::Tickle()
 	}
 
 	if (m_unk0x10a) {
-		DWORD time = timeGetTime();
+		Uint64 time = SDL_GetTicks();
 		DWORD dTime = (time - m_unk0x10c) / 100;
```
foxtacles, line 565: I think it would be good if `DWORD` also becomes `Uint64` here.

foxtacles (approved):
Works as expected! Just added one comment. The fact that we now get the time since SDL initialized vs. time since Windows started is, as expected, not relevant.

LEGO1/lego/legoomni/src/build/legocarbuild.cpp
madebr: I hope the most significant bits are never used. If they are, then we have a big problem :)

```suggestion
		Uint64 dTime = (time - m_unk0x10c) / 100;
```

PR #47, opened 2024-12-27
More use of SDL

madebr:
- pass SDL_Window into lego dll, and get HWND pointer where it is needed
-  use SDL_GetPerformanceCounter instead of win32's QueryPerformanceCounter
- fix flic structs for linux gcc
- remove left-over print debug statement in cmake script
- more s/stricmp/SDL_strcasecmp/
- Use SDL_Timer instead of win32's SetTimer


foxtacles:
Looks good and everything seems to work fine on Windows as well!

PR #49, opened 2024-12-28
Even more SDL usage

madebr:
- Make LegoImage a thin wrapper around SDL_Surface and  SDL_Palette
- Use SDL_Process to run subprocesses
- pass SDL_Window handle to `SDL_ShowSimpleMessageBox`. When Lego Island fails to start, this makes sure the dialog is shown on top and not hidden behind the full screen.
- I sprinkled `SDL_LogError` around the initialization so issues can be found in the debug stream, or on stderr
- replace `_access` with `SDL_GetPathInfo`
- replace `DeleteFile` with `SDL_RemovePath`
- replace `MoveFile` with `SDL_RenamePath`

I've got to say it's good we're using SDL3, because the process and filesystem API are new features.

As usual, please test this pr on Windows.

### DIFF
--- a/LEGO1/lego/legoomni/include/legotextureinfo.h
+++ b/LEGO1/lego/legoomni/include/legotextureinfo.h
@@ -19,7 +19,7 @@ public:
 	static BOOL SetGroupTexture(Tgl::Mesh* pMesh, LegoTextureInfo* p_textureInfo);
 	static BOOL GetGroupTexture(Tgl::Mesh* pMesh, LegoTextureInfo*& p_textureInfo);
 
-	LegoResult FUN_10066010(LegoU8* p_bits);
+	LegoResult FUN_10066010(const LegoU8* p_bits);
 
 	// private:
 	char* m_name;                   // 0x00
--- a/LEGO1/lego/legoomni/include/pizza.h
+++ b/LEGO1/lego/legoomni/include/pizza.h
@@ -6,7 +6,7 @@
 #include "isleactor.h"
 #include "legostate.h"
 
-#include <climits>
+#include <limits.h>
 
 class Act1State;
 class PizzeriaState;
--- a/LEGO1/lego/legoomni/include/pizza.h
+++ b/LEGO1/lego/legoomni/include/pizza.h
@@ -6,6 +6,8 @@
 #include "isleactor.h"
 #include "legostate.h"
 
+#include <climits>
+
 class Act1State;
 class PizzeriaState;
 class SkateBoard;
--- a/LEGO1/lego/legoomni/src/common/legoutils.cpp
+++ b/LEGO1/lego/legoomni/src/common/legoutils.cpp
@@ -729,7 +729,7 @@ void WriteDefaultTexture(LegoFile* p_file, const char* p_name)
 				}
 				else {
 					MxU8* surface = (MxU8*) desc.lpSurface;
-					LegoU8* bits = image->GetBits();
+					const LegoU8* bits = image->GetBits();
 
 					for (MxS32 i = 0; i < desc.dwHeight; i++) {
 						memcpy(surface, bits, desc.dwWidth);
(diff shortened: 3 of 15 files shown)


LEGO1/lego/legoomni/include/legotextureinfo.h
```diff
@@ -19,7 +19,7 @@ class LegoTextureInfo {
 	static BOOL SetGroupTexture(Tgl::Mesh* pMesh, LegoTextureInfo* p_textureInfo);
 	static BOOL GetGroupTexture(Tgl::Mesh* pMesh, LegoTextureInfo*& p_textureInfo);
 
-	LegoResult FUN_10066010(LegoU8* p_bits);
+	LegoResult FUN_10066010(const LegoU8* p_bits);
```
foxtacles, line 22: I think this might be a change we can add to the decomp repo?

LEGO1/lego/legoomni/include/pizza.h
```diff
@@ -6,6 +6,8 @@
 #include "isleactor.h"
 #include "legostate.h"
 
+#include <climits>
```
foxtacles, line 9: What is this header inclusion needed for? If we really need it we should make it limits.h instead of climate since that's what has been used so far.
(I'd certainly prefer the C++ versions in a new project but for consistency we shouldn't start using them)

LEGO1/lego/legoomni/src/common/legoutils.cpp
```diff
@@ -753,12 +755,14 @@ void WriteDefaultTexture(LegoFile* p_file, const char* p_name)
 						paletteEntries[i].SetBlue(entries[i].peBlue);
 					}
 
-					image->SetCount(i);
+					SDL_Palette* newPalette = SDL_CreatePalette(i);
 
 					if (i > 0) {
-						// Note: this appears to be a bug. size should be i * sizeof(LegoPaletteEntry)
-						memcpy(image->GetPalette(), paletteEntries, i);
+						for (int j = 0; j < i; j++) {
```
foxtacles, line 761: Should use MxS32 instead of int here

foxtacles (commented):
Looks good, just a few small things. Will test tomorrow - thank you!

LEGO1/lego/legoomni/include/pizza.h
madebr: It was for a `INT_MIN` which is defined by windows.h, but not when targeting Linux.

LEGO1/lego/legoomni/include/legotextureinfo.h
madebr: Yup, I think so.

LEGO1/lego/legoomni/src/common/legoutils.cpp
```diff
@@ -320,9 +320,11 @@ void InvokeAction(Extra::ActionType p_actionId, const MxAtomId& p_pAtom, MxS32 p
 		}
 
 		break;
-	case Extra::ActionType::e_run:
-		_spawnl(0, "\\lego\\sources\\main\\main.exe", "\\lego\\sources\\main\\main.exe", "/script", &p_pAtom, 0);
-		break;
+	case Extra::ActionType::e_run: {
+		const char* args[] =
+			{"/lego/sources/main/main.exe", "/lego/sources/main/main.exe", "/script", p_pAtom.GetInternal(), NULL};
```
madebr, line 325:
```suggestion
			{"/lego/sources/main/main.exe", "/script", p_pAtom.GetInternal(), NULL};
```

LEGO1/lego/legoomni/include/pizza.h
madebr: There are more similar occurrence that don't:t show up because of my incomplete windows.h wrapper.
Those will be addressed in time.

PR #51, opened 2024-12-29
Use SDL_KeyCode in LEGO1.dll

madebr:
I think this change deserves its own pr for testing purposes.

- Also use iniparser for getting mediaPath
- and replace a single isdigit with SDL_isdigit

### DIFF
--- a/ISLE/isleapp.cpp
+++ b/ISLE/isleapp.cpp
@@ -140,6 +140,10 @@ IsleApp::~IsleApp()
 	if (m_savePath) {
 		delete[] m_savePath;
 	}
+
+	if (m_mediaPath) {
+		delete[] m_mediaPath;
+	}
 }
 
 // FUNCTION: ISLE 0x401260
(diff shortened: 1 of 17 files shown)


ISLE/isleapp.cpp
```diff
@@ -544,6 +550,10 @@ void IsleApp::LoadConfig()
 	strcpy(m_cdPath, cdPath);
 	MxOmni::SetCD(m_cdPath);
 
+	const char* mediaPath = iniparser_getstring(dict, "isle:mediapath", cdPath);
```
foxtacles, line 553: I think a better default would be `hdPath` instead of `cdPath`

foxtacles:
In the registration book, we have an access violation when trying to type a name:

image

Probably just need to replace the `- 'A'` part

foxtacles:
The debug triggers behind the "ogel" password don't work right now - I think you have to replace the `g_debugPassword` with an array (null terminated) of the corresponding SDL key codes.

madebr:
> The debug triggers behind the "ogel" password don't work right now - I think you have to replace the `g_debugPassword` with an array (null terminated) of the corresponding SDL key codes.

Indeed. Long term, we should use a combination of SDL_Keycode and SDL_Scancode: SDL_Scancode for WASD and movement, and keycode for other cheats and/or text input.

foxtacles:
Everything works now, thank you!

PR #64, opened 2025-05-09
fix: update libiniparser repository link

OtterCodes101:
The libiniparser maintainers have appeared to move to a GitLab repository. I have updated the link accordingly.


foxtacles:
We should update the git dependency in `CMakeLists.txt` as well

OtterCodes101:
the old repository was partially reintroduced in e92805e

PR #68, opened 2025-05-11
Fix definition of WriteRegisterSettings()

AJenbo:
I couldn't find a way to get this working so hoping this is the correct fix.


foxtacles:
Thank you - can you contribute this to the decomp repo? We regularly merge updates from over there into portable, so it will land here too.

PR #69, opened 2025-05-11
Make things build on Linux

AJenbo:
This PR implements enough of the Windows APIs that it's possible to build Lego Island on Linux (x86_64).

```bash
cmake -S . -B build
cmake --build build -j 32
```

The current implementations are just empty stubs. They mostly assume everything went well, unless asked to create something.

This is enough that the application is able to boot, read the config, create a window (full screen or windowed), and then detect that "DirectDraw" can't find a suitable device to run no-ops on.

image

The configuration application also builds, though it fails much earlier since it doesn't have any code for creating a window and just gives up after that.

How ever both applications shut down nicely so there are no noticeable bugs just missing features :)

The next step would probably be for someone to go in and implement some of the most obviously missing functions and add `assert(false)` to all renaming functions, then iterate until everything works.

To help with the process I also added ASAN and a CI build script. Have fun building :)

P.s. this in no way aims to be a proper implementation of any of the APIs and instead is just what is needed to be source compatible from Lego Islands perspective, lots of enum values and relations are just made up on the fly and not the values used in the real APIs but I considered that out of scope.


3rdparty/CMakeLists.txt
```diff
@@ -45,5 +45,6 @@ endif()
 add_library(libsmacker STATIC
     ${libsmacker_SOURCE_DIR}/smacker.c
 )
+target_compile_options(libsmacker PRIVATE -fPIC)
```
madebr, line 48:
```suggestion
set_property(TARGET libsmacker PROPERTY POSITION_INDEPENDENT_CODE ON)
```

CMakeLists.txt
```diff
@@ -52,14 +52,21 @@ include(CheckCXXSourceCompiles)
...
 endif()
 
+if (NOT CMAKE_BUILD_TYPE STREQUAL "Release")
+  if(CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU")
+      add_compile_options(-fsanitize=address -fno-omit-frame-pointer)
+      add_link_options(-fsanitize=address)
+  endif()
+endif()
+
 function(add_cxx_warning WARNING)
```
madebr, line 70:
```suggestion
endif()

function(add_cxx_warning WARNING)
```

Users should configure with `-DCMAKE_CXX_FLAGS=-fsanitize=address -DCMAKE_EXE_LINKER_FLAGS=-fsantitize=address -DCMAKE_SHARED_LINKER_FLAGS=-fsanitize=address`.

Or we could make it available as an extra option, e.g. `iSLE_ASAN`.

Building a release binary with sanitizing is always useful.

CMakeLists.txt
```diff
@@ -124,6 +131,7 @@ macro(add_lego1_static_library NAME)
   target_link_libraries(${NAME}-interface INTERFACE $<$<BOOL:${ISLE_USE_DX5}>:DirectX5::DirectX5> SDL3::SDL3)
   target_compile_definitions(${NAME}-objects PRIVATE $<$<BOOL:${ISLE_USE_DX5}>:DIRECTX5_SDK>)
   target_compile_definitions(${NAME}-objects PRIVATE $<$<BOOL:${ISLE_D3DRM_FROM_WINE}>:D3DRM_WINE>)
+  target_compile_options(${NAME}-objects PRIVATE -fPIC)
```
madebr, line 134:
```suggestion
  set_propery(TARGET ${NAME}-objects PROPERTY POSITION_INDEPENDENT_CODE ON)
```

CMakeLists.txt
```diff
@@ -426,32 +442,49 @@ if(MINGW)
...
+  add_library(lego1 SHARED
+    LEGO1/main.cpp
+    ${lego1_objects}
+  )
+else()
+  add_library(lego1 STATIC
+    LEGO1/main.cpp
+    ${lego1_objects}
+  )
+endif()
```
madebr, line 464: Id keep this `SHARED` for now, and turn everything into a single executable at a later point.
Since the goal of this repo is no longer to be accurate, the cmake script can be simplified greatly.
```suggestion
add_library(lego1 SHARED
  LEGO1/main.cpp
  ${lego1_objects}
)
```

CMakeLists.txt
```diff
@@ -426,32 +442,49 @@ if(MINGW)
...
 # Make sure filenames are ALL CAPS
 set_property(TARGET lego1 PROPERTY OUTPUT_NAME LEGO1)
 set_property(TARGET lego1 PROPERTY PREFIX "")
-set_property(TARGET lego1 PROPERTY SUFFIX ".DLL")
+if(WIN32)
+  set_property(TARGET lego1 PROPERTY SUFFIX ".DLL")
+else()
+  set_property(TARGET lego1 PROPERTY SUFFIX ".so")
+  target_compile_options(lego1 PRIVATE -fPIC)
+endif()
```
madebr, line 475: macOS wants `.dylib` :)
So let's only change Windows.

```suggestion
set_property(TARGET lego1 PROPERTY SUFFIX POSITION_INDEPENDENT_CODE TRUE)
if(WIN32)
  set_property(TARGET lego1 PROPERTY SUFFIX ".DLL")
endif()
```

But again, this can be removed in later work since we don't care about accuracy in this repo.
What's important in this repo is to keep it easy to merge with the other isle repo.

CMakeLists.txt
madebr: This might also be the reason for the link failure on ci.

AJenbo: It's needed for ASAN

CMakeLists.txt
AJenbo: Oh is it just there for .dll vs .DLL?

madebr: Yeah, this is a silly thing.

CMakeLists.txt
AJenbo: I need ASAN somehow so iSLE_ASAN it is.

.github/workflows/linux.yml
```diff
@@ -0,0 +1,45 @@
...
+            -DISLE_WERROR=ON \
+            -Werror=dev
+
+      - name: Build
+        run: cmake --build build -- -k0
+
+      - name: Make Artifact Archive
+        run: |
+          cd build
+          zip "isle-portable-linux-${{ matrix.build-type }}.zip" CONFIG ISLE  LEGO1.so
```
AJenbo, line 39:
```suggestion
          zip "isle-portable-linux-${{ matrix.build-type }}.zip" CONFIG ISLE LEGO1.so
```

CMakeLists.txt
```diff
@@ -461,38 +486,52 @@ if (ISLE_BUILD_APP)
...
     LEGO1/mxdirectx/mxdirectxinfo.cpp
     LEGO1/mxdirectx/legodxinfo.cpp
     CONFIG/config.cpp
     CONFIG/ConfigCommandLineInfo.cpp
     CONFIG/AboutDlg.cpp
     CONFIG/MainDlg.cpp
     CONFIG/detectdx5.cpp
-    CONFIG/StdAfx.cpp
     CONFIG/res/config.rc
+    CONFIG/StdAfx.cpp
```
AJenbo, line 515:
```suggestion
    CONFIG/StdAfx.cpp
    CONFIG/res/config.rc
```

LEGO1/omni/src/video/mxdisplaysurface.cpp
```diff
@@ -1397,7 +1401,7 @@ void MxDisplaySurface::VTable0x2c(
 }
 
 // FUNCTION: LEGO1 0x100bc8b0
-LPDIRECTDRAWSURFACE MxDisplaySurface::FUN_100bc8b0(MxS32 width, MxS32 height)
+LPDIRECTDRAWSURFACE MxDisplaySurface::FUN_100bc8b0(MxS32 p_width, MxS32 p_height)
```
foxtacles, line 1404: Can you submit this change to the decomp repo as well? Thank you

.github/workflows/linux.yml
```diff
@@ -0,0 +1,46 @@
+name: Build (Linux)
```
foxtacles, line 1: For consistency, maybe we should rename build.yml to windows.yml

LEGO1/omni/src/video/mxdisplaysurface.cpp
AJenbo: Only did it to have the build not fail ...

foxtacles: Yes, I figured. I'm not sure why the build isn't failing in the decomp. But anything that isn't related to the porting effort we should ideally "back port" to the decomp, to make merging updates easier

AJenbo: https://github.com/isledecomp/isle/pull/1472

CMakeLists.txt
```diff
@@ -14,7 +17,9 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}" CACHE PATH "Directory w
 
 # By configuring CMake with -DDOWNLOAD_DEPENDENCIES=ON/OFF,
 # users can choose between downloading dependencies or using system libraries
-option(DOWNLOAD_DEPENDENCIES "Download dependencies" TRUE)
+option(DOWNLOAD_DEPENDENCIES "Download dependencies" ON)
+
+option(iSLE_ASAN "Enable AddressSanitizer" OFF)
```
AllMeatball, line 22: The capitalization on this line is messed up it should be `ISLE_ASAN` instead.

AJenbo: I would agree, but that was the way @madebr did it and I didn't want to try and be clever :D

madebr: lol bring all iThings

This was a typo on my part, sorry :)

foxtacles (approved):
Again, great work! Everything works as expected on my end on both Windows and Linux. Thanks a lot for this, it allows us to iteratively port piece by piece going forward.

foxtacles:
I've pulled updates from the decomp into `master` - could you rebase or update from `master`? Thank you!

Ramen2X (approved):

madebr:
I'd also make everything non-inline, just to avoid full recompilations.

AJenbo:
> I'd also make everything non-inline, just to avoid full recompilations.

Yeah it's a deliberate choice to have it that way atm, but it would be terrible for a real implementation.

AJenbo:
> I've pulled updates from the decomp into `master` - could you rebase or update from `master`? Thank you!

Done, that also got rid of all the ASCII defines

PR #71, opened 2025-05-12
Create config with sane defaults if not already present

AllMeatball:


ISLE/isleapp.cpp
```diff
@@ -540,6 +545,45 @@ void IsleApp::LoadConfig()
...
+		iniparser_set(dict, "isle:JoystickIndex", m_joystickIndex  ? "true" : "false");
+		iniparser_set(dict, "isle:Draw Cursor",   m_drawCursor  ? "true" : "false");
+
+		iniparser_set(dict, "isle:Back Buffers in Video RAM", "-1");
+
+		iniparser_set(dict, "isle:Island Quality", "1");
+		iniparser_set(dict, "isle:Island Texture", "1");
+
+		iniparser_dump_ini(dict, iniFP);
+		SDL_LogInfo(SDL_LOG_CATEGORY_APPLICATION, "New config written at '%s'", iniConfig);
```
madebr, line 584:
```suggestion
		SDL_LogInfo(SDL_LOG_CATEGORY_APPLICATION, "New config written at '%s'", iniConfig);
		fclose(iniFP);
```

ISLE/isleapp.cpp
```diff
@@ -540,6 +545,45 @@ void IsleApp::LoadConfig()
...
 
+	// [library:config]
+	// Load sane defaults if dictionary failed to load
+	if (!dict) {
+		SDL_LogInfo(SDL_LOG_CATEGORY_APPLICATION, "Loading sane defaults");
+		FILE* iniFP = fopen(iniConfig, "w");
+
+		if (!iniFP) {
+			SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "Failed to write config at '%s': %s", iniConfig, strerror(errno));
+			fclose(iniFP);
```
madebr, line 556: Avoid a null pointer dereference.
```suggestion
			return;
```

ISLE/isleapp.cpp
```diff
@@ -540,6 +545,45 @@ void IsleApp::LoadConfig()
 
 	dictionary* dict = iniparser_load(iniConfig);
 
+	// [library:config]
+	// Load sane defaults if dictionary failed to load
+	if (!dict) {
+		SDL_LogInfo(SDL_LOG_CATEGORY_APPLICATION, "Loading sane defaults");
+		FILE* iniFP = fopen(iniConfig, "w");
```
madebr, line 552: Do we really need `\r\n` on Windows? Binary mode is more reliable.
```suggestion
		FILE* iniFP = fopen(iniConfig, "wb");
```

ISLE/isleapp.cpp
foxtacles, line 556: No need for `fclose` if `iniFP` is `NULL`.

Also, attempting to write to the file shouldn't happen past this point.

ISLE/isleapp.cpp
```diff
@@ -35,6 +35,11 @@
 #include <iniparser.h>
 #include <time.h>
 
+extern "C" {
```
foxtacles, line 38: I don't think this linkage spec is necessary?

ISLE/isleapp.cpp
AllMeatball: Would this still load the HD and CD paths?

AllMeatball: I might go back to my previous method then.

Which would be creating a blank dictionary with `dictionary_create(0)`

then write to a file afterwards. with `iniparser_dump_ini`

ISLE/isleapp.cpp
```diff
@@ -594,9 +639,12 @@ void IsleApp::LoadConfig()
 	m_savePath = new char[strlen(savePath) + 1];
 	strcpy(m_savePath, savePath);
 
+	fclose(iniFP);
```
madebr, line 642: This cannot compile: `iniFP` is not declared in this scope.

AllMeatball:
Does this look better. (I apologize for code that doesn't work. Linux support isn't committed yet and it's a bit confusing to hop around two similar repos fixing code.)

foxtacles (approved):

ISLE/isleapp.cpp
```diff
@@ -540,6 +545,51 @@ void IsleApp::LoadConfig()
 
 	dictionary* dict = iniparser_load(iniConfig);
 
+	// [library:config]
```
madebr, line 548: The logic does not seem correct to me.
When you run `isle --ini path/to/nonexisting/config.ini`, I'd expect the app to log an error that the ini you passed does not exist instead of creating a new one from scratch.

foxtacles: That makes sense. I forgot we have the case where we explicitly specify a file on the command line. I agree in that case it should error out when it's missing. When no file is specified or found in a default path, only then should a file be created.

madebr (approved):

AJenbo:
Nice this makes things easier to work on

PR #72, opened 2025-05-12
Refactor SIMD Capability Check

AJenbo:
This change addresses an issue where the current SIMD (MMX) check fails on x86_64 Linux builds using GCC.

- Renamed the function to better reflect its purpose: MMX is an x86-specific brand of SIMD. The check is now for general SIMD support.
-  On x86_64, all CPUs support at least SSE2, so runtime checks are unnecessary there.
- Replaced inline assembly with `__cpuid()` where available to simplify the check.
- Added support for ARM NEON, which is a roughly equivalent SIMD extension for 32-bit ARM, commonly found on Android.

While this check may have originally been intended to exclude very low-end x86 machines, it is not actually required to run the game. Modern platforms universally support SIMD extensions, and systems that don't are vanishingly rare. It's also unclear whether an ARM CPU lacking NEON support would have any real issue running the game, if clocked high enough, it might perform adequately regardless.

## Notes on Legacy Compatibility

- Modern MSVC versions no longer support Windows 9x. To build for these older systems using MSVC, you would need to downgrade to a version such as 6.0 or 2003, which would also require downgrading CMake from 3.25 to 2.8.
- You can still target Windows 9x with GCC (tested up to GCC 10), but its CRTs are often pre-built assuming Pentium Pro+. As a result, binaries will crash with illegal instruction errors during early startup on CPUs that lack MMX, before the SIMD check can even be reached.
- The project also uses SDL3, which is incompatible with older versions of Windows. These systems would require SDL 1.2 instead.

## Final Thoughts

If someone is going through the effort of running this port on truly vintage hardware (e.g., pre-MMX CPUs), they're likely already aware of the performance and compatibility limitations. They're far less likely to complain than users encountering false positives from a broken or overly cautious SIMD check. Given that, I think there's a strong case for removing the check entirely, rather than maintaining fragile special-case logic for extremely niche setups, especially since properly testing it is currently impossible.

### DIFF
--- a/LEGO1/mxdirectx/legodxinfo.cpp
+++ b/LEGO1/mxdirectx/legodxinfo.cpp
@@ -171,7 +171,7 @@ int LegoDeviceEnumerate::FUN_1009d0d0()
 	int i = 0;
 	int j = 0;
 	int k = -1;
-	int cpu_mmx = SupportsMMX();
+	bool cpu_mmx = SupportsSIMD();
 
 	for (list<MxDriver>::iterator it = m_list.begin(); it != m_list.end(); it++, i++) {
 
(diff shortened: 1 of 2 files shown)


foxtacles:
> Given this, there's a strong argument to remove this check entirely rather than continue maintaining a special-case implementation for niche setups.

I'd probably remove it entirely. I'd be surprised if anything of `legodxinfo.cpp` is left at all after a full port. If this check causes problems with today's builds, I think it's fair to remove it outright (although I do appreciate you trying to make it compatible).

Happy to hear everyone else's thoughts.

madebr (approved):

PR #73, opened 2025-05-12
miniwin: Init graphics API

AJenbo:
This implements enough of DirectDraw and Direct3D that the game can now complete the initialization and hardware detection. To the end user there isn't much change but the debug output now looks like this:

```
Reading configuration from "/home/ajenbo/.local/share/isledecomp/isle/isle.ini"
ERROR: MxDirectDraw error: [0x88760001] CreateSurface for window front buffer failed
ERROR: MxDirect3D::Create failed
ERROR: Failed to create LegoVideoManager
```

Which is a big step up from:
```
Reading configuration from "/home/ajenbo/.local/share/isledecomp/isle/isle.ini"
ERROR: DirectDrawEnumerate returned error Generic failure.
ERROR: LegoDeviceEnumerate::DoEnumerate failed
ERROR: Failed to create LegoVideoManager
```
... at least I think so.

It also moves miniwin in to it's own folder and lib to lay some ground rules for how to start implementing things.

I corrected an enum value and a define which also helped Isle pick the hardware render instead of the software renderer (it thought the hardware wasn't able to output and image).

Note: QueryInterface() is implemented in a way where the user is responsible for memory management. This is the opposite of of how the real one works and so right now this is causing memory leakage.

A lot here is still fairly hard coded, but without knowing what API will be picked for the actual 3D rendering it's not clear how to list the capable devices or how to get there capabilities so I think this is makes sens for getting us to the point where someone can start working on that and then figure out how to expose the actual hardware capabilities. It's also useful as a reference for what the game will accept as a valid renderer.


foxtacles (approved):

madebr (approved):

madebr:
Can you pull miniwin in 2 pieces? miniwin and minid3drm?

I'm asking for MSVC x64 and arm64: these platforms are windows, but don't have d3drm support.

madebr:
Also, long term, it would make sense to switch all platforms to the minid3drm implementation, even MSVC x86.

AJenbo:
> Can you pull miniwin in 2 pieces? miniwin and minid3drm?
> 
> I'm asking for MSVC x64 and arm64: these platforms are windows, but don't have d3drm support.

That's doable but I don't have Windows and none of the D3DRM is implemented only a single header is provided which you can't make a lib from. So I would like to leave it up to someone else.

madebr:
> 
> That's doable but I don't have Windows and none of the D3DRM is implemented only a single header is provided which you can't make a lib from. So I would like to leave it up to someone else.

No problem. I'll do that. I need that for removing the d3drm-wine files.

AJenbo:
If you intent for it to allow the game to talk to a real d3drm.dll you will need to get the values properties and order right in the header, even the unused parts of the classes/enums will need to be in the header for them to work that way (you won't need any more enums or classes then I added, they just need to match the real ones structure to also be binary compatible).

PR #76, opened 2025-05-13
miinwin: Get to the CD check

AJenbo:
Still no graphics but this implements enough that Isle now thinks there is:

https://github.com/user-attachments/assets/817e3a42-8042-4f72-a6cd-7b5f5819a81c

(unmute for audio)

### DIFF
--- a/miniwin/miniwin/include/miniwin_ddraw.h
+++ b/miniwin/miniwin/include/miniwin_ddraw.h
@@ -212,7 +212,7 @@ struct DDSURFACEDESC {
 typedef struct DDSURFACEDESC* LPDDSURFACEDESC;
 
 typedef struct IDirectDraw* LPDIRECTDRAW;
-struct IDirectDrawPalette : public IUnknown {
+struct IDirectDrawPalette : virtual public IUnknown {
 	virtual HRESULT GetCaps(LPDWORD lpdwCaps) = 0;
 	virtual HRESULT GetEntries(DWORD dwFlags, DWORD dwBase, DWORD dwNumEntries, LPPALETTEENTRY lpEntries) = 0;
 	virtual HRESULT Initialize(LPDIRECTDRAW lpDD, DWORD dwFlags, LPPALETTEENTRY lpDDColorTable) = 0;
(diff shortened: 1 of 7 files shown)


miniwin/miniwin/src/miniwin_ddraw.cpp
```diff
@@ -80,12 +133,10 @@ struct DirectDrawImpl : public IDirectDraw2, public IDirect3D2 {
 	// IDirect3D2 interface
 	HRESULT CreateDevice(const GUID& guid, void* pBackBuffer, IDirect3DDevice2** ppDirect3DDevice) override
 	{
-		return DDERR_GENERIC;
+		*ppDirect3DDevice = new IDirect3DDevice2;
```
madebr, line 136: I think we need to create a private `miniwin_private.h` header where we put all class declarations, so ddraw can see the classes of d3d and vice versa.

```suggestion
		*ppDirect3DDevice = static_cast<IDirect3DDevice2 *>(new Direct3DDeviceImpl);
```

miniwin/miniwin/include/miniwin_ddraw.h
```diff
@@ -267,7 +267,7 @@ struct IDirectDrawSurface3 : public IDirectDrawSurface {};
 typedef IDirectDrawSurface3* LPDIRECTDRAWSURFACE3;
 
 typedef HRESULT (*LPDDENUMMODESCALLBACK)(LPDDSURFACEDESC, LPVOID);
-struct IDirectDraw : public IUnknown {
+struct IDirectDraw : virtual public IUnknown {
```
madebr, line 270: Can you apply this to all other IUnknown children?

madebr (approved):
Looks good to me, with my suggested changes applied.

Good job finding the hidden diamond! :)

miniwin/miniwin/src/miniwin_ddraw.cpp
AJenbo: This class has no methods it's basically just a fancy void pointer, no need to cast or implement.

miniwin/miniwin/include/miniwin_ddraw.h
AJenbo: sure

miniwin/miniwin/src/miniwin_ddraw.cpp
madebr: True. It's probably used to make sure the device is not accidentally cleaned up.
Something to remember.

madebr (approved):

PR #77, opened 2025-05-13
Move MFC supporting functions to minimfc

AJenbo:
This makes it a lot easier to prioritize what bits to work on.


foxtacles (approved):

madebr (approved):
Makes sense to decouple mfc requirements from isle.

PR #80, opened 2025-05-14
Use `WM_QUIT` consistently

### LINKED ISSUE #79: `ISLE.EXE` doesn't exit properly
Both when existing the game through the credits or the NOCD animation, the app seems to be stuck at the end and doesn't quit

foxtacles:
`WM_CLOSE` is being sent in `CloseMainWindow`

close #79

### DIFF
--- a/ISLE/isleapp.cpp
+++ b/ISLE/isleapp.cpp
@@ -400,7 +400,7 @@ SDL_AppResult SDL_AppEvent(void* appstate, SDL_Event* event)
 		case WM_ISLE_SETCURSOR:
 			g_isle->SetupCursor((Cursor) (uintptr_t) event->user.data1);
 			break;
-		case WM_CLOSE:
+		case WM_QUIT:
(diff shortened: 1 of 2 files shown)


ISLE/isleapp.cpp
```diff
@@ -400,7 +400,7 @@ SDL_AppResult SDL_AppEvent(void* appstate, SDL_Event* event)
 		case WM_ISLE_SETCURSOR:
 			g_isle->SetupCursor((Cursor) (uintptr_t) event->user.data1);
 			break;
-		case WM_QUIT:
+		case WM_CLOSE:
```
madebr, line 403: Reading this [stackoverflow answer](https://stackoverflow.com/questions/3155782/what-is-the-difference-between-wm-quit-wm-close-and-wm-destroy-in-a-windows-pr), I'd emit `WM_QUIT` inside lego1.

But either ways, this approach looks good to me.

madebr (approved):

PR #81, opened 2025-05-14
Implement enough DirectDraw to get the nocd video playing

AJenbo:
This implements the major parts needed to get DirectDraw working, at least well enough to render the NoCD "intro".

There are also a few minor things added to the D3DRM side of things but I'm unable to progress further down that path because of the data issues in current builds that are causing odd crashes.

https://github.com/user-attachments/assets/8bcdfad5-5d11-43a6-b6a4-1fc779434fa6

### DIFF
--- a/miniwin/minimfc/include/minimfc.h
+++ b/miniwin/minimfc/include/minimfc.h
@@ -82,15 +82,9 @@ struct OSVERSIONINFOA {
 typedef struct IUnknown* LPUNKNOWN;
 
 struct CWnd {
-	void* m_hWnd;
+	HWND m_hWnd;
 	void EnableWindow(bool bEnable) {}
 	void SetWindowText(const char* text) {}
-};
-
-struct CPaintDC {
-	void* m_hDC;
-	CPaintDC(HWND hWnd) {}
-	void Draw() {}
 };
 
 struct CDataExchange {
@@ -98,7 +92,7 @@ struct CDataExchange {
 };
 
 struct CDialog {
-	void* m_hWnd;
+	HWND m_hWnd;
 	int m_nIDTemplate;
 	CWnd* m_pParentWnd;
 	CDialog() : m_nIDTemplate(0), m_pParentWnd(nullptr) {}
@@ -111,6 +105,12 @@ struct CDialog {
 	virtual void Default() {}
 	virtual void EndDialog(int nResult) {}
 	virtual void DoDataExchange(CDataExchange* pDX) {}
+};
+
+struct CPaintDC {
+	void* m_hDC;
+	CPaintDC(CDialog* hWnd) {}
+	void Draw() {}
 };
 
 struct CMenu {
@@ -166,7 +166,7 @@ inline BOOL ShowWindow(HWND hWnd, int nCmdShow)
 	return TRUE;
 }
 
-inline BOOL SetWindowPos(HWND hWndInsertAfter, int X, int Y, int cx, int cy, UINT uFlags)
+inline BOOL SetWindowPos(CWinApp** hWnd, int X, int Y, int cx, int cy, UINT uFlags)
 {
 	return TRUE;
 }
--- a/miniwin/miniwin/src/miniwin_ddraw.cpp
+++ b/miniwin/miniwin/src/miniwin_ddraw.cpp
@@ -3,8 +3,39 @@
 #include "miniwin_d3d.h"
 
 #include <SDL3/SDL.h>
+#include <assert.h>
+#include <cstdint>
+#include <cstdlib>
+#include <cstring>
+
+SDL_Renderer* renderer;
+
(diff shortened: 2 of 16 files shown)


miniwin/miniwin/include/miniwin_d3drm.h
```diff
@@ -152,7 +152,8 @@ typedef IDirect3DRMTexture* LPDIRECT3DRMTEXTURE;
 struct IDirect3DRMTexture2 : public IDirect3DRMTexture {};
 typedef IDirect3DRMTexture2* LPDIRECT3DRMTEXTURE2;
 
-typedef struct IDirect3DRMMaterial *LPDIRECT3DRMMATERIAL, **LPLPDIRECT3DRMMATERIAL;
+struct IDirect3DRMMaterial {};
```
madebr, line 155: I'm surprised this does not need to inheris `IUnknown`. How is it freed?

AJenbo: Oops. Its a good question though, my initial thinking was that there is a lot of leakage, but I guess it's just passed to another com object which will cause it to get freed when that one is freed, we don't do so yet

madebr: When in doubt, can we add this? Along the line, I think we want to rely on correct clean-up by the reference counts.

AJenbo: It's on it's way :)

miniwin/miniwin/src/miniwin_ddraw.cpp
```diff
@@ -125,7 +176,14 @@ struct DirectDrawImpl : public IDirectDraw2, public IDirect3D2 {
 	}
 	HRESULT FlipToGDISurface() override { return DD_OK; }
 	HRESULT GetCaps(LPDDCAPS lpDDDriverCaps, LPDDCAPS lpDDHELCaps) override;
-	HRESULT GetDisplayMode(LPDDSURFACEDESC lpDDSurfaceDesc) override { return DD_OK; }
+	HRESULT GetDisplayMode(LPDDSURFACEDESC lpDDSurfaceDesc) override
+	{
+		lpDDSurfaceDesc->ddpfPixelFormat.dwRBitMask = 5;
+		lpDDSurfaceDesc->ddpfPixelFormat.dwGBitMask = 6;
+		lpDDSurfaceDesc->ddpfPixelFormat.dwBBitMask = 5;
```
AJenbo, line 183: This is all wrong, I need to fix

miniwin/miniwin/src/miniwin_ddraw.cpp
```diff
@@ -26,18 +57,51 @@ struct DirectDrawSurfaceImpl : public IDirectDrawSurface3 {
...
+		}
+		SDL_FRect dstRect = {
+			(float) dwX,
+			(float) dwY,
+			(float) (lpSrcRect->right - lpSrcRect->left),
+			(float) (lpSrcRect->bottom - lpSrcRect->top)
+		};
+		SDL_FRect srcRect = ConvertRect(lpSrcRect);
+		SDL_RenderTexture(renderer, static_cast<DirectDrawSurfaceImpl*>(lpDDSrcSurface)->texture, &srcRect, &dstRect);
+		SDL_RenderPresent(renderer);
```
madebr, line 83: I think you only need to do `SDL_RenderPresent` in `IDirectDrawSurface::Flip`?
Untested, so please verify.
```suggestion
```

AJenbo: No that won't work, the game does not call both. I know it seems wrong but they either `memcpy()` to the back buffer and call `->Flip()` or they just call `->Blt()` even though that looks wrong according to the docs I have looked at.

miniwin/minimfc/include/minimfc.h
```diff
@@ -166,7 +166,7 @@ inline BOOL ShowWindow(HWND hWnd, int nCmdShow)
 	return TRUE;
 }
 
-inline BOOL SetWindowPos(HWND hWndInsertAfter, int X, int Y, int cx, int cy, UINT uFlags)
+inline BOOL SetWindowPos(CWinApp** hWndInsertAfter, int X, int Y, int cx, int cy, UINT uFlags)
```
madebr, line 169: [SetWindowPos is a CWnd member function](http://www.icodeguru.com/VC%26MFC/MFCReference/html/_mfc_cwnd.3a3a.setwindowpos.htm)
So this should be:
```c++
BOOL CWnd::SetWindowPos(int X, int Y, int cx, int cy, UINT uFlags)
```

AJenbo: I'm not really sure how to implement that or if it requires changes to the decomp project so if you could take care of it it would be much appreciated :)

madebr (approved):

PR #82, opened 2025-05-14
Handle unaligned read/write

AJenbo:
This tries to address many of the unaligned reads. I can confirm that I still have the NOCD audio playing, but I need someone to test if the game still loads with d3drm.dll and the CD.

### DIFF
--- a/LEGO1/omni/src/action/mxdsaction.cpp
+++ b/LEGO1/omni/src/action/mxdsaction.cpp
@@ -261,21 +261,21 @@ void MxDSAction::Deserialize(MxU8*& p_source, MxS16 p_flags)
 	MxDSObject::Deserialize(p_source, p_flags);
 
 	// clang-format off
-	m_flags           = *( MxU32*) p_source;  p_source += sizeof(m_flags);
-	m_startTime       = *(MxLong*) p_source;  p_source += sizeof(m_startTime);
-	m_duration        = *(MxLong*) p_source;  p_source += sizeof(m_duration);
-	m_loopCount       = *( MxS32*) p_source;  p_source += sizeof(m_loopCount);
-	m_location[0]     = *(double*) p_source;  p_source += sizeof(double);
-	m_location[1]     = *(double*) p_source;  p_source += sizeof(double);
-	m_location[2]     = *(double*) p_source;  p_source += sizeof(double);
-	m_direction[0]    = *(double*) p_source;  p_source += sizeof(double);
-	m_direction[1]    = *(double*) p_source;  p_source += sizeof(double);
-	m_direction[2]    = *(double*) p_source;  p_source += sizeof(double);
-	m_up[0]           = *(double*) p_source;  p_source += sizeof(double);
-	m_up[1]           = *(double*) p_source;  p_source += sizeof(double);
-	m_up[2]           = *(double*) p_source;  p_source += sizeof(double);
+	m_flags           = UnalignedRead<MxU32>(p_source);   p_source += sizeof(MxU32);
+	m_startTime       = UnalignedRead<MxLong>(p_source);  p_source += sizeof(MxLong);
+	m_duration        = UnalignedRead<MxLong>(p_source);  p_source += sizeof(MxLong);
+	m_loopCount       = UnalignedRead<MxS32>(p_source);   p_source += sizeof(MxS32);
+	m_location[0]     = UnalignedRead<double>(p_source);  p_source += sizeof(double);
+	m_location[1]     = UnalignedRead<double>(p_source);  p_source += sizeof(double);
+	m_location[2]     = UnalignedRead<double>(p_source);  p_source += sizeof(double);
+	m_direction[0]    = UnalignedRead<double>(p_source);  p_source += sizeof(double);
+	m_direction[1]    = UnalignedRead<double>(p_source);  p_source += sizeof(double);
+	m_direction[2]    = UnalignedRead<double>(p_source);  p_source += sizeof(double);
+	m_up[0]           = UnalignedRead<double>(p_source);  p_source += sizeof(double);
+	m_up[1]           = UnalignedRead<double>(p_source);  p_source += sizeof(double);
+	m_up[2]           = UnalignedRead<double>(p_source);  p_source += sizeof(double);
 
-	MxU16 extraLength = *( MxU16*) p_source;  p_source += sizeof(extraLength);
+	MxU16 extraLength = UnalignedRead<MxU16>(p_source);  p_source += sizeof(extraLength);
 	// clang-format on
 
 	if (extraLength) {
--- a/LEGO1/omni/src/action/mxdsmediaaction.cpp
+++ b/LEGO1/omni/src/action/mxdsmediaaction.cpp
@@ -137,12 +137,12 @@ void MxDSMediaAction::Deserialize(MxU8*& p_source, MxS16 p_flags)
 	p_source += strlen(m_mediaSrcPath) + 1;
 
 	// clang-format off
-	m_unk0x9c.SetUnk0x00(*(MxU32*) p_source);  p_source += sizeof(m_unk0x9c.m_unk0x00);
-	m_unk0x9c.SetUnk0x04(*(MxU32*) p_source);  p_source += sizeof(m_unk0x9c.m_unk0x04);
+	m_unk0x9c.SetUnk0x00(UnalignedRead<MxU32>(p_source));  p_source += sizeof(MxU32);
+	m_unk0x9c.SetUnk0x04(UnalignedRead<MxU32>(p_source));  p_source += sizeof(MxU32);
 
-	m_framesPerSecond   = *(MxS32*) p_source;  p_source += sizeof(m_framesPerSecond);
-	m_mediaFormat       = *(MxS32*) p_source;  p_source += sizeof(m_mediaFormat);
-	m_paletteManagement = *(MxS32*) p_source;  p_source += sizeof(m_paletteManagement);
-	m_sustainTime       = *(MxS32*) p_source;  p_source += sizeof(m_sustainTime);
+	m_framesPerSecond   = UnalignedRead<MxS32>(p_source);  p_source += sizeof(MxS32);
+	m_mediaFormat       = UnalignedRead<MxS32>(p_source);  p_source += sizeof(MxS32);
+	m_paletteManagement = UnalignedRead<MxS32>(p_source);  p_source += sizeof(MxS32);
+	m_sustainTime       = UnalignedRead<MxS32>(p_source);  p_source += sizeof(MxS32);
 	// clang-format on
 }
--- a/LEGO1/omni/src/stream/mxdsbuffer.cpp
+++ b/LEGO1/omni/src/stream/mxdsbuffer.cpp
@@ -9,6 +9,7 @@
 #include "mxstreamcontroller.h"
 #include "mxstreamer.h"
 #include "mxstreamprovider.h"
+#include "mxutilities.h"
 
 DECOMP_SIZE_ASSERT(MxDSBuffer, 0x34);
 
@@ -177,11 +178,11 @@ MxResult MxDSBuffer::CreateObject(
 		return FAILURE;
 	}
 
-	if (*p_data == FOURCC('M', 'x', 'O', 'b')) {
+	if (UnalignedRead<MxU32>((MxU8*) p_data) == FOURCC('M', 'x', 'O', 'b')) {
 		MxDSAction* action = (MxDSAction*) header;
 		return StartPresenterFromAction(p_controller, p_action, action);
 	}
-	else if (*p_data == FOURCC('M', 'x', 'C', 'h')) {
+	else if (UnalignedRead<MxU32>((MxU8*) p_data) == FOURCC('M', 'x', 'C', 'h')) {
 		MxStreamChunk* chunk = (MxStreamChunk*) header;
 		if (!m_unk0x30->HasId((chunk)->GetObjectId())) {
 			delete header;
@@ -326,7 +327,7 @@ MxCore* MxDSBuffer::ReadChunk(MxDSBuffer* p_buffer, MxU32* p_chunkData, MxU16 p_
 	MxCore* result = NULL;
 	MxU8* dataStart = (MxU8*) p_chunkData + 8;
 
-	switch (*p_chunkData) {
+	switch (UnalignedRead<MxU32>((MxU8*) p_chunkData)) {
 	case FOURCC('M', 'x', 'O', 'b'): {
 		MxDSObject* obj = DeserializeDSObjectDispatch(dataStart, p_flags);
 		result = obj;
@@ -355,11 +356,12 @@ MxU8* MxDSBuffer::SkipToData()
 
 	if (m_pIntoBuffer != NULL) {
 		while (TRUE) {
-			switch (*(MxU32*) m_pIntoBuffer) {
+			switch (UnalignedRead<MxU32>(m_pIntoBuffer)) {
 			case FOURCC('M', 'x', 'O', 'b'):
 			case FOURCC('M', 'x', 'C', 'h'):
 				result = m_pIntoBuffer;
-				m_pIntoBuffer += (*(MxU32*) (m_pIntoBuffer + 4) & 1) + *(MxU32*) (m_pIntoBuffer + 4);
+				m_pIntoBuffer +=
+					(UnalignedRead<MxU32>(m_pIntoBuffer + 4) & 1) + UnalignedRead<MxU32>(m_pIntoBuffer + 4);
 				m_pIntoBuffer += 8;
 
 				if (m_pBuffer + m_writeOffset - 8 < m_pIntoBuffer) {
@@ -372,7 +374,7 @@ MxU8* MxDSBuffer::SkipToData()
 				m_pIntoBuffer += 8;
 				break;
 			case FOURCC('M', 'x', 'H', 'd'):
-				m_pIntoBuffer += *(MxU32*) (m_pIntoBuffer + 4) + 8;
+				m_pIntoBuffer += UnalignedRead<MxU32>(m_pIntoBuffer + 4) + 8;
 				break;
 			case FOURCC('L', 'I', 'S', 'T'):
 			case FOURCC('R', 'I', 'F', 'F'):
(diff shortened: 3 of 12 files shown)


LEGO1/omni/src/action/mxdsaction.cpp
```diff
@@ -258,24 +258,27 @@ void MxDSAction::AppendExtra(MxU16 p_extraLength, const char* p_extraData)
...
-	m_location[2]     = *(double*) p_source;  p_source += sizeof(double);
-	m_direction[0]    = *(double*) p_source;  p_source += sizeof(double);
-	m_direction[1]    = *(double*) p_source;  p_source += sizeof(double);
-	m_direction[2]    = *(double*) p_source;  p_source += sizeof(double);
-	m_up[0]           = *(double*) p_source;  p_source += sizeof(double);
-	m_up[1]           = *(double*) p_source;  p_source += sizeof(double);
-	m_up[2]           = *(double*) p_source;  p_source += sizeof(double);
-
-	MxU16 extraLength = *( MxU16*) p_source;  p_source += sizeof(extraLength);
+	memcpy(&m_flags,     p_source, sizeof(MxU32));  p_source += sizeof(MxU32);
```
foxtacles, line 266: Maybe instead of using memcpy everywhere we could add a small template utility:

```
template <typename T>
T UnalignedRead(MxU8* p_source) {
    T value;
    memcpy(&value, p_source, sizeof(T));
    return value;
}
```
(untested)
I think it would make the code a bit more readable, and no need for extra temporary vars in some cases

foxtacles:
I'll test this tomorrow

LEGO1/omni/src/action/mxdsaction.cpp
AJenbo: Where would you like me to put it? I did consider it but dropped it once I had a version of it in 10 files.

LEGO1/omni/src/stream/mxramstreamprovider.cpp
```diff
@@ -105,14 +105,16 @@ MxResult MxRAMStreamProvider::SetResourceToGet(MxStreamController* p_resource)
...
 	MxU32 id;
 	MxU8* data = p_buffer;
 	MxU8* data2;
 
 #define IntoType(p) ((MxU32*) (p))
 
 	while (data < p_buffer + p_size) {
-		if (*IntoType(data) == FOURCC('M', 'x', 'O', 'b')) {
+		memcpy(&fourcc, IntoType(data), sizeof(MxU32));
+		if (fourcc == FOURCC('M', 'x', 'O', 'b')) {
```
AJenbo, line 117: This one needs extra attention, without it my hunt for UB stops here.
With it ASAN yells at me: unknown-crash

LEGO1/omni/src/action/mxdsaction.cpp
foxtacles: hmm, I'd say either:

https://github.com/isledecomp/isle-portable/blob/master/LEGO1/omni/include/mxutilities.h

it already has similar functions.

or this one:

https://github.com/isledecomp/isle-portable/blob/master/util/compat.h

since it contains "compatibility" related things.

@madebr wdyt

LEGO1/omni/src/stream/mxramstreamprovider.cpp
AJenbo: And without ASAN i get this:
```
terminate called after throwing an instance of 'std::bad_array_new_length'
  what():  std::bad_array_new_length
```
The only remaining option is using a debugger and I don't find that fun so it's up to someone else :P

LEGO1/omni/src/action/mxdsaction.cpp
madebr: I'd put it in `mxutilities.h`. I think this is supported by early c++ compilers?

foxtacles:
> but I need someone to test if the game still loads with d3drm.dll and the CD.

Tested on win32 and everything works as far as I can tell. Code changes look good to me tot. I'd just apply the refactor to use the template utility as discussed above and then it should be good to merge.

LEGO1/omni/src/stream/mxdsbuffer.cpp
```diff
@@ -351,16 +355,19 @@ MxCore* MxDSBuffer::ReadChunk(MxDSBuffer* p_buffer, MxU32* p_chunkData, MxU16 p_
 // FUNCTION: BETA10 0x10157e0a
 MxU8* MxDSBuffer::SkipToData()
 {
+	MxU32 fourcc;
+	MxU32 size;
 	MxU8* result = NULL;
 
 	if (m_pIntoBuffer != NULL) {
 		while (TRUE) {
-			switch (*(MxU32*) m_pIntoBuffer) {
```
madebr, line 358: You can move `fourcc` and `size` into the while loop.

LEGO1/omni/src/stream/mxdsbuffer.cpp
```diff
@@ -458,11 +470,13 @@ void MxDSBuffer::FUN_100c6f80(MxU32 p_writeOffset)
...
 MxU8* MxDSBuffer::FUN_100c6fa0(MxU8* p_data)
 {
+	MxU32 fourcc;
 	MxU8* current = p_data ? p_data : m_pBuffer;
 	MxU8* end = m_writeOffset + m_pBuffer - 8;
 
 	while (current <= end) {
-		switch (*((MxU32*) current)) {
+		memcpy(&fourcc, current, sizeof(MxU32));
+		switch (fourcc) {
```
madebr, line 479: Here too.

LEGO1/omni/src/stream/mxramstreamprovider.cpp
AJenbo: Managed to solve both issues

LEGO1/omni/src/action/mxdsaction.cpp
```diff
@@ -261,21 +261,21 @@ void MxDSAction::Deserialize(MxU8*& p_source, MxS16 p_flags)
...
-	m_direction[2]    = *(double*) p_source;  p_source += sizeof(double);
-	m_up[0]           = *(double*) p_source;  p_source += sizeof(double);
-	m_up[1]           = *(double*) p_source;  p_source += sizeof(double);
-	m_up[2]           = *(double*) p_source;  p_source += sizeof(double);
-
-	MxU16 extraLength = *( MxU16*) p_source;  p_source += sizeof(extraLength);
+	m_flags           = UnalignedRead<MxU32>(p_source);   p_source += sizeof(m_flags);
+	m_startTime       = UnalignedRead<MxLong>(p_source);  p_source += sizeof(m_startTime);
+	m_duration        = UnalignedRead<MxLong>(p_source);  p_source += sizeof(m_duration);
+	m_loopCount       = UnalignedRead<MxS32>(p_source);   p_source += sizeof(m_loopCount);
```
madebr, line 267:
```suggestion
	m_flags           = UnalignedRead<MxU32>(p_source);   p_source += sizeof(MxU32);
	m_startTime       = UnalignedRead<MxLong>(p_source);  p_source += sizeof(MxLong);
	m_duration        = UnalignedRead<MxLong>(p_source);  p_source += sizeof(MxLong);
	m_loopCount       = UnalignedRead<MxS32>(p_source);   p_source += sizeof(MxS32);
```

LEGO1/omni/src/action/mxdsmediaaction.cpp
```diff
@@ -137,12 +137,12 @@ void MxDSMediaAction::Deserialize(MxU8*& p_source, MxS16 p_flags)
 	p_source += strlen(m_mediaSrcPath) + 1;
 
 	// clang-format off
-	m_unk0x9c.SetUnk0x00(*(MxU32*) p_source);  p_source += sizeof(m_unk0x9c.m_unk0x00);
-	m_unk0x9c.SetUnk0x04(*(MxU32*) p_source);  p_source += sizeof(m_unk0x9c.m_unk0x04);
+	m_unk0x9c.SetUnk0x00(UnalignedRead<MxU32>(p_source));  p_source += sizeof(m_unk0x9c.m_unk0x00);
+	m_unk0x9c.SetUnk0x04(UnalignedRead<MxU32>(p_source));  p_source += sizeof(m_unk0x9c.m_unk0x04);
```
madebr, line 141:
```suggestion
	m_unk0x9c.SetUnk0x00(UnalignedRead<MxU32>(p_source));  p_source += sizeof(MxU32);
	m_unk0x9c.SetUnk0x04(UnalignedRead<MxU32>(p_source));  p_source += sizeof(MxU32);
```

LEGO1/omni/src/action/mxdsmediaaction.cpp
```diff
@@ -137,12 +137,12 @@ void MxDSMediaAction::Deserialize(MxU8*& p_source, MxS16 p_flags)
...
 
-	m_framesPerSecond   = *(MxS32*) p_source;  p_source += sizeof(m_framesPerSecond);
-	m_mediaFormat       = *(MxS32*) p_source;  p_source += sizeof(m_mediaFormat);
-	m_paletteManagement = *(MxS32*) p_source;  p_source += sizeof(m_paletteManagement);
-	m_sustainTime       = *(MxS32*) p_source;  p_source += sizeof(m_sustainTime);
+	m_framesPerSecond   = UnalignedRead<MxS32>(p_source);  p_source += sizeof(m_framesPerSecond);
+	m_mediaFormat       = UnalignedRead<MxS32>(p_source);  p_source += sizeof(m_mediaFormat);
+	m_paletteManagement = UnalignedRead<MxS32>(p_source);  p_source += sizeof(m_paletteManagement);
+	m_sustainTime       = UnalignedRead<MxS32>(p_source);  p_source += sizeof(m_sustainTime);
```
madebr, line 146:
```suggestion
	m_framesPerSecond   = UnalignedRead<MxS32>(p_source);  p_source += sizeof(MxS32);
	m_mediaFormat       = UnalignedRead<MxS32>(p_source);  p_source += sizeof(MxS32);
	m_paletteManagement = UnalignedRead<MxS32>(p_source);  p_source += sizeof(MxS32);
	m_sustainTime       = UnalignedRead<MxS32>(p_source);  p_source += sizeof(MxS32);
```

madebr (approved):

LEGO1/omni/src/stream/mxramstreamprovider.cpp
```diff
@@ -109,10 +110,8 @@ MxU32 ReadData(MxU8* p_buffer, MxU32 p_size)
 	MxU8* data = p_buffer;
 	MxU8* data2;
 
-#define IntoType(p) ((MxU32*) (p))
-
 	while (data < p_buffer + p_size) {
-		if (*IntoType(data) == FOURCC('M', 'x', 'O', 'b')) {
+		if (data + sizeof(MxU32) <= p_buffer + p_size && UnalignedRead<MxU32>(data) == FOURCC('M', 'x', 'O', 'b')) {
```
foxtacles, line 114: What is this extra check for?

AJenbo: To make sure that we can read 4 byters from the buffer, sometimes there is less then 4 bytes left and it will cause an illegal read.

LEGO1/omni/src/stream/mxdsbuffer.cpp
```diff
@@ -435,7 +437,10 @@ MxResult MxDSBuffer::CalcBytesRemaining(MxU8* p_data)
 			memcpy(m_pBuffer + m_writeOffset - m_bytesRemaining, ptr, bytesRead);
 
 			if (m_writeOffset == m_bytesRemaining) {
-				*(MxU32*) (m_pBuffer + 4) = *MxStreamChunk::IntoLength(m_pBuffer) + MxStreamChunk::GetHeaderSize();
+				MxU32 length;
+				memcpy(&length, MxStreamChunk::IntoLength(m_pBuffer), sizeof(MxU32));
```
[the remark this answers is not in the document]
AJenbo: It's an assignment, not a read.

LEGO1/omni/src/stream/mxramstreamprovider.cpp
foxtacles: Would we get a *SAN error without this check?

LEGO1/omni/src/stream/mxdsbuffer.cpp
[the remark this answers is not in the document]
foxtacles: Ok. What am I missing - this wouldn't work?

MXU32 length = UnalignedRead<MxU32>(MxStreamChunk::IntoLength(m_pBuffer))

LEGO1/omni/src/stream/mxramstreamprovider.cpp
AJenbo: Yes. The fix to the UB issue causes and A error, both are fixed with this.

AJenbo: This should also work:
```cpp
	while (data + sizeof(MxU32) <= p_buffer + p_size) {
		if (UnalignedRead<MxU32>(data) == FOURCC('M', 'x', 'O', 'b')) {
			// stuff
		}
		else {
			data++;
		}
	}
```

foxtacles: yes got let's keep the current solution thank you

LEGO1/omni/src/stream/mxdsbuffer.cpp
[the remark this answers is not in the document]
AJenbo: Arh, I was thinking you where talking about the first part of the original line.

Changed it to:
```cpp
				MxU32 length =
					UnalignedRead<MxU32>((MxU8*) MxStreamChunk::IntoLength(m_pBuffer)) + MxStreamChunk::GetHeaderSize();
```

LEGO1/omni/src/stream/mxdsbuffer.cpp
```diff
@@ -435,7 +437,9 @@ MxResult MxDSBuffer::CalcBytesRemaining(MxU8* p_data)
 			memcpy(m_pBuffer + m_writeOffset - m_bytesRemaining, ptr, bytesRead);
 
 			if (m_writeOffset == m_bytesRemaining) {
-				*(MxU32*) (m_pBuffer + 4) = *MxStreamChunk::IntoLength(m_pBuffer) + MxStreamChunk::GetHeaderSize();
+				MxU32 length =
+					UnalignedRead<MxU32>((MxU8*) MxStreamChunk::IntoLength(m_pBuffer)) + MxStreamChunk::GetHeaderSize();
+				memcpy(m_pBuffer + 4, &length, sizeof(length));
```
foxtacles, line 442: I suppose for consistency we could add a UnalignedWrite helper. It seems to be the only instance where we'd need it though. Worth it?

AJenbo: Is there an easy way to not build miniaudio? If you can add a flag for that then I can continue to go through more of the game and fix more more of these.

foxtacles: I'm not aware, but maybe @madebr can help. Is the issue that *SAN triggers on miniaudio code? If yes, can we disable running those tools on 3rd party libraries?

madebr: With this patch, 32-bit fails in the same way as 64-bit for me:
```patch
--- a/LEGO1/omni/include/mxwavepresenter.h
+++ b/LEGO1/omni/include/mxwavepresenter.h
@@ -101,6 +101,7 @@ class MxWavePresenter : public MxSoundPresenter {
        } m_ab;
 
        ma_sound m_sound;
+       MxBool m_sound_initialized;
        MxU32 m_chunkLength; // 0x5c
        MxBool m_started;    // 0x65
        MxBool m_is3d;       // 0x66
--- a/LEGO1/omni/src/audio/mxwavepresenter.cpp
+++ b/LEGO1/omni/src/audio/mxwavepresenter.cpp
@@ -23,6 +23,7 @@ void MxWavePresenter::Init()
        SDL_zero(m_rb);
        SDL_zero(m_ab);
        SDL_zero(m_sound);
+       m_sound_initialized = 0;
        m_chunkLength = 0;
        m_started = FALSE;
        m_is3d = FALSE;
@@ -40,7 +41,9 @@ MxResult MxWavePresenter::AddToManager()
 // FUNCTION: LEGO1 0x100b1b10
 void MxWavePresenter::Destroy(MxBool p_fromDestructor)
 {
-       ma_sound_uninit(&m_sound);
+       if (m_sound_initialized) {
+               ma_sound_uninit(&m_sound);
+       }
        ma_pcm_rb_uninit(&m_rb);
        ma_audio_buffer_uninit(&m_ab.m_buffer);
        delete[] m_ab.m_data;
@@ -173,6 +176,7 @@ void MxWavePresenter::StartingTickle()
                        ) != MA_SUCCESS) {
                        goto done;
                }
+        m_sound_initialized = 1;
 
                // [library:audio]
                // There is an issue with certain spatialized sounds causing an audio glitch.
```

foxtacles: if you call uninit on a zero'ed sound object, it will cause corruption?
I'm not on a computer so I can't check miniaudio source but I assumed it was not problematic.

madebr: `ma_sound_uninit` starts like this:
```c
MA_API void ma_sound_uninit(ma_sound* pSound)
{
    if (pSound == NULL) {
        return;
    }
```
We're never passing it a non-null.

foxtacles: yes, but what does it do after that check

madebr: I starts to unconditionally dereference its members:
```c
    /*
    Always uninitialize the node first. This ensures it's detached from the graph and does not return until it has done
    so which makes thread safety beyond this point trivial.
    */
    ma_engine_node_uninit(&pSound->engineNode, &pSound->engineNode.pEngine->allocationCallbacks);
    ```
```
`pSound->engineNode.pEngine` is null

foxtacles: My assumption so far was that a zero'ed object was safe to call uninit on. If that's not the case we may have to refactor some.

foxtacles: ok thank you. Might have to fix the uses of the other Miniaudio objects as well. Maybe allocate them on the heap instead. Wouldn't want an extra bool/check for each

madebr: Or null sounds were never destroyed until now? What if you add an assertion on your Windows machine? This also means Linux is not yet on the happy path.

foxtacles: Opening an issue

AJenbo: > Or null sounds were never destroyed until now? 

That's very possible

PR #83, opened 2025-05-14
Fix all remaning warnings on GCC

AJenbo:


LEGO1/lego/legoomni/include/legoracespecial.h
```diff
@@ -35,14 +35,6 @@ class LegoCarRaceActor : public virtual LegoRaceActor {
 		return !strcmp(p_name, LegoCarRaceActor::ClassName()) || LegoRaceActor::IsA(p_name);
 	}
 
-	inline MxU32 VTable0x6c(
```
foxtacles, line 38: I don't think we can delete this code. I'd just remove the `inline` keyword. That should solve the warning?

LEGO1/lego/legoomni/src/worlds/historybook.cpp
```diff
@@ -141,10 +141,7 @@ void HistoryBook::ReadyWorld()
 #ifdef BETA10
 						memset(m_scores[i]->GetBitmapStart(scoreboxX, scoreboxY + lax), scoreColors[color - 1], 4);
 #else
-						if (m_scores[i]->GetAlphaMask() != NULL) {
-							memset(NULL, scoreColors[color - 1], 4);
```
foxtacles, line 145: I'd just comment out the memset call, but leave the branch intact. The more code we can keep unchanged the better

PR #85, opened 2025-05-14
isleapp: Add icon upon window creation

AllMeatball:
In window:
image

In taskbar:
image


foxtacles:
Thank you. Can someone explain why this is only needed on non-Windows systems?

ISLE/isleapp.cpp
```diff
@@ -486,6 +486,21 @@ MxResult IsleApp::SetupWindow()
 		return FAILURE;
 	}
 
+#ifndef _WIN32
```
foxtacles, line 489: We should use the SDL macro instead:

`#if !defined(SDL_PLATFORM_WINDOWS)`

AllMeatball:
@foxtacles you have a good point ISLE.exe already does this live
```
wndclass.hIcon = LoadIcon(hInstance, MAKEINTRESOURCE(APP_ICON));
```

there's a lot of apps just let it be handled by Windows itself. I just maybe thought it would do the same.

ISLE/isleapp.cpp
```diff
@@ -486,6 +486,19 @@ MxResult IsleApp::SetupWindow()
 		return FAILURE;
 	}
 
+	SDL_Surface* icon = SDL_CreateSurfaceFrom(
+		ISLE_ICON.width,
+		ISLE_ICON.height,
+		SDL_PIXELFORMAT_RGBA32,
```
AllMeatball, line 492: I'm not quite sure if this will work for big endian systems or not.

AllMeatball: I've tried SDL_PIXELFORMAT_RGBA8888 but it ends up looking like this:
image

ISLE/res/resource.h
```diff
@@ -3,3 +3,5 @@
 #define ISLE_BUSY 104
 
 #define APP_ICON 105
```
foxtacles, line 5: Isn't it possible to re-use the icon resource declared in `isle.rc`?

AllMeatball: I think a library would be needed to read .ico files.

AllMeatball: I guess SDL_image could do it but there's probably a better way with a smaller library.

madebr:
> Thank you. Can someone explain why this is only needed on non-Windows systems?

Windows has a resource compiler, which can embed icons, dialogs, bitmaps, ... (all binary data) in [the resource section of a PE image](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#the-rsrc-section).
The first icon in this section is used as the shell icon (as can be seen in Windows Explorer).
This icon is also used when creating the Window through Windows' `CreateWindowEx`


[Raymond Chen recently wrote something about it](https://devblogs.microsoft.com/oldnewthing/20250423-00/?p=111106)

foxtacles:
It sounds like the resources we currently embed aren't portable then. In that case let's get rid of them entirely

madebr:
An alternative (and I think better better solution) would be to open `ISLE.EXE`, parse the PE header, look for the first icon (using a heuristic described in Raymond Chen's article) and use that. This avoids embedding of binary data.
I don't think it's unreasonable to expect the game data to also contain the Windows executable.

madebr:
On the other hand, we currently already embed the icons (see `res` directory).
And using the original exe, there's a possibility isle-portable's exe overwrites the original ISLE.EXE.

Alternatively, we can embed all icons using `#embed`, or convert them to binary data similar as is done in this pr.

foxtacles:
Let's embed and use all the resources in a portable way then. The `isle.rc` resource file seems to be specific to Windows, so let's get rid of it (unless it can be re-used in some way?). If the resource formats like `.ico` cannot be used by SDL without a lot of extra effort / conversion, I'm fine with embedding a converted version.

AllMeatball:
In my own engine I just have files for each icon (e.g icon256.png through icon8.png). but that uses raylib and uses a png decoder (built-in to raylib). SDL doesn't support multiple icons to be set, just only one icon.

I would say the embedding is fine. maybe remove the files in the res/ folder when properly replaced?

Note: in the Windows build it embeds an icon(s) in the executable still.

AJenbo:
> It sounds like the resources we currently embed aren't portable then. In that case let's get rid of them entirely

Without it there won't be a file icon on Windows, each platform has there own way and format for icons so you will need to maintain each one there isn't a common way to have it work while the application isn't running. The .rc file is also where you would declare your manifest to let Windows know your are a modern app with HighDpi support etc.

For Apple you have .plist instead which again defines the icon to use and enables HighDpi support. It's icon format is .icns.

On Linux the meta is in .desktop and the icon is .png or .svg.

AJenbo:
On DevilutionX we have a folder called Packaging where we put all the platform specific resources: https://github.com/diasurgical/DevilutionX/tree/master/Packaging

AJenbo (approved):

ISLE/res/resource.h
```diff
@@ -3,3 +3,8 @@
...
+
+#ifndef _RES_ISLE_H
+#define _RES_ISLE_H
+#include "isle.h"
+#endif
```
madebr, line 10: This feels awkward.
I propose to name the generated `isle.h` `isle_bmp.h`.
Simply include `isle_bmp.h` in `isleapp.cpp`, and leave `resource.h` alone.

madebr (approved):

AJenbo:
In DevilutionX we did something very similar between 0.5.0 and 1.5.0 where we eventually moved it to the accompanying .mpq archive (what could be portable.si in your case)

PR #87, opened 2025-05-14
Correct check for lighting support legovideomanager.cpp

AJenbo:
It's pure luck this worked, `DDCAPS2_CERTIFIED` happens to be `1` so the compiler boils it down to an inverse boolean check which then means lighting is enabled on all hardware that sets _any_ `DDCAPS2_*` flag.
Microsoft defines `BOOL` the same as `DWORD` and `DDCAPS2_CERTIFIED` and `TRUE` are both `1`, so it makes sens that there compiler would have this as an optimization.

I feel this is a general trend, who ever was in charge of integrating with DirectX didn't know how to work with bitflags or at least forgot a lot.


AJenbo:
FFS wrong repo again

PR #94, opened 2025-05-16
Implement more D3DRM skeleton

AJenbo:


miniwin/miniwin/src/miniwin_d3drm.cpp
```diff
@@ -191,17 +206,17 @@ struct Direct3DRMImpl : public IDirect3DRM2 {
...
 	HRESULT CreateTexture(D3DRMIMAGE* image, IDirect3DRMTexture2** outTexture) override
 	{
-		assert(false && "unimplemented");
+		*outTexture = static_cast<IDirect3DRMTexture2*>(new Direct3DRMTextureImpl);
 		return DDERR_GENERIC;
 	}
 	HRESULT CreateTextureFromSurface(LPDIRECTDRAWSURFACE surface, IDirect3DRMTexture2** outTexture) override
 	{
-		assert(false && "unimplemented");
+		*outTexture = static_cast<IDirect3DRMTexture2*>(new Direct3DRMTextureImpl);
```
madebr, line 214: I think `Direct3DRMTextureImpl` should get the `SDL_Texture` pointer from the underlying `DirectDrawSurfaceImpl`, and add a reference count.
Perhaps we should wrap `SDL_Texture` with a smart pointer (`std::shared_ptr`)?

madebr: But on the other hand, we're touching the limits of the SDL_Render API: the SDL_Texture is not usable in 3d contexts.

madebr (approved):

miniwin/miniwin/src/miniwin_d3drm.cpp
AJenbo: Yeah I want to first make a dummy that does nothing similar to how DirectDraw was brought up.

PR #95, opened 2025-05-16
isleapp: Show error dialog if NOCD.si or ISLE.si fails to load

AllMeatball:
image

### DIFF
--- a/ISLE/isleapp.cpp
+++ b/ISLE/isleapp.cpp
@@ -743,6 +743,7 @@ inline bool IsleApp::Tick()
 	if (!stream) {
 		stream = Streamer()->Open("\\lego\\scripts\\nocd", MxStreamer::e_diskStream);
 		if (!stream) {
+			SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "Failed to open NOCD.si: Streamer failed to load");
 			SDL_ShowSimpleMessageBox(
 				SDL_MESSAGEBOX_ERROR,
 				"LEGO® Island Error",
@@ -759,7 +760,15 @@ inline bool IsleApp::Tick()
 		VideoManager()->EnableFullScreenMovie(TRUE, TRUE);
 
 		if (Start(&ds) != SUCCESS) {
-			return true;
+			SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "Failed to open NOCD.si: Failed to start initial action");
+			SDL_ShowSimpleMessageBox(
+				SDL_MESSAGEBOX_ERROR,
+				"LEGO® Island Error",
+				"\"LEGO® Island\" failed to start.\nPlease quit all other applications and try again."
+				"\nFailed to load NOCD.si",
+				NULL
+			);
+			return false;
 		}
 	}
 	else {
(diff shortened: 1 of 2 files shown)


Ramen2X (commented):
I'm all for more descriptive errors; especially because LEGO Island's are pretty bad, but wouldn't this type of patch be more suited for the modernization fork that is eventually derived from isle-portable? The sole goal of this repo is to make LEGO Island cross-platform with as little changes to the integrity of the code as possible, which includes potential objective improvements to the code. Quoting the README:

>Please note: this project is dedicated to achieving platform independence without altering the core gameplay, adding new features, enhancing visual quality, or rewriting code for improvement's sake. While those are worthwhile objectives, they are not within the scope of this project.

AllMeatball:
@Ramen2X Good point, in my opinion it's not a big deal and most people will usually have the game files when booting up. this PR is mostly for people who are confused why game isn't booting. I think errors are fine as long they don't take they don't take away the original version's quirks. I think this is annoying quirk and should be clear why it is happening.

You can feel free to close if you don't think this really fits into the rules.

foxtacles:
@Ramen2X has a good point and that's correct in general.

In this particular instance I wouldn't mind adding the error - after all, almost all of the frontend (`isleapp`) had to be rewritten for portability anyway so there isn't much of the original code left. And since this issue (missing files) has become a real issue on non-Windows systems I think it makes sense to add it.

ISLE/isleapp.cpp
```diff
@@ -739,6 +740,15 @@ inline void IsleApp::Tick()
 	if (!stream) {
 		stream = Streamer()->Open("\\lego\\scripts\\nocd", MxStreamer::e_diskStream);
 		if (!stream) {
+			SDL_ShowSimpleMessageBox(
```
foxtacles, line 743: Instead of raising the error here, I'd change the interface of `Tick` to return a bool.  In this case, it would return `false`, and then we can raise the error in the game loop and exit there with `SDL_APP_FAILURE`. We won't need the `g_closeResult` for that then.

AllMeatball: I might need to add error string returns into the streamer code to get an informed error.

ISLE/isleapp.cpp
```diff
@@ -758,18 +759,20 @@ inline void IsleApp::Tick()
 		VideoManager()->EnableFullScreenMovie(TRUE, TRUE);
 
 		if (Start(&ds) != SUCCESS) {
-			return;
+			return true;
```
foxtacles, line 762: I think this is another error case where we should return `false`

ISLE/isleapp.cpp
```diff
@@ -758,18 +759,20 @@ inline void IsleApp::Tick()
...
 		}
 	}
 	else {
 		ds.SetAtomId(stream->GetAtom());
 		ds.SetUnknown24(-1);
 		ds.SetObjectId(0);
 		if (Start(&ds) != SUCCESS) {
-			return;
+			return true;
```
foxtacles, line 770: I think this is another error case where we should return `false`

ISLE/isleapp.cpp
foxtacles: I'd put the more informed error as `SDL_Log` calls inside `Tick`, and have the actual error pop up be generic

PR #100, opened 2025-05-17
Map internal paths to filesystem paths

foxtacles:
Resolves the issue on non-Windows systems where files can't be found. LEGO Island uses case insensitive file path fragments which need to be mapped to the actual files on disk (or CD).

Tested on Linux:

image


AJenbo (approved):

LEGO1/omni/src/common/mxstring.cpp
```diff
@@ -204,5 +204,30 @@ void MxString::NormalizePath(char* p_path)
...
+					SDL_strlcpy(&p_path[i - 1], file.GetData(), file.GetLength() + 1);
+					SDL_Log("Resolved file path to %s", p_path);
+					return true;
+				}
+			}
+		}
+
+		return false;
+	};
+
```
madebr, line 228: This returns true when a file in the filesystem matches the ending string of a input path.
e.g. `normalize("aBc/dDf/input/FILE.bmp")` will match `"input/file.bmp"` in the file system.

so the output will be `aBc/dDf/input/file.bmp`

Is this intentional?

madebr: PhysFS has a nice API to abstract archive and filesystem accesses.

foxtacles: Yes it's intentional. Some internal paths coming from the `.SI` files have random drive letter prefixes. I believe that is sanitized away in some other part of the code base. I wanted to limit the modification of the path input to only the filesystem relevant bit, and leave the other parts (whatever it is) alone

LEGO1/omni/src/common/mxstring.cpp
```diff
@@ -204,5 +204,30 @@ void MxString::NormalizePath(char* p_path)
...
+		for (const MxString& file : p_files) {
+			// Test whether file is a suffix of the provided path (case insensitive)
+			// If yes, copy the original file system path into p_path.
+			for (size_t i = pathLen, j = file.GetLength(); i != 0 && j != 0; i--, j--) {
+				if (SDL_tolower(p_path[i - 1]) != SDL_tolower(file.GetData()[j - 1])) {
+					break;
+				}
+				else if (j == 1) {
+					SDL_strlcpy(&p_path[i - 1], file.GetData(), file.GetLength() + 1);
+					SDL_Log("Resolved file path to %s", p_path);
```
madebr, line 220: I think might become chatty. Perhaps do it as
```suggestion
					SDL_LogTrace("Resolved file path to %s", p_path);
```
The logging level can be change by setting the `SDL_LOGGING` environment variable.
(`SDL3/SDL_hints.h` documents what value the environment variable expects)

madebr (approved):

PR #101, opened 2025-05-17
Fix initilization and overlapping memcpy

AJenbo:
- This fixes trying to fix a garbage address
- Adds errors/logging/assert to a bit of unimplemented paths
- Fixes an original bug with an overlapping data copy

### DIFF
--- a/miniwin/miniwin/src/miniwin.cpp
+++ b/miniwin/miniwin/src/miniwin.cpp
@@ -1,5 +1,7 @@
 #include "miniwin.h"
 
+#include "miniwin_p.h"
+
 #include <SDL3/SDL.h>
 #include <vector>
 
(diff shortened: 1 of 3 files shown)


foxtacles (approved):

miniwin/miniwin/src/miniwin.cpp
```diff
@@ -21,6 +23,7 @@ ULONG IUnknown::Release()
 
 HRESULT IUnknown::QueryInterface(const GUID& riid, void** ppvObject)
 {
+	SDL_LogError(LOG_CATEGORY_MINIWIN, "IUnknown does not implement guid");
```
madebr, line 26: SDL3 has a `SDL_GUIDToString` function.
Perhaps we should introduce a little error function that prints the guid?

madebr (approved):

miniwin/miniwin/src/miniwin.cpp
AJenbo: That would be useful I ended up making a big list to figure out some of the ones that need implemented to get to this point

PR #102, opened 2025-05-17
Fix bugs preventing game from starting up on Linux

AJenbo:
image

Screenshot from 2025-05-17 04-24-40

### DIFF
--- a/LEGO1/lego/legoomni/src/video/legomodelpresenter.cpp
+++ b/LEGO1/lego/legoomni/src/video/legomodelpresenter.cpp
@@ -97,7 +97,7 @@ MxResult LegoModelPresenter::CreateROI(MxDSChunk* p_chunk)
 		SDL_strlwr(textureName);
 
 		if (textureName[0] == '^') {
-			strcpy(textureName, textureName + 1);
+			memmove(textureName, textureName + 1, strlen(textureName));
 
 			if (g_modelPresenterConfig) {
 				texture = new LegoTexture();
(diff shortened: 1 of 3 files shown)


LEGO1/omni/src/video/mxstillpresenter.cpp
```diff
@@ -55,7 +55,7 @@ void MxStillPresenter::CreateBitmap()
 	m_frameBitmap = new MxBitmap;
 	m_frameBitmap->ImportBitmapInfo(m_bitmapInfo);
 
-	delete m_bitmapInfo;
+	delete[] m_bitmapInfo;
```
foxtacles, line 58: Can you contribute this to original decomp. thank you!

foxtacles (approved):

LEGO1/omni/src/video/mxstillpresenter.cpp
foxtacles: and add a cast to MxU8* before delete. also there's at least one other incorrect delete in the file. thanks

AJenbo: Here or in the other PR?

foxtacles: in the decomp pr

madebr (approved):

27hectormanuel:
Could this be compiled for Steam Deck?

madebr:
> Could this be compiled for Steam Deck?

The Steam Deck is a glorified Linux machine. So anything that runs on Linux, will also run on the Steam Deck.
For isle-portable, we're not yet arrived at the stage of supporting anything.

AJenbo:
@27hectormanuel I'm on Linux so that's what I'm doing basically. Once we get the screen to stop flickering and apply the right colors we still have all the 3D to implement.

PR #115, opened 2025-05-18
Fix remaning UBSAN issues

AJenbo:
This fixes all UB issues i was able to trigger on 32bit Linux with what's working so fare.

### DIFF
--- a/LEGO1/lego/legoomni/src/common/mxcontrolpresenter.cpp
+++ b/LEGO1/lego/legoomni/src/common/mxcontrolpresenter.cpp
@@ -9,6 +9,7 @@
 #include "mxtimer.h"
 #include "mxutilities.h"
 
+#include <SDL3/SDL_log.h>
 #include <SDL3/SDL_stdinc.h>
 #include <assert.h>
 
@@ -78,7 +79,11 @@ void MxControlPresenter::EndAction()
 MxBool MxControlPresenter::FUN_10044270(MxS32 p_x, MxS32 p_y, MxPresenter* p_presenter)
 {
 	assert(p_presenter);
-	MxStillPresenter* presenter = (MxStillPresenter*) p_presenter;
+	MxStillPresenter* presenter = dynamic_cast<MxStillPresenter*>(p_presenter);
+	if (!presenter) {
+		SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "Invalid presenter");
+		return FALSE;
+	}
 
 	if (m_unk0x4c == 3) {
 		MxStillPresenter* map = (MxStillPresenter*) m_list.front();
--- a/LEGO1/omni/src/video/flic.cpp
+++ b/LEGO1/omni/src/video/flic.cpp
@@ -159,9 +159,10 @@ void WritePixelPairs(
 	short is_odd = p_count & 1;
 	p_count >>= 1;
 
-	WORD* dst = (WORD*) (((p_bitmapHeader->biWidth + 3) & -4) * p_row + p_column + p_pixelData);
+	BYTE* dst = ((p_bitmapHeader->biWidth + 3) & -4) * p_row + p_column + p_pixelData;
 	while (--p_count >= 0) {
-		*dst++ = p_pixel;
+		memcpy(dst, &p_pixel, sizeof(WORD));
+		dst += sizeof(WORD);
 	}
 
 	if (is_odd) {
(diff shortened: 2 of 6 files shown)


foxtacles (approved):

madebr:
With [this patch](https://github.com/isledecomp/isle-portable/pull/82#discussion_r2094641755) applied, do you still need to disable the sanitizers for miniaudio?

AJenbo:
> With [this patch](https://github.com/isledecomp/isle-portable/pull/82#discussion_r2094641755) applied, do you still need to disable the sanitizers for miniaudio?

This patch makes UBSanitizer not apply to miniaudio

AJenbo:
Unless I do so it dies here:
```
Resolved file path to .local/share/isledecomp/isle/lego/scripts/sndanim.si
isle-portable/build/_deps/miniaudio-src/miniaudio.h:76905:48: runtime error: member access within null pointer of type 'struct ma_engine'
```
With or without the changes here.

foxtacles:
These errors might be due to incorrect use of miniaudio resources in isle. Once that is resolved we could leave the tools enabled globally.

AJenbo:
Should I update this PR to not disable it?

foxtacles:
In principle I'd say let's try to fix the resource use issues of miniaudio in isle and not disable UBSan. I can do that but I won't get to it until tomorrow at the earliest

AJenbo:
> In principle I'd say let's try to fix the resource use issues of miniaudio in isle and not disable UBSan. I can do that but I won't get to it until tomorrow at the earliest

That would be ideal, I have removed that part of the PR so it still remains globally active.

LEGO1/lego/legoomni/src/common/mxcontrolpresenter.cpp
```diff
@@ -78,7 +78,10 @@ void MxControlPresenter::EndAction()
 MxBool MxControlPresenter::FUN_10044270(MxS32 p_x, MxS32 p_y, MxPresenter* p_presenter)
 {
 	assert(p_presenter);
-	MxStillPresenter* presenter = (MxStillPresenter*) p_presenter;
+	MxStillPresenter* presenter = dynamic_cast<MxStillPresenter*>(p_presenter);
+	if (!presenter) {
+		return FALSE;
```
madebr, line 83: Does this happen often?
As this changes behavior, should we log a message here?

LEGO1/omni/src/video/flic.cpp
```diff
@@ -159,9 +159,10 @@ void WritePixelPairs(
 	short is_odd = p_count & 1;
 	p_count >>= 1;
 
-	WORD* dst = (WORD*) (((p_bitmapHeader->biWidth + 3) & -4) * p_row + p_column + p_pixelData);
+	BYTE* dst = ((p_bitmapHeader->biWidth + 3) & -4) * p_row + p_column + p_pixelData;
 	while (--p_count >= 0) {
-		*dst++ = p_pixel;
+		memcpy(dst, &p_pixel, sizeof(WORD));
```
madebr, line 164: Perhaps also use the unaligned read function here, as big median will also need to swap the words here.

LEGO1/tgl/d3drm/group.cpp
```diff
@@ -89,7 +89,10 @@ Result GroupImpl::Add(const Group* pGroup)
 // FUNCTION: LEGO1 0x100a3430
 Result GroupImpl::Add(const MeshBuilder* pMeshBuilder)
 {
-	const MeshBuilderImpl* pMeshBuilderImpl = static_cast<const MeshBuilderImpl*>(pMeshBuilder);
+	const MeshBuilderImpl* pMeshBuilderImpl = dynamic_cast<const MeshBuilderImpl*>(pMeshBuilder);
+	if (!pMeshBuilderImpl) {
+		return Result::Error;
```
madebr, line 94: Log a message?

LEGO1/lego/legoomni/src/common/mxcontrolpresenter.cpp
foxtacles: Sounds good - I don't think that should happen at all

LEGO1/tgl/d3drm/group.cpp
foxtacles: We could add a small template helper similar to UnalignedRead which does the cast and has a logging (assert?) on failure. I don't think conversion should ever fail

AJenbo: I can add the logging but could you do the cleanups as a follow up?

madebr: That's ok! The extra proposed template function is out-of-scope for this pr.

PR #125, opened 2025-05-19
minwin: Fix bitmap colors

AJenbo:
https://github.com/user-attachments/assets/4852bfd5-0361-402a-9fd1-37682eab63ab


Fancy2209:
Was the mp4 supposed to only be an audio track?

AJenbo:
> Was the mp4 supposed to only be an audio track?

Your browser is having issues:

image

Fancy2209:
Oh I see, my bad!
image
Edit: Seems to work for me now

AJenbo:
> Oh I see, my bad!

You could at least tell me what browser your having issues with. The original video worked fine in Firefox.

> Seems to work for me now

I guess you where using Chrome which seems to be picky, I have uploaded a new one that should work better with that browser.

Fancy2209:
It is Google Chrome, anywho sorry for cluttering the thread so much about this silly issue

madebr (approved):

PR #129, opened 2025-05-19
Let LegoStorage know of memory size

madebr:
It also adds a few assertions for things that should not fail.

Please test!

cc @AJenbo


foxtacles:
I'm not getting any assertion triggers. Of course we can merge this anyway, seems useful.

foxtacles (approved):

PR #132, opened 2025-05-19
Initialize GPU 3D rendere

AJenbo:
This just gets the render working and drawing in the right relation to the 2D surfaces. It need some serious refactoring and currently is hard coded to render a single triangle.

It does bump the game to run in 32bit colors instead of being limited to 8/16bit which is nice, it's mostly done for performance to avoid expensive conversions of the buffer on the CPU.

And even with the limited 3D it's still mostly an improvement I think:
image

The shaders can be found here as well as some useful examples: https://github.com/TheSpydog/SDL_gpu_examples/tree/main/Content/Shaders


madebr:
I rebased this pr on top of current master (so we don't have to worry about miniaudio issues).
I also added precompiled includeable headers.

AJenbo:
Nice, I'll try to improve the structure a little bit.

madebr:
> Nice, I'll try to improve the structure a little bit.

The MSVC c++20 error makes me want to write the code in c++20.
It's not like we need to go crazy and write unreadable template code, but it has some nice features.

foxtacles:
> > Nice, I'll try to improve the structure a little bit.
> 
> The MSVC c++20 error makes me want to make to write the code in c++20. It's not like we need to go crazy and write unreadable template code, but it has some nice features.

Would be great if we could go to C++20. I think right now the game crashes for an unknown reason when targeting the C++17 standard though (according to Anders). So we'd have to solve that first

AJenbo:
> It's not like we need to go crazy and write unreadable template code, but it has some nice features.

:100: DevilutionX is very sparse even in OOP code, it does have some templates but nothing to crazy.

miniwin/miniwin/shaders/generated/ShaderIndex.cpp
```diff
@@ -115,11 +115,12 @@ const SDL_GPUShaderCreateInfo* GetVertexShaderCode(VertexShaderId id, SDL_GPUSha
     SDL_assert(id < SDL_arraysize(VertexShaderMSLCodes));
     return &VertexShaderMSLCodes[id];
   }
-#endif
+#else
```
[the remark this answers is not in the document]
madebr: I kept spirv around for those people wanting to try out MoltenVK.

miniwin/miniwin/src/include/miniwin_d3drmobject_p.h
```diff
@@ -0,0 +1,69 @@
...
+template <typename T>
+struct Direct3DRMObjectBase : public T {
+	ULONG Release() override
+	{
+		if (IUnknown::m_refCount == 1) {
+			for (auto it = m_callbacks.cbegin(); it != m_callbacks.cend(); it++) {
+				it->first(this, it->second);
+			}
+		}
+		return this->T::Release();
```
madebr, line 17:
```suggestion
		SDL_free(m_name);
		return this->T::Release();
```

miniwin/miniwin/src/miniwin_d3drmdevice.cpp
```diff
@@ -0,0 +1,118 @@
...
+#include "miniwin_d3drm_p.h"
+#include "miniwin_d3drmobject_p.h"
+#include "miniwin_d3drmviewport_p.h"
+
+#include <SDL3/SDL.h>
+
+Direct3DRMDevice2Impl::Direct3DRMDevice2Impl(DWORD width, DWORD height, SDL_GPUDevice* device)
+	: m_width(width), m_height(height), m_device(device), m_viewports(new Direct3DRMViewportArrayImpl)
+{
+	m_viewports->AddRef();
```
madebr, line 11:
```suggestion
```
Don't we initialize with a ref count of 1?

miniwin/miniwin/src/miniwin_d3drmviewport.cpp
```diff
@@ -0,0 +1,258 @@
...
+	transferInfo.transfer_buffer = m_downloadTransferBuffer;
+	transferInfo.offset = 0;
+	SDL_DownloadFromGPUTexture(copyPass, &region, &transferInfo);
+	SDL_EndGPUCopyPass(copyPass);
+	SDL_GPUFence* fence = SDL_SubmitGPUCommandBufferAndAcquireFence(cmdbuf);
+	if (!cmdbuf || !SDL_WaitForGPUFences(m_device, true, &fence, 1)) {
+		return DDERR_GENERIC;
+	}
+	SDL_ReleaseGPUFence(m_device, fence);
+	void* downloadedData = SDL_MapGPUTransferBuffer(m_device, m_downloadTransferBuffer, false);
```
madebr, line 119: You need to `SDL_UnmapGPUTransferBuffer` the buffer after you're done.

miniwin/miniwin/src/miniwin_d3drmviewport.cpp
```diff
@@ -0,0 +1,258 @@
...
+	if (!downloadedData) {
+		return DDERR_GENERIC;
+	}
+
+	m_updated = false;
+
+	SDL_free(m_renderedImage);
```
madebr, line 126:
```suggestion
	SDL_DestroySurface(m_renderedImage);
```

madebr:
Perhaps try applying this patch for c++ 20?
```patch
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,6 @@ include(CheckCXXSourceCompiles)
 include(CMakeDependentOption)
 include(CMakePushCheckState)
 
-set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
@@ -146,6 +145,7 @@ target_include_directories(Vec::Vec INTERFACE "${CMAKE_SOURCE_DIR}/3rdparty/vec"
 add_library(lego1 SHARED
   LEGO1/main.cpp
 )
+target_compile_features(lego1 PUBLIC cxx_std_20)
 target_precompile_headers(lego1 PRIVATE "LEGO1/lego1_pch.h")
 set_property(TARGET lego1 PROPERTY DEFINE_SYMBOL "LEGO1_DLL")
 target_include_directories(lego1 PUBLIC "$<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/util>")
```

AJenbo:
> Perhaps try applying this patch for c++ 20?

I already dropped that syntax to c++17

madebr:
Should we drop the `-k` from ci to have faster feedback?

AJenbo:
It's super hard to find the error because it keeps going, so i might drop i just for that reason.

madebr:
> It's super hard to find the error because it keeps going, so i might drop i just for that reason.

I think it's fine to keep it removed. I don't use it in any other projects.
It served its purpose imho.

AJenbo:
Things are working now. I think this PR is ok to go in, the triangle might not be how the game originally looked, you could zero out the vertex array but I think for now it serves as good feedback that things are working so some odd visuals until the rest of the D3DRM might makes sens/be expected.

PR #135, opened 2025-05-20
Move to C++17

### LINKED ISSUE #112: Building with C++17 standard breaks game

AJenbo:
Fixes #112

This also solves a build error for C++20

### DIFF
--- a/LEGO1/realtime/matrix.h
+++ b/LEGO1/realtime/matrix.h
@@ -14,9 +14,6 @@ struct UnknownMatrixType {
 // VTABLE: BETA10 0x101b8340
 // SIZE 0x08
 class Matrix4 {
-public:
-	virtual ~Matrix4() {}
-
 protected:
 	float (*m_data)[4];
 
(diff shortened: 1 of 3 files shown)


madebr (approved):

LEGO1/realtime/matrix.h
```diff
@@ -14,6 +14,9 @@ struct UnknownMatrixType {
 // VTABLE: BETA10 0x101b8340
 // SIZE 0x08
 class Matrix4 {
+public:
+	virtual ~Matrix4() {}
```
foxtacles, line 18: Interesting. Can you explain why we need this? Do we need that for the vector class as well?

AJenbo: Because MxMatrix extends Matrix4, you can assign a MxMatrix to a variable of type  Matrix4, but delete will call the Matrix4 destructor unless it's virtual.

foxtacles: Yes that part I'm aware of. What I'm confused by is how this error only happens with C++17. You are not getting it on C++11?

AJenbo: The alternative would be to change m_unk0x78 to be a `MxMatrix*` this would also work since it's the only type ever assigned to it.

AJenbo: > Yes that part I'm aware of. What I'm confused by is how this error only happens with C++17. You are not getting it on C++11?

Afaik it's undefined behaviour in C++11 but not allowed in C++17 and ASAN thus enables stricter checks for it by default when build for C++17.

PR #138, opened 2025-05-20
Show imgui window when starting app with --debug

madebr:
When started with `--debug`, isle will open a second window with imgui windows.

This will hopefully help in understanding the various unknown variables, while the game is running.
To see what's possible, I've added some basic things to a window.

Things that are making this hard are:
- private/protected class visibility don't allow the debug renderer to see inside classes. Perhaps we should add a "debug view" to classes and trees to make its iternals show.
- DLL visibility. The DLL exports a limited number of symbols. In this pr, I've added 2 `LEGO1_EXPORT` to make it link on Windows. This problem is not a possible blocker, as the previous one. But it is annoying.

The imgui window is currently shown in an external window.
Once sdlgpu is used, we can probably do everything in one window.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,7 @@ option(ISLE_BUILD_APP "Build isle application" ON)
 option(ISLE_ASAN "Enable Address Sanitizer" OFF)
 option(ISLE_UBSAN "Enable Undefined Behavior Sanitizer" OFF)
 option(ISLE_WERROR "Treat warnings as errors" OFF)
+option(ISLE_DEBUG "Enable imgui debug" ON)
 cmake_dependent_option(ISLE_USE_DX5 "Build with internal DirectX 5 SDK" "${NOT_MINGW}" "WIN32;CMAKE_SIZEOF_VOID_P EQUAL 4" OFF)
 cmake_dependent_option(ISLE_MINIWIN "Use miniwin and minimfc" ON "NOT ISLE_USE_DX5" OFF)
 cmake_dependent_option(ISLE_BUILD_CONFIG "Build CONFIG.EXE application" ON "MSVC OR ISLE_MINIWIN" OFF)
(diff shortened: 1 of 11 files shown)


ISLE/isleapp.cpp
```diff
@@ -287,9 +286,6 @@ SDL_AppResult SDL_AppInit(void** appstate, int argc, char** argv)
 
 SDL_AppResult SDL_AppIterate(void* appstate)
 {
-	if (g_closed) {
```
foxtacles, line 290: Are we sure we don't need this?

madebr: It is needed. I accidentally removed in a late refactor where I extracted debugapp from the IsleApp class into its own file.

CMakeLists.txt
```diff
@@ -467,6 +468,7 @@ if (ISLE_BUILD_APP)
   add_executable(isle WIN32
     ISLE/res/isle.rc
     ISLE/isleapp.cpp
+    ISLE/isledebug.cpp
```
foxtacles, line 471: Would it make sense to add a conditional/flag to either include or exclude this (together with IMGUI)?

madebr: If you don't run with `--debug`, then you won't notice it.

foxtacles: Yes, but IMGUI and isledebug would still be part of the executable, right? Seems like significant overhead (`ISLE.EXE` 1.8MB with debug, vs 98kb without on my win32 debug build)

madebr: Oh, yes. Good point. I'll make it an option (default ON?) then.

foxtacles:
Thank you, this is very useful. Here's a screenshot for the curious:

image

foxtacles (approved):

CMakeLists.txt
madebr: There is `ISLE_DEBUG` now, which defaults to `ON`.
Once we understand every bit, we can change this default ;)

PR #139, opened 2025-05-20
Remove `dynamic_cast` to `LegoLocomotionAnimPresenter`

foxtacles:
The underlying type of the element that the list accepts is `LegoAnimPresenter*`. In the decomp, the casts never fail (since they are `static_cast`'s). The only reason we added the casts in the decomp is because we wouldn't otherwise get the correct assembly generation. However, the behavior should remain unchanged so instead of replacing with a `dynamic_cast`, we do no cast.

The root problem is that we have trouble figuring out the type of the `m_presenters` list in the decomp. Different places in the code base imply different types based on the assembly. It might simply be the result of a refactor in the original source that hasn't been fully completed.


AJenbo:
Nice, not having to cast is the best option :)

Issue #123, opened 2025-05-19
No clear error message when some files are missing


AJenbo:
When some of the required files are missing you get various crashes instead of a clear error message saying that a nessesery file was not found.

A similar issue with the app not being able to find nocd was already solved and similar checks for the remaining files would probably be good since the game will likely not simply read from a physical disk but will require the user to place all needed files in a specific folder.

PR #147, opened 2025-05-21
Add message box errors for missing files

foxtacles:


foxtacles:
The only alternative I can think of is to turn these into logs. There is no clean way to abort loading the game without having to rewrite a bunch of stuff.

madebr:
> The only alternative I can think of is to turn these into logs. There is no clean way to abort loading the game without having to rewrite a bunch of stuff.

Perhaps isle can test existence of a few key files, and show a dialog when it cannot find {jukebox,isle}.si nor nocd.si?
If it can find {jukebox,isle}.si, then start the game.
Else, if it can find nocde.si, then start the game
Else, show a fatal error?

If a file fails to load at a later stage, then I'm fine with the current behavior. Not ideal, but better then a crash and a bug report.

foxtacles:
> > The only alternative I can think of is to turn these into logs. There is no clean way to abort loading the game without having to rewrite a bunch of stuff.
> 
> Perhaps isle can test existence of a few key files, and show a dialog when it cannot find {jukebox,isle}.si nor nocd.si? If it can find {jukebox,isle}.si, then start the game. Else, if it can find nocde.si, then start the game Else, show a fatal error?

We essentially already have this logic here (except for `jukebox`):

https://github.com/isledecomp/isle-portable/blob/master/ISLE/isleapp.cpp#L781

My understanding was that this [isn't enough](https://github.com/isledecomp/isle-portable/issues/123) since other dependent files loaded by the SI asset packs may still be missing (or `world.wdb`, for instance). I'd rather not hardcode a bunch of random checks for "critical files" if we can avoid it.

Personally I don't think we necessarily need more handling in the game itself for this. We could build a small "installer" utility/script that will prepare a filesystem and ensure files are present based on a ISO/CD, since presumably nobody will want to use (or be able to) use the original Windows installer.

madebr:
> We essentially already have this logic here (except for `jukebox`):
> 
> https://github.com/isledecomp/isle-portable/blob/master/ISLE/isleapp.cpp#L781
> 
> My understanding was that this [isn't enough](https://github.com/isledecomp/isle-portable/issues/123) since other dependent files loaded by the SI asset packs may still be missing (or `world.wdb`, for instance). I'd rather not hardcode a bunch of random checks for "critical files" if we can avoid it.

I get my first error dialog when isleapp is inside `LegoOmni::GetInstance()->CreateBackgroundAudio()`, before the nocd check:
https://github.com/isledecomp/isle-portable/blob/92ee03ec4d73c9757548805ea2e50730f5949f7d/ISLE/isleapp.cpp#L773-L781

> Personally I don't think we necessarily need more handling in the game itself for this. We could build a small "installer" utility/script that will prepare a filesystem and ensure files are present based on a ISO/CD, since presumably nobody will want to use (or be able to) use the original Windows installer.

People (me) will always want to run the game from where they compiled it.

foxtacles:
> People (me) will always want to run the game from where they compiled it.

What I was trying to say is that I think we should address the root cause. From what I gathered the main reason for these failures is that people have been trying to run the game directly by pointing at the ISO/CD. That won't work to my knowledge.  You will need to manually "install" the game by copying some files around so everything is in its expected place.

I'm fairly sure we will need a utility to address this down the road, and it would at the same time solve the quiet failures of missing files, for all files. Checking only for a few "critical" files feels like kicking the can down the road: if any of the dependent files are missing, you'll get the same obscure failures again, which will certainly happen to *someone*. So I'd rather solve it cleanly once and for all.

That being said, if you see enough value at this point in just adding checks for a couple files, I'd be ok with that. Just trying to be clear that I don't think it will save us from future reports of random failures (crashes and bug reports, as you mentioned) due to other missing files.

madebr:
Looks like a good idea!

Perhaps create an issue to discuss how to handle a (mini) installer?
We can perhaps integrate `config.exe` to generate an initial ini.

foxtacles:
I've opened an issue here: https://github.com/isledecomp/isle-portable/issues/148

Closing this PR now - happy to accept a patch for "minimal" sanity checks for initial files as part of isleapp, as you mentioned. Once we have the utility we can remove those bits.

madebr:
I was able to recreate my previously broken install, and this pr does emit a warning about missing `world.wdb`, but then continues and fails the assertion I added to `LegoMemory::Read`.

It's very simple to recreate:
- Extract/mount the LEGO Island cd somewhere
- let `diskpath`, `cdpath` and `mediapath` in the ini point to the root of the cd

madebr:
@foxtacles 
I think I created a nice behavior in this branch, based on your code:
- warnings about missing files are shown only once before the nocd animation (correctly!) plays
- when in a broken install, there is a warning about missing world.wdb and then a final fatal error

https://github.com/madebr/isle-portable/tree/missing-file-message-box-errors

Perhaps you can take over and re-open this pr based on my branch?

There's a single issue with it: I added legoomni code to mxomni, which I believe is a no-no.

foxtacles:
Thanks I'll take a look

> There's a single issue with it: I added legoomni code to mxomni, which I believe is a no-no.

Yeah, let's try not to do that

PR #149, opened 2025-05-22
Sky color controls

AJenbo:
https://github.com/user-attachments/assets/7420b710-a9bb-4ba5-8a52-f5a6057cd287

This implements a few of the color functions, cleans up the vertex uploading a bit and corrects the flow for how the scene is gathered.

This is enough that we get rid of the HoM effect and the sky color can now be control from the panel in the center.

### DIFF
--- a/miniwin/miniwin/src/include/miniwin_d3drmframe_p.h
+++ b/miniwin/miniwin/src/include/miniwin_d3drmframe_p.h
@@ -28,6 +28,8 @@ struct Direct3DRMFrameImpl : public Direct3DRMObjectBase<IDirect3DRMFrame2> {
 	HRESULT SetMaterialMode(D3DRMMATERIALMODE mode) override;
 	HRESULT GetChildren(IDirect3DRMFrameArray** children) override;
 
+	D3DCOLOR m_backgroundColor = 0xFFFFFFFF;
+
 private:
 	IDirect3DRMFrameArray* m_children;
 	IDirect3DRMLightArray* m_lights;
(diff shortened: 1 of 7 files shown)


miniwin/miniwin/src/include/miniwin_d3drmframe_p.h
```diff
@@ -28,6 +28,8 @@ struct Direct3DRMFrameImpl : public Direct3DRMObjectBase<IDirect3DRMFrame2> {
 	HRESULT SetMaterialMode(D3DRMMATERIALMODE mode) override;
 	HRESULT GetChildren(IDirect3DRMFrameArray** children) override;
 
+	D3DCOLOR m_backgroundColor = 0xFFFFFFFF;
```
madebr, line 31: Is there a reason you set the default background color of a frame to white, and that of a viewport to black?

When I tried some simple d3drm examples, they had black backgrounds.

miniwin/miniwin/src/include/miniwin_d3drmviewport_p.h
```diff
@@ -5,6 +5,7 @@
 #include "miniwin_d3drmobject_p.h"
 
 #include <SDL3/SDL.h>
+#include <span>
```
madebr, line 8: I think this is unused.
```suggestion
```

 Perhaps you tried it for `PushVertices`?

miniwin/miniwin/src/include/miniwin_d3drmframe_p.h
AJenbo: Oh, that's my bad i wanted to set it as black actually. I guess that just shows it never actually rendered an uninitialized scene.

AJenbo:
```suggestion
	D3DCOLOR m_backgroundColor = 0xFF000000;
```

miniwin/miniwin/src/include/miniwin_d3drmviewport_p.h
AJenbo: Yeah, it needed C++20

madebr (approved):

PR #151, opened 2025-05-24
Fixup LegoOmni::World LegoOmni::GetWorldId()

kernaltrap8:
Hopefully this is fine... No idea if its still matching but the previous impl relies on UB and I *heavily* doubt that was anything like the original code that Mindscape wrote.


foxtacles:
Thank you. I've looked into the decomp again and fixed the code: https://github.com/isledecomp/isle/commit/a860e76dbac0e1860486397cb63082432a8fbb37 This still matches and isn't UB.

PR #161, opened 2025-05-24
Add D3DRMFrame parent <-> child relation

madebr:
/cc @AJenbo

### DIFF
--- a/miniwin/miniwin/src/include/miniwin_d3drm_p.h
+++ b/miniwin/miniwin/src/include/miniwin_d3drm_p.h
@@ -11,33 +11,37 @@ typedef struct PositionColorVertex {
 	Uint8 r, g, b, a;
 } PositionColorVertex;
 
-template <typename InterfaceType, typename ArrayInterface>
+template <typename InterfaceType, typename ActualType, typename ArrayInterface>
 class Direct3DRMArrayBase : public ArrayInterface {
 public:
 	~Direct3DRMArrayBase() override
 	{
-		for (auto* item : items) {
+		for (auto* item : m_items) {
 			if (item) {
 				item->Release();
 			}
 		}
 	}
-	DWORD GetSize() override { return static_cast<DWORD>(items.size()); }
+	DWORD GetSize() override { return static_cast<DWORD>(m_items.size()); }
 	HRESULT AddElement(InterfaceType* in) override
 	{
 		if (!in) {
 			return DDERR_INVALIDPARAMS;
 		}
-		in->AddRef();
-		items.push_back(in);
+		auto inImpl = static_cast<ActualType*>(in);
+		if (!inImpl) {
+			return DDERR_INVALIDPARAMS;
+		}
+		inImpl->AddRef();
+		m_items.push_back(inImpl);
 		return DD_OK;
 	}
 	HRESULT GetElement(DWORD index, InterfaceType** out) override
 	{
-		if (index >= items.size()) {
+		if (index >= m_items.size()) {
 			return DDERR_INVALIDPARAMS;
 		}
-		*out = static_cast<InterfaceType*>(items[index]);
+		*out = static_cast<InterfaceType*>(m_items[index]);
 		if (*out) {
 			(*out)->AddRef();
 		}
@@ -45,32 +49,16 @@ public:
 	}
 	HRESULT DeleteElement(InterfaceType* element) override
 	{
-		auto it = std::find(items.begin(), items.end(), element);
-		if (it == items.end()) {
+		auto it = std::find(m_items.begin(), m_items.end(), element);
+		if (it == m_items.end()) {
 			return DDERR_INVALIDPARAMS;
 		}
 
 		(*it)->Release();
-		items.erase(it);
+		m_items.erase(it);
 		return DD_OK;
 	}
 
 protected:
-	std::vector<InterfaceType*> items;
-};
-
-struct Direct3DRMFrameArrayImpl : public Direct3DRMArrayBase<IDirect3DRMFrame, IDirect3DRMFrameArray> {
-	using Direct3DRMArrayBase::Direct3DRMArrayBase;
-};
-
-struct Direct3DRMLightArrayImpl : public Direct3DRMArrayBase<IDirect3DRMLight, IDirect3DRMLightArray> {
-	using Direct3DRMArrayBase::Direct3DRMArrayBase;
-};
-
-struct Direct3DRMViewportArrayImpl : public Direct3DRMArrayBase<IDirect3DRMViewport, IDirect3DRMViewportArray> {
-	using Direct3DRMArrayBase::Direct3DRMArrayBase;
-};
-
-struct Direct3DRMVisualArrayImpl : public Direct3DRMArrayBase<IDirect3DRMVisual, IDirect3DRMVisualArray> {
-	using Direct3DRMArrayBase::Direct3DRMArrayBase;
+	std::vector<ActualType*> m_items;
 };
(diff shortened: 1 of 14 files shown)


miniwin/miniwin/src/miniwin_d3drmviewport.cpp
```diff
@@ -86,13 +76,22 @@ void Direct3DRMViewportImpl::PushVertices(const PositionColorVertex* vertices, s
...
+		return DDERR_GENERIC;
+	}
 
-	CollectSceneData(group);
+	m_backgroundColor = frameImpl->m_backgroundColor;
+
+	std::vector<PositionColorVertex> vertices =
+		{{-1, -1, 0, 0, 255, 0, 255}, {1, -1, 0, 0, 0, 255, 255}, {0, 1, 0, 255, 0, 0, 128}};
+
+	PushVertices(vertices.data(), vertices.size());
```
AJenbo, line 94: I wanted to keep this in a separate method to avoid a big hard to follow function ad more parts of the render process gets implemented.

miniwin/miniwin/src/miniwin_d3drmframe.cpp
```diff
@@ -101,11 +143,18 @@ HRESULT Direct3DRMFrameImpl::GetVisuals(IDirect3DRMVisualArray** visuals)
 
 HRESULT Direct3DRMFrameImpl::SetTexture(IDirect3DRMTexture* texture)
 {
+	if (!texture) {
+		return DDERR_GENERIC;
+	}
+	auto textureImpl = static_cast<Direct3DRMTextureImpl*>(texture);
+	if (!textureImpl) {
+		return DDERR_GENERIC;
+	}
```
AJenbo, line 152: Not having this check means we get nice ASAN errors when it's used incorrectly, This is helful since the return code is often not checked by the engine.

miniwin/miniwin/src/miniwin_d3drmviewport.cpp
madebr: I'll revert this.
Are we going to upload all geometry to the gpu on every render?

miniwin/miniwin/src/miniwin_d3drmframe.cpp
madebr: This test will never fail, since we're using `static_cast` instead of `dynamic_cast`. (I originally used `dynamic_cast`)
So I think it can be removed.

madebr, line 152:
```suggestion
```

miniwin/miniwin/src/miniwin_d3drmviewport.cpp
```diff
@@ -29,16 +29,6 @@ Direct3DRMViewportImpl::~Direct3DRMViewportImpl()
...
-void Direct3DRMViewportImpl::CollectSceneData(IDirect3DRMFrame* group)
-{
-	m_backgroundColor = static_cast<Direct3DRMFrameImpl*>(group)->m_backgroundColor;
-
-	std::vector<PositionColorVertex> vertices =
-		{{-1, -1, 0, 0, 255, 0, 255}, {1, -1, 0, 0, 0, 255, 255}, {0, 1, 0, 255, 0, 0, 128}};
-
-	PushVertices(vertices.data(), vertices.size());
-}
-
```
madebr, line 41:
```suggestion
void Direct3DRMViewportImpl::CollectSceneData(IDirect3DRMFrame* group)
{
	m_backgroundColor = static_cast<Direct3DRMFrameImpl*>(group)->m_backgroundColor;

	std::vector<PositionColorVertex> vertices =
		{{-1, -1, 0, 0, 255, 0, 255}, {1, -1, 0, 0, 0, 255, 255}, {0, 1, 0, 255, 0, 0, 128}};

	PushVertices(vertices.data(), vertices.size());
}

```

miniwin/miniwin/src/miniwin_d3drmviewport.cpp
madebr, line 94:
```suggestion
	CollectSceneData(group);
```

AJenbo: Hopefully we can figure out how to avoid that, but for now I don't know how else to do it.

miniwin/miniwin/src/include/miniwin_d3drmviewport_p.h
```diff
@@ -58,4 +60,12 @@ struct Direct3DRMViewportImpl : public Direct3DRMObjectBase<IDirect3DRMViewport>
 	SDL_GPUTransferBuffer* m_downloadTransferBuffer;
 	SDL_GPUBuffer* m_vertexBuffer = nullptr;
 	SDL_Surface* m_renderedImage = nullptr;
+	D3DVALUE m_front = .2f; // FIXME: unknown default
+	D3DVALUE m_back = .8f;  // FIXME: unknown default
```
AJenbo, line 64:
```suggestion
	D3DVALUE m_front = 1.f;
	D3DVALUE m_back = 10.f;
```

miniwin/miniwin/src/miniwin_d3drmviewport.cpp
madebr: Makes sense :)

miniwin/miniwin/src/miniwin_d3drmviewport.cpp
AJenbo: I guess now that my PR has been merged you can refactor it in a way you think makes sens since it's a bit more clear what I was aiming for.

madebr: My goal here was to create a geometry buffer for every frame, such that `PushVertices/CollectSceneData` would be necessary anymore.
But I ran into an issue about where to store this data. The api makes it possible to have multiple devices, to which you can all render a frame. So I'd have to move the buffer to the device. This can probably be implemented through smart data structures, but I did not think it through yet.

miniwin/miniwin/src/include/miniwin_d3drm_p.h
```diff
@@ -11,66 +11,54 @@ typedef struct PositionColorVertex {
...
 	{
 		if (!in) {
 			return DDERR_INVALIDPARAMS;
 		}
-		in->AddRef();
-		items.push_back(in);
+		auto inImpl = static_cast<ActualType*>(in);
+		if (!inImpl) {
+			return DDERR_INVALIDPARAMS;
+		}
```
madebr, line 34:
```suggestion
```

miniwin/miniwin/src/miniwin_d3drmframe.cpp
```diff
@@ -37,12 +46,32 @@ HRESULT Direct3DRMFrameImpl::QueryInterface(const GUID& riid, void** ppvObject)
 
 HRESULT Direct3DRMFrameImpl::AddChild(IDirect3DRMFrame* child)
 {
+	Direct3DRMFrameImpl* childImpl = static_cast<Direct3DRMFrameImpl*>(child);
+	if (!childImpl) {
+		return DDERR_GENERIC;
+	}
```
madebr, line 52:
```suggestion
	if (!child) {
		return DDERR_GENERIC;
	}
	Direct3DRMFrameImpl* childImpl = static_cast<Direct3DRMFrameImpl*>(child);
```

miniwin/miniwin/src/miniwin_d3drmframe.cpp
```diff
@@ -131,25 +177,28 @@ HRESULT Direct3DRMFrameImpl::GetTexture(IDirect3DRMTexture** texture)
...
-	MINIWIN_NOT_IMPLEMENTED();
+	m_color = c;
 	return DD_OK;
 }
 
 HRESULT Direct3DRMFrameImpl::SetColorRGB(float r, float g, float b)
 {
-	MINIWIN_NOT_IMPLEMENTED();
+	SetColor(r, g, b, 1.f);
 	return DD_OK;
```
madebr, line 202:
```suggestion
	return SetColor(r, g, b, 1.f);
```

Issue #152, opened 2025-05-24
Allow build without 3D graphics


foxtacles:
When a 3D backend isn't available for the current platform(i.e. Emscripten, WSL2 setups), we should allow a build of the game without it

AJenbo:
It seems to me this should be an explicit build flag that needs to be set manually as it will by definition product a broken game, so it's more a type of help for developers to focus on bringing up a sub section of the game then something that allows it to work on more platforms.

`cmake .. -DISLE_NO3D=ON`

madebr:
Or make this a runtime option?

foxtacles:
Either would be fine imo

AJenbo:
I guess we could have a dummy rendere that does no-op on the input. I don't see where in the ini the rendere is configured?

madebr:
Just a warning, I'm moving the miniwin around "agressively". Please don't do big code changes while I'm working on this :)

AJenbo:
This is coming a bit later I'm working on things currently on my laptop, but I'm out driving atm

wyattthewyattman:
> When a 3D backend isn't available for the current platform(i.e. Emscripten, WSL2 setups), we should allow a build of the game without it

wait this game is gonna be on WebAssembly? How would that even be possible for 3D rendering of the game? (I've only ever known of Emscripten stuff in 2D)

AJenbo:
https://en.wikipedia.org/wiki/WebGL
https://en.wikipedia.org/wiki/WebGPU

foxtacles:
Since we have software rendering now this is obsolete

PR #170, opened 2025-05-25
Move miniwin to sdl3gpu subdirectory + add suffix classes

madebr:
The suffix allows multiple implementations to live in parallel.

By running `Miniwin_ConfigureBackend`, you can select an alternative backend.

I like to be positive and think somebody will create a miniwin implementation, using opengl or some alternative 
rendering api.
To avoid name clashes, I added a `_SDL3GPUImpl` suffix to (almost?) all sdl3gpu classes.

Stepping stone toward #152

/cc @AJenbo


AJenbo:
This seems a bit of lumping everything together, I don't think 2d and 3d should be combined like that

AJenbo:
... Oh merged as I was typing

madebr:
> This seems a bit of lumping everything together, I don't think 2d and 3d should be combined like that

Yeah, I see this pr as a first step to make things more modular. It makes sense for another backend to not use the SDL_Renderer at all.

madebr:
I was also thinking, if we make the miniwin types/enums/values 100% compatible with the original ones, I think I can drop the CMake `ISLE_USE_DX5` option.
By adding a d3drm backend, we'd be able to support both miniwin@sdl3gpu and an original d3drm dll with the same build.

AJenbo:
A lot of the code that was moved isn't SDL specific.  All the 2d rendering would work fine with a different 3d back end.

The driver for splitting this was so systems the can run the current implementation could still run the 2d implementation.

If you're going to move everything wholesale does that mean all the event code will also be moved to this sub folder?

SDL_rendere isn't being used in the current code, maybe your mixing it up with SDL_gpu which I guess explains this approach.

AJenbo:
It would make more sens to break out the sdl_gpu part as a separate device implementation that d3drm can use and permit all the common code to work with a different implementation. The world means we can also use the existing code for picking a functional render rather then adding miniwin specific ifdef to the code

AJenbo:
If we lump all 2d and 3d under one implementation users with no sdl_gpu support won't have any 2d either which seems to largely defeat the purpose

madebr:
You make good points, but I'd first like to see another backend implemented and decide later what we can make public.

AJenbo:
I wonder say that the large refactoring should happen when someone is actually implementing another backend

madebr:
> I wonder say that the large refactoring should happen when someone is actually implementing another backend

Same, I'm currently trying something experimental which will use this pr.
If it pans out, we can think again about what and how to refactor.

miniwin/sdl3gpu/src/miniwin_ddraw.cpp
```diff
@@ -54,11 +54,11 @@ HRESULT DirectDrawImpl::CreatePalette(
 		return DDERR_INVALIDPARAMS;
 	}
 
-	*lplpDDPalette = static_cast<LPDIRECTDRAWPALETTE>(new DirectDrawPaletteImpl(lpColorTable));
+	*lplpDDPalette = static_cast<LPDIRECTDRAWPALETTE>(new DirectDrawPalette_SDL3GPUImpl(lpColorTable));
 	return DD_OK;
 }
 
-HRESULT DirectDrawImpl::CreateSurface(
+HRESULT DirectDraw_SDL3GPUImpl::CreateSurface(
```
AJenbo, line 61: This names are just really misleading, it's a SDL_Surface implementation and the goal is to move it to SDL_Render / SDL_Texture.

madebr: True. Since this class is member of the sdl3gpu backend, I did a simple `s/Impl/_SDL3GPUIml/` substitution.
It made sense for all classes of a backend to have the same suffix.

This is just naming. As long as we don't have name clashes.

PR #171, opened 2025-05-25
Set viewport frustrum again

AJenbo:
Looks like a conflict got resolved incorrectly


AJenbo:
@madebr can you please resolve this so I don't have to fix it a third time

madebr:
No problem! Will do!

PR #176, opened 2025-05-26
Correct FOV calculation

AJenbo:
This makes the projection matrix match the original calculations which appears to have been a bit different then what's common. Basically instead of contracting on the Y axis it stretches the X, but the format also means that most values are pre-calculated and we can simplify function.

Since mine was done from scratch it's not based on the previous calculations so there might be a bit more diff then strictly needed, hope that's ok. 

image

image

image

### DIFF
--- a/miniwin/sdl3gpu/src/miniwin_d3drmviewport.cpp
+++ b/miniwin/sdl3gpu/src/miniwin_d3drmviewport.cpp
@@ -64,13 +64,13 @@ static void D3DRMMatrixInvertOrthogonal(D3DRMMATRIX4D out, const D3DRMMATRIX4D m
 static void CalculateProjectionMatrix(D3DRMMATRIX4D Result, float field, float aspect, float near, float far)
 {
 	float f = near / field;
-	float debth = far - near;
+	float depth = far - near;
 
 	D3DRMMATRIX4D perspective = {
 		{f, 0, 0, 0},
 		{0, f * aspect, 0, 0},
-		{0, 0, far / debth, 1},
-		{0, 0, (-near * far) / debth, 0},
+		{0, 0, far / depth, 1},
+		{0, 0, (-near * far) / depth, 0},
 	};
 
 	memcpy(Result, &perspective, sizeof(D3DRMMATRIX4D));


miniwin/sdl3gpu/src/miniwin_d3drmviewport.cpp
```diff
@@ -208,10 +202,8 @@ HRESULT Direct3DRMViewport_SDL3GPUImpl::CollectSceneData(IDirect3DRMFrame* group
 	recurseFrame(group, identity);
 
 	PushVertices(verts.data(), verts.size());
-	HMM_Perspective_LH_NO(m_uniforms.perspective, m_field, 4.f / 3.f, m_front, m_back);
 
-	// SDL_Log("FOV: %f", m_field);
-	HMM_Perspective_LH_NO(m_uniforms.perspective, m_field * SDL_PI_F * 4, 4.f / 3.f, m_front, m_back);
+	CalculateProjectionMatrix(m_uniforms.perspective, m_field, (float) m_width / (float) m_height, m_front, m_back);
```
AllMeatball, line 206: I do like width / height solution. this solution better than what I did.

AllMeatball:
You should probably commit the compiled shaders if you can.

miniwin/sdl3gpu/src/miniwin_d3drmviewport.cpp
```diff
@@ -61,25 +61,19 @@ static void D3DRMMatrixInvertOrthogonal(D3DRMMATRIX4D out, const D3DRMMATRIX4D m
...
-	Result[2][2] = (Near + Far) / (Near - Far);
-	Result[3][2] = (2.0f * Near * Far) / (Near - Far);
+	float f = near / field;
+	float debth = far - near;
+
+	D3DRMMATRIX4D perspective = {
+		{f, 0, 0, 0},
+		{0, f * aspect, 0, 0},
+		{0, 0, far / debth, 1},
+		{0, 0, (-near * far) / debth, 0},
```
madebr, line 73: Assuming you meant depth
```suggestion
	float depth = far - near;

	D3DRMMATRIX4D perspective = {
		{f, 0, 0, 0},
		{0, f * aspect, 0, 0},
		{0, 0, far / depth, 1},
		{0, 0, (-near * far) / depth, 0},
```

miniwin/sdl3gpu/src/miniwin_d3drmviewport.cpp
AJenbo: Hopefully it can support ultra wide screen in the future :)

miniwin/sdl3gpu/src/miniwin_d3drmviewport.cpp
AJenbo: Thanks spelling is not my strong suit :)

madebr (approved):

PR #177, opened 2025-05-26
Fix 'warning: implicit truncation of vector type [-Wconversion]'

madebr:
/cc @AllMeatBall


AllMeatball:
I probably shouldn't have ignored the warning. this is a good thing to fix, best not to have warnings if don't have to.

madebr:
Overruled by #176

PR #178, opened 2025-05-26
Move DirectDraw entry functions to backends

madebr:
The DirectDraw functions might conflict when implementing an alternative miniwin backend.
Also add a suffix to global variables DDWindow and DDBackBuffer.

/cc @AJenbo


foxtacles (approved):

AJenbo:
All of the DirectDraw implementations where done before we chose to use SDL_GPU for the 3D rendering so I really don't have to see why these are now getting married as if they go together. The 2D rendering code is not hard coupled to the 3D or voice versa except for the two mentioned variables with is a bit of a hack but another 3D render should be able to do the same to pump it's results back in to the shared buffer..

AJenbo:
I suppose this PR actually makes things better, but the DDBackBuffer buffer is owned by the DirectDraw portion of the code, not the SDL3_GPU portion. Maybe it could just be called DDBackBuffer_SDL3 instead?

AJenbo:
It makes as little sens as calling this event_SDL3GPU:

https://github.com/isledecomp/isle-portable/blob/e1cddc5c0ec43c1a4dc036f635aa652f971a6941/LEGO1/lego/legoomni/src/common/legoutils.cpp#L585

The 2D and 3D code should be kept separate. That was my log term goals anynway. There is nothing implementation specific or hardware requirements about the 2D rendering code.

I would really appreciate if you moved miniwin_ddpalette.cpp miniwin_ddraw.cpp miniwin_ddsurface.cpp back out of the SDL3GPU folder.

madebr:
> The 2D and 3D code should be kept separate. That was my log term goals anyway. There is nothing implementation specific or hardware requirements about the 2D rendering code.
> 
> I would really appreciate if you moved miniwin_ddpalette.cpp miniwin_ddraw.cpp miniwin_ddsurface.cpp back out of the SDL3GPU folder.

Right, I'm going to create an issue for this, so we don't forget.
Since we currently have a single in-repo backend, I don't think this is urgent?
As long as the classes have a unique name (through e.g. a suffix), they can be put anywhere.

AJenbo:
The added suffixes are making it harder for me to work on the code base so that's part of my frustration. It's very possible that this is something unusual about how I work.

So fare the only parts with SDL3GPU specific code are the Direct3DRM, Direct3DRMViewport and Direct3DRMDevice. But ideally we can push it all into Direct3DDevice2 which would let the application detect and pick a functional render via the GUID:

https://github.com/isledecomp/isle-portable/blob/e1cddc5c0ec43c1a4dc036f635aa652f971a6941/LEGO1/mxdirectx/mxdirect3d.cpp#L162

(more specifically Direct3DDevice2::CreateDevice() would serve to instanciate the chosen backend and inject it into the constructor of `Direct3DRMDevice` which would in turn propagate it to the `Direct3DRMViewport`).

That would allow the whole of d3drm to be common code across backends. It is obviously not quite there at the moment so it's fair enough if it lives in the SDL3GPU folder, but feel free to look into the needed changes to make it 3D render backend agnostic :)
We would have to design a backend interface for this which is something I wasn't going to do before we had a more solid idea of how rendering works, but if it's nessesery for allowing some to test the code then lets look in to doing so so that we can avoid major churn in the code as well as duplication or multiple layers of inheritance.

It's my understanding that this is also similar to how the original D3DRM worked.

madebr:
Feel free to move and rename the classes however you like, as long as it makes sense  and you don't feel blocked :)

madebr:
Alternatively, we can put all classes in a namespace. Then we don't need the suffixes anymore.

AJenbo:
Now there is an ideal. I'll wait until I'm back from vacation to save my wrists a bit on that account

PR #183, opened 2025-05-26
miniwin: Make Lego Island playable

AJenbo:
We can now interact with the 3D world, and there by go places and do things:
image

@madebr this PR is a bit thrown together and somethings might be in none-ideal files or not implemented in the best manner. If you have time and are willing I would appreciate if you can help with cleaning it up and getting in to shape for merging.

### DIFF
--- a/miniwin/sdl3gpu/src/miniwin_d3drmviewport.cpp
+++ b/miniwin/sdl3gpu/src/miniwin_d3drmviewport.cpp
@@ -5,7 +5,9 @@
 #include "miniwin_p.h"
 
 #include <SDL3/SDL.h>
+#include <float.h>
 #include <functional>
+#include <math.h>
 
 Direct3DRMViewport_SDL3GPUImpl::Direct3DRMViewport_SDL3GPUImpl(
 	DWORD width,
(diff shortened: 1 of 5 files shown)


kernaltrap8:
Awesome work!

madebr (approved):
Very good first version!

foxtacles:
@AJenbo let me know if you want it merged now or wait until it's improved as you mentioned

miniwin/sdl3gpu/src/miniwin_d3drmviewport.cpp
```diff
@@ -468,10 +471,242 @@ HRESULT Direct3DRMViewport_SDL3GPUImpl::InverseTransform(D3DVECTOR* world, D3DRM
 	return DD_OK;
 }
 
+struct Ray {
+	D3DVECTOR origin;
+	D3DVECTOR direction;
+};
+
+// Ray-box intersection: slab method
+bool RayIntersectsBox(const Ray& ray, const D3DRMBOX& box, float& outT)
```
AllMeatball, line 480: We should probably start putting functions like this into a math utility library inside miniwin.

foxtacles: Unrelated to this specific piece of code, but isle already uses vec.h for many common vector/matrix operations: https://github.com/isledecomp/isle-portable/blob/master/3rdparty/vec/vec.h

Could consider using it too, I see several ops miniwin wrote out manually that are implemented in this.

AJenbo:
> @AJenbo let me know if you want it merged now or wait until it's improved as you mentioned

Let not perfect be the enemy of good.

From testing it works well and the code can always be improved going forward.. as long as things aren't obviously done wrong I think it's good to merge. Once merged it's available for others to improve and in this case especially helps with figuring out other issues.

miniwin/sdl3gpu/src/miniwin_d3drmviewport.cpp
madebr: I'd keep lego and miniwin math separate. At max copy some logic here and there.

AJenbo: Agree, but making miniwin depend on vec3 should be fine

PR #187, opened 2025-05-28
Fix crash in `LegoAnimationManager::FUN_100609f0`

kernaltrap8:
I've come across an issue where the game will crash when you ~~try to play two animations at once~~ click on an actor/object from too far away. Without this "fix", the game crashes. Making a PR here since I don't think this happens on win32, however I can't test because I'm on Linux and Wine doesn't really like Lego Island.

Hopefully this is the "correct" fix, I haven't had crashing after applying the fix.

### DIFF
--- a/LEGO1/lego/legoomni/src/common/legoanimationmanager.cpp
+++ b/LEGO1/lego/legoomni/src/common/legoanimationmanager.cpp
@@ -1081,7 +1081,9 @@ MxResult LegoAnimationManager::FUN_100609f0(MxU32 p_objectId, MxMatrix* p_matrix
 	char buf[256];
 	sprintf(buf, "%s:%d", g_strANIMMAN_ID, info->m_index);
 
-	action.SetAtomId(*Lego()->GetWorldAtom(m_worldId));
+	if (m_worldId != LegoOmni::e_undefined) {
+	    action.SetAtomId(*Lego()->GetWorldAtom(m_worldId));
+	}
 	action.SetObjectId(p_objectId);
 	action.SetUnknown24(-1);
 	action.AppendExtra(strlen(buf) + 1, buf);


foxtacles:
How can the issue be reproduced? If there really is a discrepancy between platforms, I'd like to understand why before making a change to the game's code. Generally there shouldn't be a need to do so if all UB/ASAN and data types issues that might lead to problems are fixed.

If the issue does occur on every platform we need to first ascertain it is not due to incomplete or faulty graphics/miniwin implementations.

In other words, let's address possible root causes outside the game code first if there are any.

madebr:
UBSAN also often flags this error for me. I doubt this is the correct fix though.

kernaltrap8:
> How can the issue be reproduced? If there really is a discrepancy between platforms, I'd like to understand why before making a change to the game's code. Generally there shouldn't be a need to do so if all UB/ASAN and data types issues that might lead to problems are fixed.
> 
> If the issue does occur on every platform we need to first ascertain it is not due to incomplete or faulty graphics/miniwin implementations.
> 
> In other words, let's address possible root causes outside the game code first if there are any.

Its a little inconsistent, but if you click on a actor for example, and then click the skateboard right after, the game crashes. I think it happens for any actor/object, so if you were to click on an actor and then quickly click another, it should crash. The check I added *should* fix it because in LLDB, `m_worldId` was `e_undefined` and it was causing a nullptr de-reference

foxtacles:
> > How can the issue be reproduced? If there really is a discrepancy between platforms, I'd like to understand why before making a change to the game's code. Generally there shouldn't be a need to do so if all UB/ASAN and data types issues that might lead to problems are fixed.
> > If the issue does occur on every platform we need to first ascertain it is not due to incomplete or faulty graphics/miniwin implementations.
> > In other words, let's address possible root causes outside the game code first if there are any.
> 
> Its a little inconsistent, but if you click on a actor for example, and then click the skateboard right after, the game crashes. I think it happens for any actor/object, so if you were to click on an actor and then quickly click another, it should crash. The check I added _should_ fix it because in LLDB, `m_worldId` was `e_undefined` and it was causing a nullptr de-reference

Ok thank you. I'll try to get to the bottom of this soon

kernaltrap8:
I checked again and it seems to just happen randomly. What I did was sign in, and then choose Papa, and go to the Pizzeria. Click on Mama, and the game should crash. You should see something similar to this in LLDB:
```
Process 451912 stopped
* thread #1, name = 'isle', stop reason = signal SIGSEGV: address not mapped to object (fault address: 0x0)
    frame #0: 0x00007ffff7c4cf9c liblego1.so`LegoAnimationManager::FUN_100609f0(this=0x0000555556a473f0, p_objectId=734, p_matrix=0x0000000000000000, p_und1='\0', p_und2='\x01') at legoanimationmanager.cpp:1084:18
   1081         char buf[256];
   1082         sprintf(buf, "%s:%d", g_strANIMMAN_ID, info->m_index);
   1083
-> 1084         action.SetAtomId(*Lego()->GetWorldAtom(m_worldId));
   1085         action.SetObjectId(p_objectId);
   1086         action.SetUnknown24(-1);
   1087         action.AppendExtra(strlen(buf) + 1, buf);
   ```
If needed I can make an issue as well.

foxtacles:
What I can tell for now:

I can easily reproduce this issue in Linux 64bit.  The issue is not present on Win32 or Win64 (all miniwin builds). Note that I've not just tested for a crash, but for the condition that `m_worldId == e_undefined` when this function is called. That is never the case except for Linux 64bit.

foxtacles:
Found the issue: https://github.com/isledecomp/isle-portable/pull/190

Closing

PR #191, opened 2025-05-29
legoanimationmanager: use delete[]

madebr:


foxtacles:
Can you put it in the decomp repo? Thank you

PR #194, opened 2025-05-29
Fixing some ubsan errors while walking around on the island

madebr:


foxtacles:
Can you also share it on the decomp (except the unaligned reads)

foxtacles (approved):

Issue #192, opened 2025-05-29
rand() has different behavior on *nex


madebr:
`rand` on Windows [returns an integer in the range [0, 32767]](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/rand?view=msvc-170).
However, [linux rand's return a number in the range [0, RAND_MAX].](https://www.man7.org/linux/man-pages/man3/rand.3.html)

This variable on my system is defined as:
```
stdlib.h:87:#define	RAND_MAX	2147483647
```
which is identical to `INT_MAX`.

Instead, we should probably add a `minwin_rand` and `miniwin_srand` that mimics the windows implementation, and define 2 macros doing `rand=miniwin_rand`and `srand=miniwin_srand`.

AJenbo:
Indeed this is the issue causing the NPC to behave erratic. Simply applying RAND_MAX in the places it was missed is enough to get them to behave correctly:

https://github.com/user-attachments/assets/fa3bc2d9-af23-4c98-9b49-7078c22e26b6

I created a PR for this in decomp https://github.com/isledecomp/isle/pull/1528

AJenbo:
Arh but it does cause overflow lines like this:

https://github.com/isledecomp/isle-portable/blob/8c9feaac1e7cd71dc9c5ee1a19552acfad571e9f/LEGO1/lego/legoomni/src/common/legoanimationmanager.cpp#L1565

(i tested with `SHRT_MAX` corrected to `RAND_MAX`)

madebr:
I "fixed" those locally (temporarily) by replacing rand with `(rand() % SHORT_MAX)`

foxtacles:
This should be resolved now? @madebr

madebr:
> This should be resolved now? [@madebr](https://github.com/madebr)

For now, yes!

PR #195, opened 2025-05-29
Fix integer overflow in `LegoAnimationManager::Tickle`

kernaltrap8:
This PR fixes a integer overflow in the `rand()` call inside of `LegoAnimationManager::Tickle`. During animation sequences, like the one that plays when you first get in-game and the Infomaniac tells you about the Pizzeria, the `rand` function could return a massive value that would overflow the `int` type of `MxLong`.


foxtacles:
We should solve this generically as described in https://github.com/isledecomp/isle-portable/issues/192

kernaltrap8:
Yes I just saw the issue. I'll close this and re-do it properly if someone doesn't beat me to it.

PR #196, opened 2025-05-29
Fix sky color

AJenbo:
I managed this when implementing the geometry


AJenbo:
Merged as it was already approved in https://github.com/isledecomp/isle-portable/pull/149

PR #201, opened 2025-05-30
software: Add support for transparent surfaces

AJenbo:
image

This also fixes fixes the issues related to the software render not working properly in 8bit mode, which can give a nice boost to performance.


madebr:
With https://github.com/AJenbo/isle-portable/pull/3, this pr is good to merge.

AJenbo:
Arh thanks I have had UBSAN off because of the RNG bugs but it should be safe for me to turn it back on now. Thanks for catching that.

madebr:
> Arh thanks I have had UBSAN off because of the RNG bugs but it should be safe for me to turn it back on now. Thanks for catching that.

I only get ubsan errors when I try something new, like doing a car build in the petrol station recently.
It's pretty good. Car building ubsan-errors because of missing `IDirect3DRMViewport::Transform` and `IDirect3DRMViewport::InverseTransform`.

PR #204, opened 2025-05-30
Allow entering helicopter build minigame

madebr:
Fixing these 2 casts allows me to enter the helicopter build game.

(Building itself fails due to fallout of a missing `Direct3DRMViewportImpl::Transform` implementation)


foxtacles:
Seems like bugs in the decomp. Can you open a PR there (too)?

PR #208, opened 2025-05-31
Unhook turn speed from the frame rate

itsmattkc:
This is a simple fix adapted from the patch I wrote for LEGO Island Rebuilder. Crucially it multiplies `m_rotationalVel` by `deltaTime` so the speed is unaffected by the frame rate.

Also multiplies by `m_maxRotationalVel` to achieve full "intended" turning speed. Not sure if this is actually the correct thing to do, but the results feel about right to me.


kernaltrap8:
Yes!!!! Been waiting for this!! Thanks Matt!

jpeisach:
We should maybe have an option to remove any 60FPS cap, if one exists

AJenbo:
i think there is a 90fps cap

TheBoxyBear:
The elusive patch is finally here for real this time.

Issue #128, opened 2025-05-19
Support different resolutions (make startup more robust with regards to full screen)


ItzSwirlz:
Some monitors, like the one my M1 MBP has, does not have a 640x480 resolution setting. Currently, the game appears to demand a 640x480 resolution, crashing without one.

AJenbo:
I hear that the maintainers do not want to do so. But  I have some alternate suggestions.
- When no fitting mode is detected switch to windowed mode
- Use [up-scaling](https://wiki.libsdl.org/SDL3/SDL_SetRenderLogicalPresentation)
- Show an error if no fitting resolution can be set
- Do not check for supported resolution when running in window mode

MaddTheSane:
From what I've heard, this repository is to keep it as close as possible to the original Lego Island with minimal improvements, generally to get it to work (and port it to other architectures). My guess is that the next stage of this project, which will be more geared towards improvements and Quality of Life fixes, will look more into fixing the issue.

foxtacles:
Getting the game to launch properly in both windowed and full screen mode is what it's really about. This has worked in the original LEGO Island as well, so we should make it work here too.

This won't involve changing the actual game resolution, just trivial things like up-scaling

AJenbo:
https://github.com/isledecomp/isle-portable/pull/291 + https://github.com/isledecomp/isle-portable/pull/290 should resolve this issue.

PR #210, opened 2025-05-31
Port CONFIG to Qt

JPeisach:
Does not compile as of opening PR. The only file complete is directx5.cpp.


madebr:
The qt6 dialog now has functionality parity with the mfc version, but has regressed a bit on graphics side.
- the dialog can be resized, and behaves not nice. It should perhaps be minimally sized
- When pressing the advanced button, the dialog is not resized.
- The row count in the list widget needs to be limited
- The about dialog can be resized (use F1 to visit it)

AJenbo:
image

jpeisach:
Tomorrow I'll look at this, but on M1 Pro seems to crash.
image

jpeisach:
> Tomorrow I'll look at this, but on M1 Pro seems to crash. image

The issue was it trying to look for a valid display device, I didn't modify legodxinfo.cpp to a resolution my laptop supports, so it crashed there.

The app works - thanks to @madebr for the help.

Of course, the issue of hardcoding resolutions needs to be fixed, as foxtacles suggested by looking into upscaling - #128 

Maybe also update the README to clarify usage of qt

CONFIG/MainDlg.h
```diff
@@ -1,55 +1,54 @@
...
 public:
-	CMainDialog(CWnd* pParent);
-	enum {
-		IDD = IDD_MAIN
-	};
+	CMainDialog(QWidget* pParent = nullptr);
 
 protected:
-	void DoDataExchange(CDataExchange* pDX) override;
+	// void DoDataExchange(CDataExchange* pDX) override;
```
AJenbo: Looks like this can just be removed

madebr:
```suggestion
```

AJenbo (approved):
LGTM already useful

PR #218, opened 2025-06-02
Add highlights to software rendering

AJenbo:
This should now be very close to the original game, it's leaning more to the hardware renderer then the software render with brighter lighting and blended transparency.

image

image

This makes the software renderer feature complete as fare as I'm concerned.


kernaltrap8:
Looks awesome!

madebr (approved):

PR #222, opened 2025-06-02
miniwin: Compartmentalize EnumDevices

AllMeatball:
This compartmentalizes everything into a corresponding EnumDevice function for each renderer backend.

This makes it easier to add new renderers. I think this could be done better, so I made a draft PR.


AJenbo:
Good first move, it feels like it can be build on so I think it's good to merge.

AJenbo:
I think a good next step would be to drop the context related methods and instead make them part of the _EnumDevice() call.


	virtual void GetDesc(D3DDEVICEDESC* halDesc, D3DDEVICEDESC* helDesc) = 0;
	virtual const char* GetName() = 0;

Issue #207, opened 2025-05-31
`Pick` doesn't work reliably


foxtacles:
It doesn't seem to work for certain faraway objects (i.e. try clicking the Infocenter when you are around the Pizzeria)

3D elements also seem to sometimes obstruct 2D clicks (i.e. try clicking the "Right" navigation button in the room with the spinning score cube; it won't reliably register the click)

AJenbo:
I can explain:

https://github.com/isledecomp/isle-portable/blob/0ab5070b46148767f3dd987939d45cc77d960679/miniwin/src/d3drm/d3drmviewport.cpp#L548

Right now it's only checking for possible object using there bounding box. But we need to then do an extra check on the candidate object to see if we hit an actual triangle, or move on to the next candidate.

PR #227, opened 2025-06-03
Make Pick() more picky

AJenbo:
Fixes https://github.com/isledecomp/isle-portable/issues/207

Tested by clicking the bank from here:
Screenshot from 2025-06-03 19-58-16


AllMeatball:
> Make Pick() more picky

image

PR #232, opened 2025-06-04
Log FPS with SDL_Log for now

AJenbo:
Cause I'm tired of inserting it


madebr:
Could [`SDL_RenderDebugText`](https://wiki.libsdl.org/SDL3/SDL_RenderDebugText) be useful here? It'd avoid console spamming.

AJenbo:
> Could [`SDL_RenderDebugText`](https://wiki.libsdl.org/SDL3/SDL_RenderDebugText) be useful here? It'd avoid console spamming.

Nice find, that would have been perfect, but at least for now we can't use it as it requires the use of SDL_Renderer.

I'm still not 100% if we will use SDL_Renderer or output directly from the backends, But if we do end up with SDL_Renderer contorting the screen we should come back and implement it with this.

Regarding console spam, it's only every 5 sec and only if you enable it via the debug commands.

PR #233, opened 2025-06-04
Viewport cleanup

AJenbo:
Some further cleanup of the viewport implementation

### DIFF
--- a/miniwin/src/d3drm/d3drmviewport.cpp
+++ b/miniwin/src/d3drm/d3drmviewport.cpp
@@ -84,11 +84,8 @@ static void ComputeFrameWorldMatrix(IDirect3DRMFrame* frame, D3DRMMATRIX4D out)
 	IDirect3DRMFrame* cur = frame;
 	while (cur) {
 		auto* impl = static_cast<Direct3DRMFrameImpl*>(cur);
-		D3DRMMATRIX4D local;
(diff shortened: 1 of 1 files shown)


miniwin/src/d3drm/d3drmviewport.cpp
```diff
@@ -99,39 +96,58 @@ static void ComputeFrameWorldMatrix(IDirect3DRMFrame* frame, D3DRMMATRIX4D out)
...
+inline D3DVECTOR TransformNormal(const D3DVECTOR& v, const Matrix3x3& m)
+{
+	return {
+		v.x * m[0][0] + v.y * m[1][0] + v.z * m[2][0],
+		v.x * m[0][1] + v.y * m[1][1] + v.z * m[2][1],
+		v.x * m[0][2] + v.y * m[1][2] + v.z * m[2][2]
+	};
+}
+
+inline D3DVECTOR TransformPoint(const D3DVECTOR& v, const D3DRMMATRIX4D& m)
```
madebr: little bikeshed, but `p` for point makes more sense here.
A point is a position with `w` component != 0.
A vector is a directtion with `w` component= = 0.

AJenbo: Thanks, i was a little unsure where to go with that and just neded letting them all have the same name across helpers.

PR #239, opened 2025-06-05
Add texture support for OpenGL

AJenbo:
So smooth
image

### DIFF
--- a/miniwin/src/internal/d3drmrenderer_opengl15.h
+++ b/miniwin/src/internal/d3drmrenderer_opengl15.h
@@ -10,6 +10,12 @@
 
 DEFINE_GUID(OPENGL15_GUID, 0x682656F3, 0x0000, 0x0000, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x03);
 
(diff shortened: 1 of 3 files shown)


miniwin/src/internal/d3drmrenderer_opengl15.h
```diff
@@ -10,6 +10,12 @@
 
 DEFINE_GUID(OPENGL15_GUID, 0x682656F3, 0x0000, 0x0000, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x03);
 
+struct GLTextureCacheEntry {
+	IDirect3DRMTexture* texture;
+	Uint32 version;
+	GLuint glTextureId;
+};
```
madebr, line 17: I'd put this struct inside the `OpenGL15Renderer` as a subclass.

madebr (approved):

miniwin/src/internal/d3drmrenderer_opengl15.h
AJenbo: Ok, I look forward to seeing how so I can learn :)

PR #243, opened 2025-06-06
Split up pixel loop

AJenbo:
This improves performance roughly 8%.
It does mean we end up with some duplicate logic - but that's the name of the game.


AJenbo:
After fixing the bugs this looses 6% performance

PR #246, opened 2025-06-06
miniwin: Use the GLEW::GLEW target to find glew.

sulix:
On some systems, the old-style GLEW_TARGET_LIBRARIES variable is not set by find_package(GLEW), and instead we must use the newer GLEW::GLEW target, as we do for OpenGL.

This fixes a build error on openSUSE Tumbleweed, which has a glew CMake config in /usr/lib64/cmake/glew/glew-config.cmake.

**Please check that this doesn't break your existing config. I suspect there are two different ways GLEW can be found — ``glew-config.cmake`` and ``FindGLEW.cmake`` — and I've only tried and tested one of them.**


madebr (approved):

AJenbo:
It build on CI so works for me

PR #248, opened 2025-06-06
Simplify calculations for flat shading

AJenbo:
Software renderer before:
100% CPU 90fps 

After:
92% CPU 90fps 

`GeometryVertex` and `D3DRMVERTEX` had become identical in memory layout so dropping `GeometryVertex` to avoid some unnecessary data copying.

OpenGL can do flat shading in GPU.
Software can skip some calculations when it knows it's flat shading.
SDL GPU won't see benefits now but both it and Software will benefit once meshes are uploaded instead of pushed every frame.

Test scene:
image
(high detail model cheat)

### DIFF
--- a/miniwin/src/d3drm/backends/software/renderer.cpp
+++ b/miniwin/src/d3drm/backends/software/renderer.cpp
@@ -146,14 +146,14 @@ void Direct3DRMSoftwareRenderer::BlendPixel(Uint8* pixelAddr, Uint8 r, Uint8 g, 
 	memcpy(pixelAddr, &blended, m_bytesPerPixel);
 }
 
-SDL_Color Direct3DRMSoftwareRenderer::ApplyLighting(const D3DRMVERTEX& vertex, const Appearance& appearance)
+SDL_Color Direct3DRMSoftwareRenderer::ApplyLighting(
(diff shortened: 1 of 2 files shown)


foxtacles:
I'll do some testing tomorrow and see how software perf compares

foxtacles:
Performance-wise this doesn't seem to change much for me (maybe it's slightly better, not sure). I'm only testing Software rendering. Tested both Windows x64 and Emscripten perf:

Before PR:

Untitled

After PR:

Untitled2

Note that FPS in Emscripten seems to be capped at 60FPS. The drop to the 20-30FPS happens when standing in front of the Pizza sign.

I did notice differences in rendering though and it seems to me like that might be a regression.

Before this PR (see the sign):

Untitled3

After this PR:

Untitled4

AJenbo:
@foxtacles the visual change has now been addressed.

foxtacles (approved):

AJenbo:
It's an almost 8% drop in CPU load, it's still higher then before https://github.com/isledecomp/isle-portable/pull/257 but I expect we will win more then we lost once the models are preprocessed instead of doing it per frame.

PR #251, opened 2025-06-07
Deduplicate

AJenbo:
About 26% of vertices can be discarded as duplicates once we know it's a flat shaded mesh.
This game never changes the shading quality after creating a mesh, but others might so ideally this deduplication should happen when the data is submitted to the renderer not in the mesh objects.


AJenbo:
This has now been rolled into https://github.com/isledecomp/isle-portable/pull/257

PR #252, opened 2025-06-07
OpenGL 8xMSAA

AJenbo:
This probably should be some sort of option or something, but it looks amazing

image

### DIFF
--- a/miniwin/src/d3drm/backends/opengl15/renderer.cpp
+++ b/miniwin/src/d3drm/backends/opengl15/renderer.cpp
@@ -6,12 +6,13 @@
 #include <GL/glew.h>
 #include <cstring>
 #include <vector>
-
 Direct3DRMRenderer* OpenGL15Renderer::Create(DWORD width, DWORD height)
 {
 	SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 1);
 	SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 5);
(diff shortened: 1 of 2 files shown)


kernaltrap8:
Holy crap. Looks awesome! Is there any side effects (e.g., artifacting, lower performance)?

madebr (approved):
Looks good, though I'm not a OpenGL dev :)

kernaltrap8:
> Looks good, though I'm not a OpenGL dev :)

Its very obvious that it looks good when you side-by-side. Top is MSAA, bottom without. 

image
image

Just don't try higher than 8 ;)
image

AJenbo:
> Looks good, though I'm not a OpenGL dev :)

Thanks, I have it in draft since I want to make it gratefully reduce it to what the users hardware supports rather then require 8x support.

AJenbo:
> Holy crap. Looks awesome! Is there any side effects (e.g., artifacting, lower performance)?

It does tax the GPU a bit more, but it's not like it's a GPU taxing game in the first place.

AJenbo:
@madebr anychance you would be willing to take over this PR and expose the setting via `IDirect3DRMMiniwinDevice`?
It should go from 1-8, eventually SDL_GPU should be able to support this as well, but probably not software :D

AJenbo:
this would make it a bit less likely that the renderer would be OpenGL 1.1 compatible and I think that should be this backends main focuse (backwards compatability), it at the very least would increase complexity, so I think if we are to support AA it should be on one of the modern backends. But high resoolution rendering already goes a long way so not sure how needed it will be in the end.

PR #255, opened 2025-06-08
Add status badges for platforms

AJenbo:
Things now work well under miniwin, there are still some improvement like faster rendering and full screen that can be done but this is true for most Windows builds and the other emojis does a better job of showing where specific short comings are present.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Please note: this project is dedicated to achieving platform independence withou
 
 ### Supported platforms
 
-| Platform | status |
+| Platform | Status |
 | - | - | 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,11 +10,11 @@ Please note: this project is dedicated to achieving platform independence withou
 
 ### Supported platforms
 
-| Platform | Implementation status |
+| Platform | status |
```
foxtacles:
```suggestion
| Platform | Status |
```

foxtacles (approved):

PR #257, opened 2025-06-08
Use indexed faces

AJenbo:
Instead of rebuilding every face on the CPU we send the indexes directly to the GPU since that's something it can actually handle.

Todo: 
- Update Software and SDL_GPU to handle the indexed format

This isn't intended to improve performance yet but it's a necessary step on that path.


madebr:
I think there's something wrong with textures.
Screenshot from 2025-06-09 04-09-55
image

AJenbo:
@madebr looks like your just running the game in low quality mode
image

madebr (approved):

PR #261, opened 2025-06-08
Some cleanups for the software rendere

AJenbo:
Most of these are just code cleanups but there are some potentials for small performance wins, and there should be no visual changes from any of it.


foxtacles (approved):
LGTM - I can't test it right now but if you did a quick sanity check running the renderer and didn't notice any issues then please merge

AJenbo:
I want to the prison and the pizza sign looked like in your before screen shot.

PR #263, opened 2025-06-08
Fix UBSAN warning when starting the jetski race

madebr:


LEGO1/lego/legoomni/src/video/legopartpresenter.cpp
```diff
@@ -91,7 +91,7 @@ MxResult LegoPartPresenter::Read(MxDSChunk& p_chunk)
 		SDL_strlwr(textureName);
 
 		if (textureName[0] == '^') {
-			strcpy(textureName, textureName + 1);
+			memmove(textureName, textureName + 1, strlen(textureName + 1));
```
foxtacles, line 94: Is the 3rd parameter correct? 

https://github.com/isledecomp/isle-portable/blob/f3a4d8fcb5d26c505d9eab9181b72c496896433f/LEGO1/lego/legoomni/src/video/legomodelpresenter.cpp#L101

AJenbo (approved):

AJenbo:
... sorry

LEGO1/lego/legoomni/src/video/legopartpresenter.cpp
madebr: Yes, you want to move the texture one character to the front.
So you want to copy the length of the string, starting from the second character.
`strlen(textureName + 1) == strlen(textureName) - 1`

foxtacles: In that case the other instance is bugged, no

PR #267, opened 2025-06-09
Add IDirect3DRMMiniwinDevice interface

madebr:


miniwin/src/d3drm/backends/software/renderer.cpp
```diff
@@ -370,8 +370,8 @@ void Direct3DRMSoftwareRenderer::AddTextureDestroyCallback(Uint32 id, IDirect3DR
 
 Uint32 Direct3DRMSoftwareRenderer::GetTextureId(IDirect3DRMTexture* iTexture)
 {
-	auto texture = static_cast<Direct3DRMTextureImpl*>(iTexture);
-	auto surface = static_cast<DirectDrawSurfaceImpl*>(texture->m_surface);
+	auto texture = dynamic_cast<Direct3DRMTextureImpl*>(iTexture);
+	auto surface = dynamic_cast<DirectDrawSurfaceImpl*>(texture->m_surface);
```
AJenbo: how come this is needed?

AJenbo (approved):

miniwin/src/d3drm/backends/software/renderer.cpp
madebr: I had an issue with extending from IUnknown. I think this change is no longer needed.

madebr:
@foxtacles Can you test this with the MS d3drm driver?
This pr fetches information about the d3drm backend using standard COM, and is unconditionally built.

miniwin/include/miniwin/miniwindevice.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+DEFINE_GUID(IID_IDirect3DRMMiniwinDevice, 0x6eb09673, 0x8d30, 0x4d8a, 0x8d, 0x81, 0x34, 0xea, 0x69, 0x30, 0x12, 0x01);
```
madebr, line 3: For those interested, this was generated with `uuidgen`.

AJenbo: So it's not a valid GUID?

foxtacles:
image

image

foxtacles:
Is the expectation that something is logged with MS d3drm? If yes, it's not working.

`if (d3drmMiniwinDev) {`

evaluates to false, because of

```
	if (!SUCCEEDED(d3drmdev->QueryInterface(IID_IDirect3DRMMiniwinDevice, (void**) &d3drmMiniwinDev))) {
		return nullptr;
	}
```

miniwin/include/miniwin/miniwindevice.h
```diff
@@ -0,0 +1,9 @@
...
+
+struct IDirect3DRMMiniwinDevice : virtual public IUnknown {
+	virtual float GetShininessFactor() = 0;
+	virtual HRESULT SetShininessFactor(float factor) = 0;
+	virtual HRESULT GetDescription(char* buffer, int bufferSize) = 0;
```
AJenbo: A more viable option would be to use the names gotten from `EnumDevicesCallback()` that way it works with DirectX as well..

madebr:
> Is the expectation that something is logged with MS d3drm? If yes, it's not working.

This was expected. I was checking whether the MS driver gracefully ignored the extra interface.
Using ajenbo's suggestions, and modifying LEGO1 slightly, it should now support all d3d(rm) backends.

AJenbo (approved):

PR #271, opened 2025-06-10
Fix `MxStopWatch`

foxtacles:
This actually broke the game in at least one way. Scaling up the ambient Lego pedestrians depends on the performance counter. Right now it is stuck at 5 maximum characters in the world because the counter always returns `0`, and never scales to expected values.


madebr (approved):
Crazy this even worked before.

PR #272, opened 2025-06-10
Upload mesh

AJenbo:
Adds a mesh upload phase to the renderer. This allows use to pre-process the meshes once and for SDL_GPU critically upload the meshes to the GPU once.

There are still some wins that have not been implemented, notably:
- Only process lighting once in Software Renderer
- Use indicies for for triangle format in SDL_GPU

But these can come later.

Performance after this PR, all renderers now run at 90FPS:
SDL_GPU: 23% CPU load (up from ~40fps at 100% load)
OpenGL: 25% CPU load (27% previously)
Software: 82% CPU load (this seems a little higher then a few versions ago, but not by much)

### DIFF
--- a/miniwin/src/d3drm/backends/opengl15/renderer.cpp
+++ b/miniwin/src/d3drm/backends/opengl15/renderer.cpp
@@ -192,6 +192,91 @@ Uint32 OpenGL15Renderer::GetTextureId(IDirect3DRMTexture* iTexture)
 	return (Uint32) (m_textures.size() - 1);
 }
 
+GLMeshCacheEntry GLUploadMesh(const MeshGroup& meshGroup)
+{
+	GLMeshCacheEntry cache{&meshGroup, meshGroup.version};
+
+	cache.flat = meshGroup.quality == D3DRMRENDER_FLAT || meshGroup.quality == D3DRMRENDER_UNLITFLAT;
+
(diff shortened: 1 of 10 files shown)


foxtacles:
Software mode performance seems on par with before refactor.

Windows x64: 90FPS ocean view, 50FPS Pizza sign and everything in between while walking around
Emscripten: 60FPS ocean view (capped at that number), 25FPS Pizza sign and everything in between while walking around.

(maxed CPU core)

foxtacles:
For reference, DirectX RGB Software emulation numbers:

Windows 32bit: 90FPS ocean view, 77FPS Pizza sign and everything in between while walking around (but mostly 85-90)

(maxed CPU core)

AJenbo:
> Software mode performance seems on par with before refactor.

I was hoping for a slight bump in performance :) But at mentioned there are still some relatively easy win we can take advantage of after this gets merged.

> For reference, DirectX RGB Software emulation numbers:
> 
> Windows 32bit: 90FPS ocean view, 77FPS Pizza sign and everything in between while walking around

wow we are doing much better then I thought, thanks for the comparison. Doing the 2D via SDL_Renderer should get us very close to the DirectX numbers I would think.

madebr (approved):

PR #273, opened 2025-06-10
Add `Max LOD` and `Max Allowed Extras` configs

foxtacles:
These aren't really new configuration options in the game; this PR just exposes them in the `.ini`:

* `Max LOD`: specify maximum LOD. Higher value draws higher LOD in further distance. Default is the game's default value of `3.6`.
* `Max Allowed Extras`: specify the max. number of Lego characters walking around in the world. Normally this is automatically derived from `Island Quality`, but there's value in unlocking this to set arbitrary values higher than the current maximum of `20`.  The absolute maximum the game will be able to add to the world is `47`, that's all the characters that exist.

("extras" is the term they use in the code base for extra actors)

Note that the game will ramp up the number of characters over time depending on game performance. It will start out with `5`, and increase up to `Max Allowed Extras` while performance is acceptable.

Also note until a moment ago there was a bug that was preventing the game from scaling up the number of characters, [but this is fixed now](https://github.com/isledecomp/isle-portable/pull/271).

### DIFF
--- a/LEGO1/mxdirectx/mxstopwatch.h
+++ b/LEGO1/mxdirectx/mxstopwatch.h
@@ -49,7 +49,7 @@ private:
 inline MxStopWatch::MxStopWatch()
 {
 	Reset();
-	m_ticksPerSeconds = SDL_GetPerformanceCounter();
+	m_ticksPerSeconds = SDL_GetPerformanceFrequency();
 }
 
 // FUNCTION: BETA10 0x100d8be0


madebr (approved):
Can any of these options be exposed by the debug ui?
Or are these options only configurable at start-up?

foxtacles:
> Can any of these options be exposed by the debug ui? Or are these options only configurable at start-up?

Yes:

`Max LOD` can be set and changed on the fly using `RealtimeView::SetUserMaxLOD`
`Max Allowed Extras` can be set by changing `AnimationManager()->m_maxAllowedExtras` (just need to allow debug to access the private member)

Also, `AnimationManager()->m_numAllowedExtras` is the current target value for actors in the world in case you want to display that in the debug UI

PR #275, opened 2025-06-10
Upload mesh data to GPU using VBOs when available

AJenbo:
CPU load is now down to 24% so very close to SDL_GPU:Vulkan.

This should be the last of the OpenGL 1.x related optimizations. VBO is a 1.5 feature, there are some later features.

Display lists could be used to further improve performance on systems that lack VBO, but lets wait till someone finds one to test current performance one, it seems using display lists is a bit less compatible and more messy to implement.

The class is renamed from OpenGL15 to OpenGL1 since it should work just fine on OpenGL 1.2, the only blocker for OpenGL 1.1 is setting the texture format to GL_RGBA8 instead of GL_RGBA which ensures that it interacts correctly with CPU access to the pixels. 1.0 might work with how things where in f3a4d8fcb5d26c505d9eab9181b72c496896433f which should be doable as an option, but without a way to verify things I don't think it's work keeping it around.

This also does a check for the FBO extension before trying to use it.

No mip-mapping is implemented but this is general for all backends, texture compression hasn't been explored either.

### DIFF
--- a/miniwin/CMakeLists.txt
+++ b/miniwin/CMakeLists.txt
@@ -29,8 +29,8 @@ add_library(miniwin STATIC EXCLUDE_FROM_ALL
 find_package(OpenGL)
 find_package(GLEW)
 if(OpenGL_FOUND AND GLEW_FOUND)
-  target_sources(miniwin PRIVATE src/d3drm/backends/opengl15/renderer.cpp)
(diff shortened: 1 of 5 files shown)


madebr (approved):
I think I got a 1fps increase :)

PR #280, opened 2025-06-11
Use edge walking for triangle rasterization

AJenbo:
Instead of testing every pixel in the triangle bounding box it's possible to calculates it's row start and end making it much faster to draw it out line by line.
The code for this isn't as easy to follow but the performance is well worth it.

46% CPU load compared to 82% before!

@foxtacles I think this might be faster then DirectX RGB Software emulation, could you please run the numbers on your system?

### DIFF
--- a/miniwin/src/d3drm/backends/software/renderer.cpp
+++ b/miniwin/src/d3drm/backends/software/renderer.cpp
@@ -201,14 +201,34 @@ SDL_Color Direct3DRMSoftwareRenderer::ApplyLighting(
 	};
 }
 
-inline float EdgeFloat(float x0, float y0, float x1, float y1, float x, float y)
-{
-	return (x - x0) * (y1 - y0) - (y - y0) * (x1 - x0);
-}
+struct VertexXY {
+	float x, y, z, w;
+	SDL_Color color;
+	float u_over_w, v_over_w;
+	float one_over_w;
+};
 
-inline float EdgeDouble(double x0, double y0, double x1, double y1, double x, double y)
+VertexXY InterpolateVertex(float y, const VertexXY& v0, const VertexXY& v1)
 {
-	return (x - x0) * (y1 - y0) - (y - y0) * (x1 - x0);
+	float dy = v1.y - v0.y;
+	if (fabsf(dy) < 1e-6f) {
+		dy = 1e-6f;
+	}
+	float t = (y - v0.y) / dy;
+	VertexXY r;
+	r.x = v0.x + t * (v1.x - v0.x);
+	r.z = v0.z + t * (v1.z - v0.z);
+	r.w = v0.w + t * (v1.w - v0.w);
+	r.color.r = static_cast<Uint8>(v0.color.r + t * (v1.color.r - v0.color.r));
+	r.color.g = static_cast<Uint8>(v0.color.g + t * (v1.color.g - v0.color.g));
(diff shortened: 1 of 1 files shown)


miniwin/src/d3drm/backends/software/renderer.cpp
```diff
@@ -278,61 +281,92 @@ void Direct3DRMSoftwareRenderer::DrawTriangleProjected(
...
+
+	verts[1].u_over_w = v1.texCoord.u / p1.w;
+	verts[1].v_over_w = v1.texCoord.v / p1.w;
+	verts[1].one_over_w = 1.0f / p1.w;
+
+	verts[2].u_over_w = v2.texCoord.u / p2.w;
+	verts[2].v_over_w = v2.texCoord.v / p2.w;
+	verts[2].one_over_w = 1.0f / p2.w;
+
+	std::sort(std::begin(verts), std::end(verts), [](const VertexXY& a, const VertexXY& b) { return a.y < b.y; });
```
madebr: Does this optimize well? Sorting a 3 element array can be done with 3 compares. Also, I think it's worth to investigate sorting a `struct { int index; int y;}` array to avoid the moves of the other members.

You're in the hot loop here.

AJenbo: 3 if's with swaps was about 2% faster, using the struct was about 1% faster.

miniwin/src/d3drm/backends/software/renderer.cpp
```diff
@@ -278,61 +288,101 @@ void Direct3DRMSoftwareRenderer::DrawTriangleProjected(
...
+	verts[2].one_over_w = 1.0f / p2.w;
+
+	// Sort verts
+	if (verts[0].y > verts[1].y) {
+		std::swap(verts[0], verts[1]);
+	}
+	if (verts[1].y > verts[2].y) {
+		std::swap(verts[1], verts[2]);
+	}
+	if (verts[0].y > verts[1].y) {
```
jpeisach, line 320: repeated code (verts[0].1 > verts[1].y)

madebr: This is correct. We're sorting a size 3 array here using a  expanded bubble sort.
I suggest to work this out on paper.

madebr:
I'm seeing these line artifacts, probably related to this pr
image

AJenbo:
> I'm seeing these line artifacts, probably related to this pr !

I'm unable to replicate this, any setting or places it happens more so? can you please confirm if it only happens on this PR?

madebr:
It only happens when checking out this pr.

image
image
image
image
image
image
image

Low or high texture quality does not matter.

AJenbo:
I figured out the issue very small triangles in the distance where causing division issues leading them to be rendered as a long line, even overwriting the z-buffer and leaving other things to be rendered on top of the line.

I have clamped the value before making the division which solves the issue: 
```cpp
	float dy = v1.y - v0.y;
	if (fabsf(dy) < 1e-6f) {
		dy = 1e-6f;
	}
	float t = (y - v0.y) / dy;
```

madebr (approved):
I think I'm seeing the fps double.

foxtacles:
The performance is significantly improved. The pizza sign gives me something between 78-82FPS now, compared to ~77FPS with DirectX so you beat it :)   Virtually all other areas of the game give 90FPS. Great job!

PR #283, opened 2025-06-11
SIMD optimize z-buffer clear

AJenbo:
This knocks a few of percent of the load, also this finally makes the SIMD support check at start up relevant ...

Before 46% CPU load after 40% (on x86_64)

### DIFF
--- a/miniwin/src/d3drm/backends/software/renderer.cpp
+++ b/miniwin/src/d3drm/backends/software/renderer.cpp
@@ -10,6 +10,18 @@
 #include <cmath>
 #include <cstring>
 #include <limits>
+#if defined(__x86_64__) || defined(_M_X64) || defined(__i386__) || defined(_M_IX86)
+#include <xmmintrin.h>
+#if defined(__i386__) || defined(_M_IX86)
+#include <xmmintrin.h>
+#endif
+#endif
+#if defined(__arm__) || defined(__aarch64__)
+#include <arm_neon.h>
+#endif
+#if defined(__wasm_simd128__)
+#include <wasm_simd128.h>
+#endif
 
 Direct3DRMSoftwareRenderer::Direct3DRMSoftwareRenderer(DWORD width, DWORD height) : m_width(width), m_height(height)
 {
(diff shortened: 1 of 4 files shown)


foxtacles (approved):

miniwin/src/d3drm/backends/software/renderer.cpp
```diff
@@ -30,7 +42,37 @@ void Direct3DRMSoftwareRenderer::SetProjection(const D3DRMMATRIX4D& projection,
...
 {
-	std::fill(m_zBuffer.begin(), m_zBuffer.end(), std::numeric_limits<float>::infinity());
+	const size_t size = m_zBuffer.size();
+	const float inf = std::numeric_limits<float>::infinity();
+	size_t i = 0;
+
+#if defined(__SSE2__)
+	__m128 inf4 = _mm_set1_ps(inf);
+	for (; i + 4 <= size; i += 4) {
+		_mm_storeu_ps(&m_zBuffer[i], inf4);
```
madebr: You need to allocate `m_zBuffer` with `SDL_aligned_alloc` to satisfy alignment requirements.
Or you do a small loop before the SIMD loop until you're aligned.

miniwin/src/d3drm/backends/software/renderer.cpp
```diff
@@ -30,7 +42,37 @@ void Direct3DRMSoftwareRenderer::SetProjection(const D3DRMMATRIX4D& projection,
 
 void Direct3DRMSoftwareRenderer::ClearZBuffer()
 {
-	std::fill(m_zBuffer.begin(), m_zBuffer.end(), std::numeric_limits<float>::infinity());
+	const size_t size = m_zBuffer.size();
+	const float inf = std::numeric_limits<float>::infinity();
+	size_t i = 0;
+
+#if defined(__SSE2__)
```
madebr: This depends on the compiler and compiler flags.
SDL has helper functions such as `SDL_HasSSE2` to detect CPU SIMD at run-time.
This allows one executable to support both SSE for a Pentium III and avx512 for recent cpus.

miniwin/src/d3drm/backends/software/renderer.cpp
madebr: Looks like CPI and latency of `_mm_store_ps` and `_mm_storeu_ps` is similar on recent cpus.
I don't think this was always true.

AJenbo:
@foxtacles you have to add this to the cmake setup to take advantage of this:
```
-DCMAKE_C_FLAGS="-msimd128" -DCMAKE_CXX_FLAGS="-msimd128"
```

miniwin/src/d3drm/backends/software/renderer.cpp
AJenbo: Could you fix that part, I'm exhausted by SIMD already.

miniwin/src/d3drm/backends/software/renderer.cpp
AJenbo: Done

miniwin/src/d3drm/backends/software/renderer.cpp
```diff
@@ -10,6 +10,18 @@
 #include <cmath>
 #include <cstring>
 #include <limits>
+#if defined(__x86_64__) || defined(_M_X64) || defined(__i386__) || defined(_M_IX86)
+#include <xmmintrin.h>
+#if defined(__i386__) || defined(_M_IX86)
+#include <xmmintrin.h>
```
AJenbo, line 16:
```suggestion
#include <mmintrin.h>
```

miniwin/src/d3drm/backends/software/renderer.cpp
```diff
@@ -123,34 +172,58 @@ void Direct3DRMSoftwareRenderer::DrawTriangleClipped(const D3DRMVERTEX (&v)[3],
 	}
 }
 
-void Direct3DRMSoftwareRenderer::BlendPixel(Uint8* pixelAddr, Uint8 r, Uint8 g, Uint8 b, Uint8 a)
+Uint32 Direct3DRMSoftwareRenderer::BlendPixel(Uint8* pixelAddr, Uint8 r, Uint8 g, Uint8 b, Uint8 a)
```
madebr, line 175: Does the SIMD in this function give better performance?
I'd assume the setup cost is greater then the performance gained by the SIMD. SIMD win if you do loops.

AJenbo: Hm not your right this might actually have hurt it a little, gone

madebr (approved):

PR #287, opened 2025-06-12
Add a OpenGL ES 2.0 backend

AJenbo:
- Fixes full screen cause I broke it in 79726d25a4f88ef92dfde4f81e0d5c312d3d6033
- Adds OpenGL ES 2.0, it's rendering should be identical with SDL_GPU
- Some cleanups for OpenGL 1.2 and other bits as well

9% CPU 90fps


miniwin/src/ddraw/ddsurface.cpp
```diff
@@ -158,11 +158,12 @@ HRESULT DirectDrawSurfaceImpl::Flip(LPDIRECTDRAWSURFACE lpDDSurfaceTargetOverrid
...
-	SDL_UpdateTexture(HWBackBuffer, nullptr, copy->pixels, copy->pitch);
-	SDL_DestroySurface(copy);
-	SDL_RenderTexture(DDRenderer, HWBackBuffer, nullptr, nullptr);
-	SDL_RenderPresent(DDRenderer);
+	//SDL_Surface* copy = SDL_ConvertSurface(DDBackBuffer, HWBackBufferFormat);
+	//SDL_UpdateTexture(HWBackBuffer, nullptr, copy->pixels, copy->pitch);
+	//SDL_DestroySurface(copy);
+	//SDL_RenderTexture(DDRenderer, HWBackBuffer, nullptr, nullptr);
+	//SDL_RenderPresent(DDRenderer);
+	//SDL_UpdateWindowSurface(DDWindow);
```
AJenbo:
```suggestion
	// SDL_Surface* copy = SDL_ConvertSurface(DDBackBuffer, HWBackBufferFormat);
	// SDL_UpdateTexture(HWBackBuffer, nullptr, copy->pixels, copy->pitch);
	// SDL_DestroySurface(copy);
	// SDL_RenderTexture(DDRenderer, HWBackBuffer, nullptr, nullptr);
	// SDL_RenderPresent(DDRenderer);
	// SDL_UpdateWindowSurface(DDWindow);
```

ISLE/isleapp.cpp
```diff
@@ -648,6 +648,9 @@ MxResult IsleApp::SetupWindow()
 	SDL_SetNumberProperty(props, SDL_PROP_WINDOW_CREATE_HEIGHT_NUMBER, g_targetHeight);
 	SDL_SetBooleanProperty(props, SDL_PROP_WINDOW_CREATE_FULLSCREEN_BOOLEAN, m_fullScreen);
 	SDL_SetStringProperty(props, SDL_PROP_WINDOW_CREATE_TITLE_STRING, WINDOW_TITLE);
+#ifdef MINIWIN
+	SDL_SetBooleanProperty(props, SDL_PROP_WINDOW_CREATE_OPENGL_BOOLEAN, true);
```
foxtacles, line 652: this won't cause issues if no OpenGL rendering is used?

AJenbo: It really shouldn't. The docs says that it may cause the window to flicker on startup, but it already does.

> If this window is being created to be used with an [SDL_Renderer](https://wiki.libsdl.org/SDL3/SDL_Renderer), you should not add a graphics API specific property ([SDL_PROP_WINDOW_CREATE_OPENGL_BOOLEAN](https://wiki.libsdl.org/SDL3/SDL_PROP_WINDOW_CREATE_OPENGL_BOOLEAN), etc), as SDL will handle that internally when it chooses a renderer. However, SDL might need to recreate your window at that point, which may cause the window to appear briefly, and then flicker as it is recreated. The correct approach to this is to create the window with the [SDL_PROP_WINDOW_CREATE_HIDDEN_BOOLEAN](https://wiki.libsdl.org/SDL3/SDL_PROP_WINDOW_CREATE_HIDDEN_BOOLEAN) property set to true, then create the renderer, then show the window with [SDL_ShowWindow](https://wiki.libsdl.org/SDL3/SDL_ShowWindow)().

https://wiki.libsdl.org/SDL3/SDL_CreateWindowWithProperties

madebr: The window is always flickering for me on Linux because during initial detection, the window needs vulkan and opengl support. (using current master)

foxtacles: any reason we are not doing the SDL_PROP_WINDOW_CREATE_HIDDEN_BOOLEAN thing?

AJenbo: > The window is always flickering for me on Linux because during initial detection, the window needs vulkan and opengl support. (using current master)

It spawns a separate hidden Window for that, the flicking has gone back long before adding an OpenGL backend.

AJenbo: > any reason we are not doing the SDL_PROP_WINDOW_CREATE_HIDDEN_BOOLEAN thing?

We kinda do already:

https://github.com/isledecomp/isle-portable/blob/1689a0063fe439bd78277435dfb0821c97722c7e/miniwin/src/d3drm/backends/opengl1/renderer.cpp#L21

foxtacles: I meant the main window in Setup window

AJenbo: The other reason is that we would prefer to just use our renderer backend for 2D as well, and Emscripten doesn't support it.

foxtacles (approved):

PR #288, opened 2025-06-12
Clear mesh callbacks on clone

AJenbo:


madebr (approved):

miniwin/src/d3drm/d3drmmesh.cpp
```diff
@@ -25,6 +25,9 @@ HRESULT Direct3DRMMeshImpl::Clone(int flags, GUID iid, void** object)
 
 	auto* clone = new Direct3DRMMeshImpl(*this);
 
+	// We should not call the other meshes callbacks
+	clone->m_callbacks.clear();
```
madebr: Isn't the proper fix to override the clone constructor of `Direct3DRMObjectBaseImpl` and do this there?

miniwin/src/d3drm/d3drmmesh.cpp
```diff
@@ -25,6 +25,9 @@ HRESULT Direct3DRMMeshImpl::Clone(int flags, GUID iid, void** object)
...
+
```
AJenbo:
```suggestion
```

miniwin/src/internal/d3drmobject_impl.h
```diff
@@ -65,7 +70,7 @@ struct Direct3DRMObjectBaseImpl : public T {
 	}
 
 private:
-	std::vector<std::pair<D3DRMOBJECTCALLBACK, void*>> m_callbacks;
 	LPD3DRM_APPDATA m_appData = nullptr;
+	std::vector<std::pair<D3DRMOBJECTCALLBACK, void*>> m_callbacks;
```
AJenbo:
```suggestion
	std::vector<std::pair<D3DRMOBJECTCALLBACK, void*>> m_callbacks;
	LPD3DRM_APPDATA m_appData = nullptr;
```

miniwin/src/internal/d3drmobject_impl.h
```diff
@@ -7,6 +7,11 @@
 
 template <typename T>
 struct Direct3DRMObjectBaseImpl : public T {
+	Direct3DRMObjectBaseImpl(const Direct3DRMObjectBaseImpl& other) : T(other)
+	{
+		// We should not call the callback twice
+		m_callbacks.clear();
```
foxtacles, line 13: This won't be necessary. Since you overrode the copy constructor the default behavior of copying the members won't apply  anymore. So no need to clear callbacks since it's already empty.  All the other members won't be copied either m If that's desired you can have an empty body for this ctor

miniwin/src/internal/d3drmobject_impl.h
```diff
@@ -7,6 +7,11 @@
 
 template <typename T>
 struct Direct3DRMObjectBaseImpl : public T {
+	Direct3DRMObjectBaseImpl(const Direct3DRMObjectBaseImpl& other) : T(other)
```
madebr, line 10: Do we need to `SDL_strdup` `m_name`?

AJenbo: idk, i don't really do OOP in C++

miniwin/src/internal/d3drmobject_impl.h
AJenbo: just trying to follow @madebr's suggestion :shrug:

AJenbo:
Some one else can fix this now that the issue is known

PR #289, opened 2025-06-12
Fix `m_callbacks` resource issue

### LINKED ISSUE #279: miniwin renderer: use after free on freeing of mesh cache after shutdown
This happens to me reliably (but randomly) after 5-10 minutes of gameplay

Image

foxtacles:
Close #279 - haven't been able to reproduce the crashes in 15 minutes, so it's probably fixed


AJenbo (approved):
Thanks :)

Issue #298, opened 2025-06-13
Unable to open registration book


AJenbo:
```
==3631252==ERROR: AddressSanitizer: heap-use-after-free on address 0x51b000025a80 at pc 0x60b385c03b9e bp 0x7ffcc40427e0 sp 0x7ffcc40427d0
READ of size 8 at 0x51b000025a80 thread T0
    #0 0x60b385c03b9d in SDL_AppEvent /home/ajenbo/code/isle-portable/ISLE/isleapp.cpp:571
    #1 0x740ca62bdd15 in SDL_DispatchMainCallbackEvent /home/ajenbo/code/isle-portable/build/_deps/sdl3-src/src/main/SDL_main_callbacks.c:50
    #2 0x740ca62bddce in SDL_DispatchMainCallbackEvents /home/ajenbo/code/isle-portable/build/_deps/sdl3-src/src/main/SDL_main_callbacks.c:66
    #3 0x740ca62bdfaf in SDL_IterateMainCallbacks /home/ajenbo/code/isle-portable/build/_deps/sdl3-src/src/main/SDL_main_callbacks.c:126
    #4 0x740ca6549e57 in GenericIterateMainCallbacks /home/ajenbo/code/isle-portable/build/_deps/sdl3-src/src/main/generic/SDL_sysmain_callbacks.c:51
    #5 0x740ca6549f45 in SDL_EnterAppMainCallbacks_REAL /home/ajenbo/code/isle-portable/build/_deps/sdl3-src/src/main/generic/SDL_sysmain_callbacks.c:62
    #6 0x740ca62747bf in SDL_EnterAppMainCallbacks /home/ajenbo/code/isle-portable/build/_deps/sdl3-src/src/dynapi/SDL_dynapi_procs.h:207
    #7 0x60b385bfc1a9 in SDL_main /home/ajenbo/code/isle-portable/build/_deps/sdl3-src/include/SDL3/SDL_main_impl.h:59
    #8 0x740ca62be0a8 in SDL_RunApp_REAL /home/ajenbo/code/isle-portable/build/_deps/sdl3-src/src/main/SDL_runapp.c:40
    #9 0x740ca626be57 in SDL_RunApp_DEFAULT /home/ajenbo/code/isle-portable/build/_deps/sdl3-src/src/dynapi/SDL_dynapi_procs.h:804
[...]

PR #299, opened 2025-06-13
Don't crash on desktop builds

AJenbo:
Fixes https://github.com/isledecomp/isle-portable/issues/298

But I don't know if it's a proper fix.


foxtacles:
We want his change: https://github.com/isledecomp/isle-portable/pull/300

Doesn't crash for me, but can you test as well?

AJenbo:
Who is he :thinking:

PR #301, opened 2025-06-13
Early culling

AJenbo:
The idea here is to cull all vertices early in view space so that can avoid having to reprocess all vertices for every triangle and instead take advantage of them being indexed so vertices are only processed once and only when visible.

This improves performance by maybe 2% so honestly not a lot, but it does make it more feasible to do multi threaded rasterization.

Originally I also intended to only calculate lighting per plane for flat shading, but it doesn't seem it will be a meaningful performance increase so I left it at once per triangle like before.

Indecies are now also uint16_t like they are on SDL_GPU so things takes up a bit less memory.

Update: I tested out multi threading on top of this, I was able to get good, but not stable performance with 2-6 threads. So I won't be pushing further in that direction. Please only merge this after testing performance on your system and if you think the changes make sens. I do think there are over all improvements to this PR both for software and the other renderers.

### DIFF
--- a/miniwin/include/miniwin/d3drm.h
+++ b/miniwin/include/miniwin/d3drm.h
@@ -145,10 +145,10 @@ struct D3DRMVERTEX {
 	D3DVECTOR position;
 	D3DVECTOR normal;
 	union {
+		TexCoord texCoord;
 		struct {
 			D3DVALUE tu, tv;
 		};
-		TexCoord texCoord;
 	};
 };
 
--- a/miniwin/src/d3drm/backends/opengl1/renderer.cpp
+++ b/miniwin/src/d3drm/backends/opengl1/renderer.cpp
@@ -109,6 +109,10 @@ void OpenGL1Renderer::PushLights(const SceneLight* lightsArray, size_t count)
 	m_lights.assign(lightsArray, lightsArray + count);
 }
 
+void OpenGL1Renderer::SetFrustumPlanes(const Plane* frustumPlanes)
+{
+}
+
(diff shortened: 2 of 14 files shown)


madebr (approved):

foxtacles:
Performance at the pizza sign in Emscripten seems to be slightly lower for me (~39FPS vs. ~42FPS in master). Other than that I can't tell a difference

AJenbo:
@madebr @foxtacles should I close this and just pull out the general cleanups parts in to a separate?

foxtacles:
> @madebr @foxtacles should I close this and just pull out the general cleanups parts in to a separate?

Sounds good

AJenbo:
@foxtacles please see if performance is about the same or improved. I have retained most of the benefits from earlier without going all in on de-duplication so light is still calculated on a per triangle basis. I also included a fix for the 3DS build where a data type was miss matched.

foxtacles:
It seems worse overall - here's the numbers I'm getting in Emscripten using the branch:

Untitled

In master:

master

foxtacles:
I'd say performance is the same at best. Hard to tell because values fluctuate too much

AJenbo:
The cleaner matrix juggling and plan extraction is good enough to warrant it being merged if there are no regression imo.

AJenbo:
Oh, I didn't notice that you had thought it might be worse in the previous comment. Well I hope it's the same else you can try and play around to try and find the change that might cause the change.

foxtacles:
> Oh, I didn't notice that you had thought it might be worse in the previous comment. Well I hope it's the same else you can try and play around to try and find the change that might cause the change.

I've done a few tests and I think at this point I can't really tell a difference, so I think it's fine

PR #307, opened 2025-06-14
Bundle for macOS

ItzSwirlz:
WIP bundle from what [DevilutionX does](https://github.com/diasurgical/DevilutionX/blob/master/CMakeLists.txt#L389).

TODO:
- Bundle config
- Remove the lib folder from the final dmg


CMakeLists.txt
```diff
@@ -638,6 +639,33 @@ set(CPACK_PACKAGE_DIRECTORY "dist")
 set(CPACK_PACKAGE_FILE_NAME "isle-${PROJECT_VERSION}-${ISLE_PACKAGE_NAME}-${CMAKE_SYSTEM_PROCESSOR}")
 if(MSVC)
   set(CPACK_GENERATOR ZIP)
+elseif(APPLE)
+  set(MACOSX_BUNDLE_GUI_IDENTIFIER com.github.isledecomp.isle-portable)
+  set(MACOSX_BUNDLE_COPYRIGHT LGPLv3)
```
jpeisach, line 644: Still awaiting license confirmation: https://github.com/isledecomp/isle/pull/1497

jpeisach:
Actually, maybe we could make this a universal binary?

AJenbo:
brew does not build there libs in a way that allows for that. On the plus side this application does build most of it's own so that would make it more feasible. But I would strongly suggest that you make a working binery before you try to make a universal one.

airpioa:
I build a .app file from the macos prebuild binary but their is something wrong with the binary so that it can never load the libraries
[isle-portable-arm.zip](https://github.com/user-attachments/files/20847571/isle-portable-arm.zip)

jpeisach:
> I build a .app file from the macos prebuild binary but their is something wrong with the binary so that it can never load the libraries [isle-portable-arm.zip](https://github.com/user-attachments/files/20847571/isle-portable-arm.zip)

Not sure what you did here - but CPack can take care of this for us. I just need to come around to making it not complain about the lack of a code signature and make sure the libraries get bundled into `config.app`

AJenbo:
Lack of a code signature will likely require you to pay apple a license fee for a signing certificate.

airpioa:
> > I build a .app file from the macos prebuild binary but their is something wrong with the binary so that it can never load the libraries [isle-portable-arm.zip](https://github.com/user-attachments/files/20847571/isle-portable-arm.zip)
> 
> Not sure what you did here - but CPack can take care of this for us. I just need to come around to making it not complain about the lack of a code signature and make sure the libraries get bundled into `config.app`

oh I have a way to sign it https://stackoverflow.com/questions/77809654/xcode-15-provisioning-profile-without-developer-account or this https://ionic.io/blog/deploying-to-a-device-without-an-apple-developer-account

airpioa:
you can get the profile and then use the certificate it signs for 3 days

Helloyunho:
> Lack of a code signature will likely require you to pay apple a license fee for a signing certificate.

arm64 Macs requires binary to be codesigned, but it doesn't have to be from a dev account. In fact, you don't even need an Apple account to sign it

PR #309, opened 2025-06-14
Add Emscripten customizations for `libwasmfs_fetch.js`

foxtacles:
This adds a `.patch` file that one may apply to the current version of Emscripten in order to improve FetchFS file streaming. Ideally some of these improvements can be upstreamed, but that involves getting more familiar with Emscripten, writing tests etc. and I don't have the time right now (maybe in the future).

Applying this isn't strictly necessary, but it does improve performance and fixes bugs. I'm using it on https://isle.pizza


AJenbo:
I don't feel qualified for FS reviews :)

PR #311, opened 2025-06-14
Add 32bit color support

AJenbo:
This simplifies color conversion which should allow for some further optimizations since we don't have to convert the back buffer or and various other rendering operations each frame.

8 bit and 16 bit support is retained and 24 bit should mostly be there though there are probably a few edge cases that aren't correctly handled (I don't have a ).

This is done by being more generic in the various parts of the game that handles color conversion so this actually results in a reduction in code size. I did end up having to rewrite how the game blanks the screen since it was using the CPU to `memset()` the backbuffer to zeros, which ends up as transparent and there for not having any effect in 32bit color mode, but this is just a big win as it removes the first of the slow CPU graphics operations from the game code, so one stop closer to full GPU rendering.
This should retain DirectX compatibility but it wouldn't hurt if someone verified it by checking if the end credits looks correct.

P.s. I did not convert the cursor, it already had a check to make sure it only displays on 16bit mode which I left in place, this serves as a fix for non-voodoo 1/2 systems that do not need the cursor enabled.

This is also my best offer to mitigate not having ditering support in miniwin as we now get the full color range instead:

image

P.s. the palette conversion is called m_palbit32 instead of `m_pal32bit` to make the CI happy cause apparently only old code can have names like `m_16bitPal` :shrug:


foxtacles:
MS DirectX seems broken on the end credits:

image

miniwin mode looks all good

if you add `m_32bitPal` to `tools\ncc\skip.yml` we can use that name

AJenbo:
This is already a significant improvement to performance:
OpenGL: 6% CPU
SDL_GPU: 15%  CPU
Software: 42% CPUI

AJenbo:
> if you add m_32bitPal to tools\ncc\skip.yml we can use that name

Done and thanks for helping with fixing the background clear on DirectX :)

PR #314, opened 2025-06-15
Add release CI workflow

foxtacles:
It will generate this on `master` updates: https://github.com/foxtacles/isle-portable/releases/tag/continuous

I think these 3 builds are what 99% of people would want to be able to download and test, other builds using other compilers or 32-bit builds we don't need to publish imo

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -86,7 +86,7 @@ jobs:
       - name: Upload Artifact
         uses: actions/upload-artifact@main
         with:
-          name: Release
+          name: Release-${{ matrix.name }}
           path: |
             build/dist/isle-*
 
@@ -98,8 +98,9 @@ jobs:
       - name: Download All Artifacts
         uses: actions/download-artifact@main
         with:
-          name: Release
+          pattern: Release-*
           path: Release
+          merge-multiple: true
 
       - name: Checkout uploadtool
         uses: actions/checkout@v4


AJenbo:
The xp retro community often appreciate a 32bit build but I guess the original works pretty well there

AJenbo:
Would be nice with a screenshot or something showing the game as it current looks, preferably with a modern window boarder from a system where it would have been non trivial to run

.github/workflows/release.yml
```diff
@@ -0,0 +1,115 @@
...
+          cpack .
+
+      - name: Upload Artifact
+        uses: actions/upload-artifact@main
+        with:
+          name: Release-${{ matrix.name }}
+          path: |
+            build/dist/isle-*
+
+  release:
```
madebr, line 93: I'd integrate it with the other workflow to avoid duplication and having to cancel multiple jobs on failure (so github startsvbuilding your latest commit faster). There's also a lot of duplicated code.

foxtacles: That was actually my first attempt, but it ended up looking really ugly. But I'm not super experience with Actions

AJenbo:
Some times a bit of duplication can make things easier to understand

madebr:
I originally chose to build everything in Debug mode because less optimizations mean shorter build times and faster feedback.
Since we are now building Release binaries, and debug MSVC binaries are not really useful for debugging anyways (missing PDB's), how about we only keep the `release.yml` workflow?

foxtacles:
> I originally chose to build everything in Debug mode because less optimizations mean shorter build times and faster feedback. Since we are now building Release binaries, and debug MSVC binaries are not really useful for debugging anyways (missing PDB's), how about we only keep the `release.yml` workflow?

Sounds good. We still need the other parts of `ci.yml` next to the build of course. Personally I'm not using artifacts generated here one way or another so as long as we have a build in the CI that confirms that it builds that's fine for me.

We'd still need to build with `ISLE_DEBUG` in the CI though to make sure it compiles, but not for release.

Issue #142, opened 2025-05-21
ISLE crashes when flip surfaces are enabled


kernaltrap8:
Tried testing out the new changes to Miniwin, and ISLE no longer functions correctly.  The issue was introduced with this PR: #132 <br>
With my minimal understanding of this codebase, as well as my minimal debugging experience, I found that it dies at ```liblego1.so`MxDisplaySurface::VTable0x28```. I noticed in LLDB that `p_bitmap` is NULL. Not sure if this is intended or not.
The specific line that it dies at (in the commit from #132) is mxbitmap.h:111:52, which is:
```MxLong GetBmiHeightAbs() const { return HeightAbs(m_bmiHeader->biHeight); }```
Here's a video showcasing the behavior:
https://github.com/user-attachments/assets/8b451491-7571-44f1-a6ff-fbafba123941

**Comment by foxtacles** [comment]
Doesn't seem to work for me either, although my error is different

![Image](https://github.com/user-attachments/assets/0ace22e3-a455-42fa-bbb7-581c93410314)

**Comment by kernaltrap8** [comment]
> Doesn't seem to work for me either, although my error is different
> 
> ![Image](https://github.com/user-attachments/assets/0ace22e3-a455-42fa-bbb7-581c93410314)

Interesting... I can't even see where `m_data` comes from. Its not passed to that function. Maybe its defined elsewhere and I just can't find it.

**Comment by foxtacles** [comment]
My issue was that my Intel iGPU isn't available on my WSL2 distro.

It works fine on Windows.

**Comment by kernaltrap8** [comment]
Setting `flip surfaces = "true"` in `isle.ini` causes this behavior.

**Comment by AJenbo** [comment]
@kernaltrap8 can you please share some details about your systems, based on the the info you have given it could be anything from Arch Linux and a properly installed RTX 5090 to NetBSD in a VirtualBox with no GPU access.

**Comment by kernaltrap8** [comment]
@AJenbo I'm on Debian 13 Trixie running with a RX 6750 XT using the Mesa RADV driver.

**Comment by AJenbo** [comment]
Interesting I'm using Ubuntu 24.04 on RX 7800 XT using the RADV driver so our systems are actually very similar.

**Comment by AJenbo** [comment]
> Setting `flip surfaces = "true"` in `isle.ini` causes this behavior.

Arh i see, can you please update the title to reflect this, something like "ISLE crashes when flip surface is enabled".

The error is also detected by ASAN:
```

[...]

PR #317, opened 2025-06-15
Move 2D rendering to GPU: part 1

AJenbo:
Fixes https://github.com/isledecomp/isle-portable/issues/142

Split surfaces up in frambuffer, 2d surfaces, and 3d textures. This makes it a lot easier to reason about there logic, which makes moving the rendering portion of it to the backends much more feasible.

All miniwin 2D rendering has now been moved to the GPU, but interaction with the 3D rendering still takes a round trip by the CPU until the pipelines get more direct access to each other. Some systems may benefit from this if they have a relatively weak CPU.

On my system SDL_GPU saw a drop from 15% to 9% on CPU load and the other backs seems to have dropped by maybe 1%, but that wasn't really the goal here.

It's notable that CPU load jumps from 5% to 10% on OpenGL when getting on a vehicle with a gas meter, I don't know what it was before but it tells us a little bit about how much performances is possibly still lost to GPU->CPU->GPU even when not on a vehicle.

Another point is that it was technically possible for me to use SDL_TEXTUREACCESS_TARGET instead of SDL_TEXTUREACCESS_STREAMING and do surface copying without looping though the CPU, but it's unclear to me what the downside of setting all textures to SDL_TEXTUREACCESS_TARGET would be. texture->texture transfer is also only required in a few niche cases where the game copies surfaces on to other ones to animate them instead of simply rendering two instances of the same surface (like the blinking green V in the registration book), if that was changed we could change to SDL_TEXTUREACCESS_STATIC which is the optimal type, even though this would lower miniwin's DirectDraw compliance... but if another project wants to use miniwin they can dig for a less optimized and more compatible version in git.

### DIFF
--- a/LEGO1/omni/src/video/mxdisplaysurface.cpp
+++ b/LEGO1/omni/src/video/mxdisplaysurface.cpp
@@ -61,11 +61,6 @@ void MxDisplaySurface::ClearScreen()
 		backBuffers = m_videoParam.GetBackBuffers() + 1;
 	}
 
-	MxS32 width = m_videoParam.GetRect().GetWidth();
-	MxS32 height = m_videoParam.GetRect().GetHeight();
-
-	RECT rc = {0, 0, width, height};
-
 	memset(&desc, 0, sizeof(desc));
 	desc.dwSize = sizeof(desc);
 	if (m_ddSurface2->GetSurfaceDesc(&desc) != DD_OK) {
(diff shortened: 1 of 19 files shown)


foxtacles:
The Emscripten version doesn't start anymore for me. For some reason it seems it's trying to launch SDL3_GPU renderer, which obvsiouly fails. I do have this in the `.ini` and everything works fine on `master` branch:

`3D Device ID=0 0x682656f3 0x0 0x0 0x2000000`

Windows build launches fine with software

AJenbo:
Can you tell if it's just during launch where it enumerates the renderers? Or is it actually selecting it and trying to run the game with it? Any errors messages?

foxtacles:
> Can you tell if it's just during launch where it enumerates the renderers? Or is it actually selecting it and trying to run the game with it? Any errors messages?

Looks like it's not selecting it, the error message is just from the enumeration.

No error messages, logs look like this and after that nothing happens:

image

foxtacles:
image

regbook doesn't show names (windows software render)

foxtacles:
Emscripten launch gets stuck in this function call: https://github.com/AJenbo/isle-portable/blob/2d/LEGO1/omni/src/video/mxdisplaysurface.cpp#L348 because `fmt.dwRGBAlphaBitMask` is `0`

AJenbo:
Trying to get all the formats aligned isn't worth it tbh. OpenGL is already down to 6% CPU and the only reason the Software renderer isn't there also is because it's using the CPU...

PR #320, opened 2025-06-16
Fix an optimiaze OpenGL rendere

AJenbo:
Broken in https://github.com/isledecomp/isle-portable/commit/2828452b35112d3bd21efab634991453c47129df when size was changed from 4byte to 2byte indexes.

This also disables  blending when not needed which would help performance on some slow hardware.


miniwin/src/d3drm/backends/opengl1/renderer.cpp
```diff
@@ -271,7 +271,7 @@ GLMeshCacheEntry GLUploadMesh(const MeshGroup& meshGroup, bool useVBOs)
 		glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, cache.ibo);
 		glBufferData(
 			GL_ELEMENT_ARRAY_BUFFER,
-			cache.indices.size() * sizeof(Uint32),
+			cache.indices.size() * sizeof(Uint16),
```
foxtacles:
```suggestion
			cache.indices.size() * sizeof(cache.indices[0]),
```

foxtacles (approved):

PR #322, opened 2025-06-17
Fix UB in software renderer

foxtacles:
One cannot access elements of a vector that don't exist. `reserve` allocates memory, but doesn't add elements.

`D3DRMVERTEX& dst = m_transformedVerts[i];` is UB and crashes on Windows debug build

### DIFF
--- a/miniwin/src/d3drm/backends/software/renderer.cpp
+++ b/miniwin/src/d3drm/backends/software/renderer.cpp
@@ -711,9 +711,8 @@ void Direct3DRMSoftwareRenderer::SubmitDraw(
 	// Pre-transform all vertex positions and normals
 	m_transformedVerts.clear();
 	m_transformedVerts.resize(mesh.vertices.size());
-	for (size_t i = 0; i < mesh.vertices.size(); ++i) {
-		const D3DRMVERTEX& src = mesh.vertices[i];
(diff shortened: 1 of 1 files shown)


miniwin/src/d3drm/backends/software/renderer.cpp
```diff
@@ -710,7 +710,7 @@ void Direct3DRMSoftwareRenderer::SubmitDraw(
 
 	// Pre-transform all vertex positions and normals
 	m_transformedVerts.clear();
-	m_transformedVerts.reserve(mesh.vertices.size());
+	m_transformedVerts.resize(mesh.vertices.size());
```
AJenbo, line 713: this is faster since resize() will cause it to init all elements.
```
m_transformedVerts.reserve(mesh.vertices.size());
for (const auto& src : mesh.vertices) {
	D3DRMVERTEX dst;
	dst.position = TransformPoint(src.position, modelViewMatrix);
	dst.normal = src.normal;
	dst.texCoord = src.texCoord;
	m_transformedVerts.push_back(dst);
}
```

foxtacles: Since we are micro optimizing now I made it even faster

AJenbo (approved):

PR #324, opened 2025-06-17
Flip UseJoystick and JoystickIndex types

VoxelTek:
In the config tool, `usejoystick` is supposed to be a `bool`, and `joystickindex` is supposed to be an `int`. However, when writing to `isle.ini`, the types are mistakenly flipped.

This PR fixes this, and flips them back around.


AJenbo:
Nice, I wonder if there is a similar issue with bite depth, it shows up as "false" in my config which seems broken. Or maybe it was meant as 16bit on/off?

PR #327, opened 2025-06-17
Support for international versions of `isle`

### LINKED ISSUE #303: Allow special characters outside A-Z to be selected in the registration book
Doesn't work in non-English versions

foxtacles:
Close #303 

This allows the game to work with any language asset pack in the Registration Book and History Book. Presently, characters native to certain languages cannot be selected/used. The PR fixes that. Only 3 languages add new characters (Danish, German, Spanish)

Internationalization implemented for, and verified to work with original save games of that language:

- [X] Danish

image

- [X] German

image

- [X] Spanish

image

The original English game executables will crash when loading save files that contain international characters beyond the English alphabet. `isle-portable` will fix this problem and fall back to a placeholder `X` character in case the international character isn't present in the current version of the game.

The goal is to allow any save game to be compatible with any language. The only drawback (and something we can't fix without breaking save file compatibility) is that, when switching between multiple languages, extra characters of one language may show up as different extra characters of another language instead of the generic `X` placeholder. This is because we don't know which language the save file originates in, and there's no way to otherwise distinguish the alphabet values since the save file only stores indices into the alphabet array.

### DIFF
--- a/LEGO1/lego/legoomni/include/historybook.h
+++ b/LEGO1/lego/legoomni/include/historybook.h
@@ -46,6 +46,8 @@ private:
 	MxStillPresenter* m_name[20][7]; // 0x164
 
 	MxStillPresenter* m_scores[20]; // 0x394
+
+	MxStillPresenter* m_intAlphabet[sizeOfArray(LegoGameState::g_intCharacters)];
 };
 
 #endif // HISTORYBOOK_H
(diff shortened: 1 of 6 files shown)


AJenbo (commented):
Åhh (Uhh)

AJenbo (approved):

PR #328, opened 2025-06-17
Fix Draw Cursor checkbox in config tool

VoxelTek:
The original config tool appears to be partially broken. While, like the original tool, the `Draw Cursor` checkbox is disabled when a software mode is selected, it should still be adjustable in hardware modes. As such, the code for the config tool disables the checkbox and forces `m_draw_cursor` to be `false` when in a software mode. However, when a hardware mode is selected, while it re-enables the checkbox, it forces it to always be `true`.

https://github.com/user-attachments/assets/d812dd65-a3ed-4ee6-9eca-8159e76627ce

Additionally, I believe that the Draw Cursor checkbox being disabled in software mode is no longer relevant, and as such, this adjusts the code so that the checkbox is always enabled, and is always able to be checked or unchecked, regardless of video mode.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,6 +50,8 @@ message(STATUS "Internal miniwin:       ${ISLE_MINIWIN}")
 message(STATUS "Isle debugging:         ${ISLE_DEBUG}")
 message(STATUS "Compile shaders:        ${ISLE_COMPILE_SHADERS}")
 
+message(STATUS "Fetching SDL3, please wait...")
+
 if (DOWNLOAD_DEPENDENCIES)
     # FetchContent downloads and configures dependencies
     include(FetchContent)
--- a/CONFIG/MainDlg.cpp
+++ b/CONFIG/MainDlg.cpp
@@ -121,6 +121,10 @@ void CMainDialog::OnList3DevicesSelectionChanged(int selected)
 {
 	LegoDeviceEnumerate* device_enumerator = currentConfigApp->m_device_enumerator;
 	device_enumerator->GetDevice(selected, currentConfigApp->m_driver, currentConfigApp->m_device);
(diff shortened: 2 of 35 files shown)


VoxelTek:
It is also worth mentioning that the `Draw Cursor` option appears to have no effect on modern platforms, regardless of renderer selected. As such, it may be worth removing the option entirely.

VoxelTek:
Another change made was the addition of a line to the cmake script, so that users can know what's happening when the cmake output freezes for a short while, and are less inclined to believe it's crashed.

AJenbo:
> It is also worth mentioning that the Draw Cursor option appears to have no effect on modern platforms, regardless of renderer selected. 

This was intentionally Left unimpressed for all except 16bit platforms as it's intended to aid for voodoo 1/2 users , since it was not able to draw the os mouse.

VoxelTek:
I'm sorry, I don't quite understand. What do you mean by "unimpressed"?

AJenbo:
Unimplemented (autocorrect).

It doesn't have status indication and is one frame delayed compared to the os one and there is no code for turning off the os one so systems with the ability to draw 2d and 3d ends up with both.
At the time voodoo was the only common consumer GPU, so this explains why it's off for emulators but on for hardware.

The reason it doesn't do anything is because your not running in 16bit mode which was what the voodoo card did.

VoxelTek:
Given this, would it not make sense to remove the option, as it's no longer relevant? Or are you saying it should be kept for 16-bit reasons?

AJenbo:
Removing it would likely make sense, at the very least it should not be on by default for any one.

The config tool is unlikely to be backported to a platform where voodoo works, but the game could.

VoxelTek:
I might make a new PR, exclusively to remove the `Draw Cursor` option.

PR #329, opened 2025-06-17
Use extensions to probe for 24 or 32bit depth buffer

AJenbo:
OpenGL ES 2.0 has optional extensions for 24 and 32bit depth buffer that can be proved for at runtime


foxtacles (approved):

AJenbo:
Unfortunately I don't appear to have any hardware with this extension enabled, so best effort.

PR #330, opened 2025-06-18
Remove Draw Cursor checkbox from config tool

VoxelTek:
The `Draw Cursor` option has no effect on modern platforms, and the checkbox has been broken in the config tool. As such, this PR removes the checkbox entirely.


VoxelTek:
image

PR #331, opened 2025-06-18
Show that CMake is grabbing SDL3

VoxelTek:
During the `cmake ..` process, during the first run, CMake grabs SDL3. This process takes a moment, and CMake shows no indication that anything is happening, which could lead to users believing the process has crashed or otherwise frozen.

This PR adds a simple status message to show what it is doing.

Before:
image

After:
image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,13 +50,9 @@ message(STATUS "Internal miniwin:       ${ISLE_MINIWIN}")
 message(STATUS "Isle debugging:         ${ISLE_DEBUG}")
 message(STATUS "Compile shaders:        ${ISLE_COMPILE_SHADERS}")
 
-
-
 if (DOWNLOAD_DEPENDENCIES)
     # FetchContent downloads and configures dependencies
-
     message(STATUS "Fetching SDL3 and iniparser. This might take a while...")
-
     include(FetchContent)
     FetchContent_Declare(
         SDL3


CMakeLists.txt
```diff
@@ -50,6 +50,8 @@ message(STATUS "Internal miniwin:       ${ISLE_MINIWIN}")
 message(STATUS "Isle debugging:         ${ISLE_DEBUG}")
 message(STATUS "Compile shaders:        ${ISLE_COMPILE_SHADERS}")
 
+message(STATUS "Fetching SDL3, please wait...")
+
```
madebr: This message should only be displayed when downloading dependencies.
When you use a pre-built SDL3, this message is meaningless.

So move it under `if (DOWNLOAD_DEPENDENCIES)`.

Perhaps also change the message to something like `"Fetching SDL3 and iniparser. This might take a while..."`.
Unfortunately, I don't see a way to not display this message when re-configuring the project.

foxtacles: Related to this, but can we skip downloading imgui if debug is disabled?

VoxelTek: > Related to this, but can we skip downloading imgui if debug is disabled?

Looks like that might already be the case. `target_link_libraries(isle PRIVATE imgui)` is only run if `ISLE_DEBUG` is `ON`. Then again, I'm not that familiar with CMake, so I could be missing something.

madebr: imgui is a git submodule, so it is currently always available (miniaudio is also a submodule and is unconditionally used)
Do you want imgui added here and conditionally do `FetchContent`?

VoxelTek: > So move it under `if (DOWNLOAD_DEPENDENCIES)`.
> 
> Perhaps also change the message to something like `"Fetching SDL3 and iniparser. This might take a while..."`. 

Done!

CMakeLists.txt
```diff
@@ -50,8 +50,13 @@ message(STATUS "Internal miniwin:       ${ISLE_MINIWIN}")
...
 message(STATUS "Compile shaders:        ${ISLE_COMPILE_SHADERS}")
 
+
+
 if (DOWNLOAD_DEPENDENCIES)
     # FetchContent downloads and configures dependencies
+
+    message(STATUS "Fetching SDL3 and iniparser. This might take a while...")
+
     include(FetchContent)
```
madebr: Reduce whitespace.
```suggestion
message(STATUS "Compile shaders:        ${ISLE_COMPILE_SHADERS}")

if (DOWNLOAD_DEPENDENCIES)
    # FetchContent downloads and configures dependencies
    message(STATUS "Fetching SDL3 and iniparser. This might take a while...")
    include(FetchContent)
```

PR #332, opened 2025-06-18
Implement DirectX 9 backend

AJenbo:
This implements a DirectX 9.0 (the original) backend for miniwin (it's as if Microsoft never stopped supporting D3DRM). The main motivation here is to make the project compatible with RTX Remix which only functions with this specific version or various wrappers. It does less to cull polygons as that's something that actually works against what RTX Remix tries to do. It might be beneficial to also add a switch to disable object culling in the Viewport.

Additionally this also helps dial in the highlights since the lighting model pretty much fitted like a hand in a glove.

I had to make an abstraction to avoid windows.h fighting with miniwin's implementations, and this also involved a few `void *` cats from one type to another, coupled with the fact that I can't run ASAN or UBSAN under Wine I likely made a couple of mistakes so it would be helpful if someone on Windows could test this out with some analyzers enabled to see if it detects some obvious issues. In fact I'm sure I made at least one mistak as some objects appear to be swapped sometimes but maybe that's just because I'm not freeing meshes currently.

Lastly this PR also contains a MinGW64 cross compiler toolchain that can be used to build the project from Linux:
```cpp
mkdir build-w64
cd build-w64
cmake .. -DCMAKE_TOOLCHAIN_FILE=../CMake/cmake-toolchain-mingw64-x86_64.cmake -DISLE_BUILD_CONFIG=OFF
make -j32
```

image

image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,7 @@ if (NOT MINGW)
   set(NOT_MINGW ON)
 else()
   set(NOT_MINGW OFF)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -static -static-libgcc -static-libstdc++")
 endif()
 
 find_program(SDL_SHADERCROSS_BIN NAMES "shadercross")
--- a/CMake/cmake-toolchain-mingw64-x86_64.cmake
+++ b/CMake/cmake-toolchain-mingw64-x86_64.cmake
@@ -1,0 +1,18 @@
+set(CMAKE_SYSTEM_NAME Windows)
+set(CMAKE_SYSTEM_PROCESSOR x86_64)
+
+find_program(CMAKE_C_COMPILER NAMES x86_64-w64-mingw32-gcc)
+find_program(CMAKE_CXX_COMPILER NAMES x86_64-w64-mingw32-g++)
+find_program(CMAKE_RC_COMPILER NAMES x86_64-w64-mingw32-windres windres)
+
+if(NOT CMAKE_C_COMPILER)
+	message(FATAL_ERROR "Failed to find CMAKE_C_COMPILER.")
+endif()
+
+if(NOT CMAKE_CXX_COMPILER)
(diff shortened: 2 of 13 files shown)


CMake/mingwcc64.toolchain.cmake
```diff
@@ -0,0 +1,21 @@
...
+
+SET(CMAKE_SYSTEM_NAME Windows)
+
+# workaround
+list(APPEND CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES "${CROSS_PREFIX}/x86_64-w64-mingw32/include")
+
+SET(CMAKE_C_COMPILER   "x86_64-w64-mingw32-gcc")
+SET(CMAKE_CXX_COMPILER "x86_64-w64-mingw32-g++")
+set(CMAKE_RC_COMPILER "x86_64-w64-mingw32-windres")
+set(CMAKE_STRIP "${CROSS_PREFIX}/x86_64-w64-mingw32/bin/strip")
```
madebr: This toolchain file is debian/ubuntu-specific.
What do you think about borrowing the CMake toolchain files from SDL?
https://github.com/libsdl-org/SDL/blob/main/build-scripts/cmake-toolchain-mingw64-x86_64.cmake

CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ if (NOT MINGW)
   set(NOT_MINGW ON)
 else()
   set(NOT_MINGW OFF)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -static -static-libgcc -static-libstdc++")
```
madebr, line 25: Doesn't this belong in the CMake toolchain file?

AJenbo: I was thinking it also made sens for anyone building with MSYS2, is it not affected by this line?

foxtacles (approved):
LGTM, I ran it with UBSAN and ASAN for a few minutes of gameplay and no issues

Issue #312, opened 2025-06-14
Update Config with the new options


AJenbo:
Models: low, medium, high
LOD: range
Actors: Range
Language?

foxtacles:
> Language?

This setting is only used by Emscripten and it's not part of isle, since all it needs to control is which assets are served.  In an offline scenario the config tool can just change the diskpath to a different language asset pack and that's it

AJenbo:
Path selection would be good

foxtacles:
Yeah, I suppose `CONFIG` can be turned into a launcher/mini-installer similar to isle.pizza. Then it can also do some verification that all the necessary files etc. are present

VoxelTek:
Since #333 basically covers this, should this issue be closed now?

PR #333, opened 2025-06-18
Remake config tool, adding new settings and removing some obsolete ones

VoxelTek:
image

Mainly regarding #312 
- Addition of path editing
- Add "Low" model quality option (previously only accessible if the platform had limited RAM)
- Addition of LoD (Level of Detail) and Max Actors sliders
- Removal of `Draw Cursor` (already done in #330)
- Shrinking of `Direct 3D Devices` box (basically only 4 options on modern platforms anyway)

### DIFF
--- a/CONFIG/MainDlg.cpp
+++ b/CONFIG/MainDlg.cpp
@@ -130,6 +130,9 @@ bool CMainDialog::OnInitDialog()
 		driver_i += 1;
 	}
 	m_ui->devicesList->setCurrentRow(selected);
+
+	m_ui->maxLoDSlider->setValue((int)currentConfigApp->m_max_lod * 10);
+	m_ui->maxActorsSlider->setValue(currentConfigApp->m_max_actors);
 	UpdateInterface();
 	return true;
 }
@@ -353,9 +356,12 @@ void CMainDialog::SelectDiskPathDialog()
                                                     disk_path,
                                                     QFileDialog::ShowDirsOnly
                                                     | QFileDialog::DontResolveSymlinks);
-	currentConfigApp->m_base_path = disk_path.toStdString();
-	m_modified = true;
-	UpdateInterface();
+
+	if (disk_path.toStdString() != "") {
+		currentConfigApp->m_base_path = disk_path.toStdString();
+		m_modified = true;
+		UpdateInterface();
+	}
 }
 
 void CMainDialog::SelectCDPathDialog()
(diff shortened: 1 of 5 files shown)


AJenbo:
> * basically only 4 options on modern platforms anyway)

I can try and make some more if the low number is a problem 😄

AJenbo:
Could you remove "16bit" it now just mirror the desktop color setting ... None of the current drives support 8-bit, even if forced performance is bad on 256 colors

kernaltrap8:
Awesome!

VoxelTek:
> Could you remove "16bit" it now just mirror the desktop color setting ... None of the current drives support 8-bit, even if forced performance is bad on 256 colors

So just remove the colour depth setting altogether? I can do that.

AJenbo:
Yeah that would probably make the most sens

AJenbo:
"Draw 3D to Video Memory" also doesn't do anything so can also be removed.

VoxelTek:
> "Draw 3D to Video Memory" also doesn't do anything so can also be removed.

What about Flip Video Memory Pages, should that also be dropped?

AJenbo:
I prefer the applications behavior when it's enabled, but I mapped both states to the same result, so um yeah probably just remove it. Hopefully the code that it affects will be removed since it has not real effect on portable.

VoxelTek:
> I prefer the applications behavior when it's enabled, but I mapped both states to the same result, so um yeah probably just remove it. Hopefully the code that it affects will be removed since it has not real effect on portable.

Currently it's forced to always be on anyway so...

AJenbo:
Feel free to set "Use joystick" to true by default.

VoxelTek:
> Feel free to set "Use joystick" to true by default.

Done, as are the other recommended changes.

AJenbo:
@VoxelTek if you want to tackle a few more change i have tagged issues related to the config application here: https://github.com/isledecomp/isle-portable/issues?q=is%3Aissue%20state%3Aopen%20label%3Aconfig.exe

It turns out the Media Path doesn't do anything so that can actually be removed.

VoxelTek:
> @VoxelTek if you want to tackle a few more change i have tagged issues related to the config application here: https://github.com/isledecomp/isle-portable/issues?q=is%3Aissue%20state%3Aopen%20label%3Aconfig.exe
> 
> It turns out the Media Path doesn't do anything so that can actually be removed.


For #398 I think I'll add tooltips, so that you can hover over an option and get the popup text. I've already done that for a few of the options (LOD, max actors).

In regards to #339, I was under the mistaken impression that we *had* to play in full-screen, for some reason. But yeah this is an obvious option to add.

As for #337, I really don't know anything about CI building and whatnot, but it should be simple enough to get it to build with the rest of `isle`.

I'll get started on this, and make another PR when I'm done.

AJenbo:
> As for https://github.com/isledecomp/isle-portable/issues/337, I really don't know anything about CI building and whatnot, but it should be simple enough to get it to build with the rest of isle.

Yeah I got a bit frustrated at getting CPack to pack it with it's dependencies, don't worry about that one if it's not up your ally.

PR #340, opened 2025-06-19
Fix memory leaks in config application

AJenbo:


foxtacles (approved):

AJenbo:
The rest seems to be internal to SDL or obscured by threading

PR #349, opened 2025-06-20
Fix 'incomplete type ‘LegoGameState’ used in nested name specifier'

madebr:
This fixed a build error introduced by 6cdbfa86a:
`sizeOfArray(LegoGameState::g_intCharacters)` needs to see the declaration of the `LegoGameState` type


foxtacles:
ah, I guess the PCH was hiding this

foxtacles (approved):

madebr:
> ah, I guess the PCH was hiding this

Yup. I disabled it locally. PCH is useful for one-time builds, but not useful when you regularly modify the headers.

PR #350, opened 2025-06-20
miniwin: fix OpenGL include on macOS

Ramen2X:
Apple doesn't conform to the otherwise-universal standard and places OpenGL headers in `OpenGL` instead of `GL`, which causes macOS compilation to fail.


AJenbo (approved):

AJenbo:
Thanks

PR #352, opened 2025-06-20
Check for files

### LINKED ISSUE #123: No clear error message when some files are missing

foxtacles:
image

Close #123


AJenbo:
Nice work :)

Issue #355, opened 2025-06-21
Can't locate CREDITS.SI


AJenbo:
Linux

It's there:
![Image](https://github.com/user-attachments/assets/6d624453-cf4e-42ea-ae6f-cc67dca46fd9)

Found 58 game files in /home/ajenbo/.local/share/isledecomp/isle/cd/

StevenSYS:
I found the issue!

SDL checks if the relative path to `CREDITS.SI` exists instead of the absolute path

Relative Path: `LEGO/Script/CREDITS.SI`
Absolute Path Example: `/home/user/.local/share/isledecomp/isle/LEGO/Scripts/CREDITS.SI`

AJenbo:
Yeah I think the current code only works if you dump you cd in the build folder, yet the mapper takes care to also check the cd and HD path which leads to a miss match between how the two checks for files. I made a PR to address it.

PR #357, opened 2025-06-21
Fix loading files from CD or HD

AJenbo:
Fixes https://github.com/isledecomp/isle-portable/issues/355


AJenbo:
Merging because of recent influx of new users

PR #373, opened 2025-06-21
Correct link to Actions yml file in README

Lenni-builder:
This just corrects a link in the README to point to ci.yml instead of build.yml (which is a 404).


AJenbo:
Thanks

Issue #366, opened 2025-06-21
Please provide Debian installer and possibly packages for other Linux distors


dldropper:
OS: Linux Mint
liblego.so is in lib folder.
log: ./isle
./isle: error while loading shared libraries: liblego1.so: cannot open shared object file: No such file or directory

ry-diffusion:
you need to set `LD_LIBRARY_PATH` to make the game able to find the library.
Like this:
`LD_LIBRARY_PATH=$(pwd)/lib ./bin/isle`

dldropper:
how do i do that?
libraries are in ~/Applications/isle-0.1-Linux-x86_64-x86_64/lib
executable is in ~/Applications/isle-0.1-Linux-x86_64-x86_64/bin

Mystery24601:
It resulted in this error:
```
./bin/isle: error while loading shared libraries: libGLEW.so.2.2: cannot open shared object file: No such file or directory
```

dldropper:
mystery24601 is libglew installed?

dldropper:
if not, you need to install it.

Mystery24601:
Should probably be included with the linux build, thanks for the info though

dldropper:
ok i figured it out, just put this at the end of ~/.bashrc
export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/home/dldropper/Applications/isle-0.1-Linux-x86_64-x86_64/lib

ArjixWasTaken:
instead of modifying your bashrc, I'd suggest either creating a launch script, or adding the environment variable to the desktop file (assuming you have one)

AJenbo:
> thanks for the info though

It's a system lib so not really

Mystery24601:
Real issue is, why isn't this included with the Linux build? (Continuation Build)

itsmattkc:
The Linux build should probably be replaced with an AppImage or something like that, it's not like Windows where you can put a few files together and expect it to work on other systems.

dldropper:
Mattkc app image doesn't work on all distros

dldropper:
Flatpaks do though

foxtacles:
I think it's clear more effort needs to be put into release builds to ideally make them work out of the box on as many systems as possible. Right now it's really just a copy of the CI build artifacts.

ArjixWasTaken:
[...]

PR #379, opened 2025-06-21
Package for Debian-based distributions

ItzSwirlz:
Closes: #366 

I moved the install section a bit later so we ensure that we don't include our built SDL3 in our package, otherwise we will overwrite the system sdl3

I put myself down as the maintainer.

My only thing with this is that we *could* use a [Personal Package Archive](https://launchpad.net/ubuntu/+ppas) (PPA) that then users can add to their apt source list, and then install with `sudo apt install isle`. Then, when we push an update (new backend or whatnot), they can `sudo apt upgrade isle`.

If we did that, we would have to maintain our own *actual* Debian packaging (debian/control, Debian/copyright, debian/rules, all that fun stuff). I can do that, I just don't know if anyone here would be okay with maintaining those files. But we would need to do this in order to actually upload to a PPA (or any archive for that matter), otherwise people will need to reinstall the debs every update from GitHub

---------

TODO: .desktop file entries, but I guess maybe we can save that for another PR since it would probably rely on an `Assets` folder for us to store icons (which we could really use tbh) and generating the executable file path

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -656,8 +656,11 @@ elseif(DPKG)
   # Don't replace system SDL3
   list(REMOVE_ITEM install_extra_targets "SDL3-shared")
 
-  # "config" is a vague binary name
-  set_property(TARGET config PROPERTY OUTPUT_NAME "isle-config")
+
+  if(ISLE_BUILD_CONFIG)
+    # "config" is a vague binary name
+    set_property(TARGET config PROPERTY OUTPUT_NAME "isle-config")
+  endif()
 
   set(CPACK_GENERATOR DEB)
 else()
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -644,9 +640,36 @@ endif()
...
+  set(CPACK_DEBIAN_PACKAGE_DESCRIPTION "LEGO Island for Debian-based distros")
+  set(CPACK_DEBIAN_PACKAGE_VERSION ${PROJECT_VERSION})
+  set(CPACK_DEBIAN_PACKAGE_DEPENDS "libqt6core6t64 libsdl3-0")
+  set(CPACK_DEBIAN_PACKAGE_SECTION games)
+  set(CPACK_DEBIAN_PACKAGE_PRIORITY optional)
+  set(CPACK_DEBIAN_PACKAGE_HOMEPAGE "https://github.com/isledecomp/isle-portable")
+  set(CPACK_DEBIAN_PACKAGE_MAINTAINER "Joshua Peisach <itzswirlz2020@outlook.com>")
+
+  # Don't replace system SDL3
+  list(REMOVE_ITEM install_extra_targets "SDL3-shared")
```
AJenbo, line 657: It's build on Ubuntu 24.04 which doesn't even have a package for SDL3 yet, that feels a bit problematic.

jpeisach: True.. that would move us to 25.04 or later

PR #395, opened 2025-06-22
Add option to remove startup delay

carson-coder:
Added a cmake option (ISLE_STARTUP_DELAY) to allow for disabling the startup delay (startup delay enabled by default)

### DIFF
--- a/ISLE/isleapp.cpp
+++ b/ISLE/isleapp.cpp
@@ -908,8 +908,13 @@ inline bool IsleApp::Tick()
 	// GLOBAL: ISLE 0x4101c0
 	static MxLong g_lastFrameTime = 0;
 
+#ifdef STARTUP_DELAY
 	// GLOBAL: ISLE 0x4101bc
 	static MxS32 g_startupDelay = 200;
+#else
+	// GLOBAL: ISLE 0x4101bc
+	static MxS32 g_startupDelay = 1;
+#endif
 
 	if (IsleDebug_Paused() && IsleDebug_StepModeEnabled()) {
 		IsleDebug_SetPaused(false);
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,6 +28,7 @@ endif()
 find_program(SDL_SHADERCROSS_BIN NAMES "shadercross")
 find_package(Python3 3.11 COMPONENTS Interpreter)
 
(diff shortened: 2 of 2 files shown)


ISLE/isleapp.cpp
```diff
@@ -908,8 +908,13 @@ inline bool IsleApp::Tick()
 	// GLOBAL: ISLE 0x4101c0
 	static MxLong g_lastFrameTime = 0;
 
+#ifdef STARTUP_DELAY
```
foxtacles: I don't see the usefulness in making this a build config option. If you are confident that reducing the startup delay to 1 will not break start up on any platforms you can reduce it to that.

carson-coder: Ok

ISLE/isleapp.cpp
```diff
@@ -908,8 +908,7 @@ inline bool IsleApp::Tick()
 	// GLOBAL: ISLE 0x4101c0
 	static MxLong g_lastFrameTime = 0;
 
-	// GLOBAL: ISLE 0x4101bc
```
foxtacles, line 911: Please keep this comment

foxtacles:
@Ramen2X I know rebuilder has an option to remove the delay - is this the same and did you ever encounter any issues with it?

ISLE/isleapp.cpp
carson-coder: I accidentally deleted it whoops.

Ramen2X:
> @Ramen2X I know rebuilder has an option to remove the delay - is this the same and did you ever encounter any issues with it?

Yes, the code in this PR is functionally equivalent.

https://github.com/isledecomp/LEGOIslandRebuilder/blob/master/lib/dllmain.cpp#L58

It never caused issues for any of us in our testing; however, it was never tested on machines from the era. One of the original developers' confirmed to us that the intent of the 200 frame delay was a hacky way to fix a race condition in ISLE initialization, so it's entirely possible the race condition is dependent on speed and isn't hit on faster machines by nature.

AJenbo:
I'm hoping some of the thread sanitizer detected issues will turn out to be the cause of the issue.

Right now it's bad enough that TSAN craps itself without extra options, so I haven't added it to the setup before we can get it to a point where it's more generally usable.

Issue #213, opened 2025-06-01
FPS counter


AJenbo:
Miniwin does not have text support. Adding something like SDL_ttf seems like a huger overkill for this. The ingame sprite font seems to only have A-Z. Could we put it in the Window title instead? Print it to the terminal?

StevenSYS:
Couldn't you add an FPS counter to the debug window?

kernaltrap8:
I tried making a FPS counter a while ago, my code sucked though and wasn't accurate. It was showing like 1000FPS in terminal, but comparing with the overlay in Steam, it was more like 88FPS. I'd say either what Steven said, put into the ImGUI debug window, or if thats disabled, put it in titlebar.

PR #420, opened 2025-06-23
Implement DrawFPS

AJenbo:
Fixes https://github.com/isledecomp/isle-portable/issues/213

This is very useful for full screen and stops spamming the console

image

There is still some font code used in LEGO1/mxdirectx/mxdirectdraw.cpp which I'm not sure how to trigger so I haven't replaced that and so we can't quite strip the stubs from miniwin yet.


LEGO1/lego/legoomni/src/video/legovideomanager.cpp
```diff
@@ -403,40 +403,70 @@ inline void LegoVideoManager::DrawCursor()
 		->BltFast(m_cursorXCopy, m_cursorYCopy, m_cursorSurface, &m_cursorRect, DDBLTFAST_WAIT | DDBLTFAST_SRCCOLORKEY);
 }
 
+// 4x5 bitmap font
+const uint8_t digitFont[5][10] = {
```
foxtacles: Can you move new code / functions to the bottom of the compilation unit? That's what we have generally done with new stuff. It should help alleviate some merge conflict work when updating from the decomp

AJenbo: And forward declare it? Do you have an example?

foxtacles: I'd add them as member functions to the header at the end of the existing function declarations. Just to stay in style with the current code base (which is 95% class/object based):

image

AJenbo: We are from different worlds, I feel tempted to just inline it :)

foxtacles (approved):

PR #426, opened 2025-06-24
Allow running game without audio device

### LINKED ISSUE #421: Failed to Create LegoSoundManager when no audio device connected (Win 11)
I was testing the game on a vga monitor without a sound device and saw in the command line the title's error, alongside a generic "Lego Island failed to start" and when i plugged in audio, the game started fine after.

`Found 62 game files in C:\Users\ldroberts\Desktop\Lego Island Portable\
`
`Found 691 game files in F:
`
`ERROR: Failed to create LegoSoundManager
^C
`

foxtacles:
Close #421

The game simply won't play any audio with this change. Maybe in the future we can also look into allowing hotplugging an audio device, but I feel like the use case for this would be pretty niche.

### DIFF
--- a/LEGO1/omni/src/audio/mxsoundmanager.cpp
+++ b/LEGO1/omni/src/audio/mxsoundmanager.cpp
@@ -105,7 +105,11 @@ MxResult MxSoundManager::Create(MxU32 p_frequencyMS, MxBool p_createThread)
 		SDL_ResumeAudioDevice(SDL_GetAudioStreamDevice(m_stream));
 	}
 	else {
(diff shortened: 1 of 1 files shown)


madebr (approved):
Would it be hard to disable reading/streaming audio when audio is disabled?
This might reduce load on already-constrained devices (probably not much).

foxtacles:
> Would it be hard to disable reading/streaming audio when audio is disabled?

Doable, but not trivial, would require changes in multiple places - but something to be considered in the future

AJenbo:
The event system should probably be made to listen for new audio devices and connect to it if it's not already connected to one.

Issue #398, opened 2025-06-22
Config: explain the different options


gwuen:
First of all, great project! I'm really amazed how simple it was to compile this under Linux and have LEGO Island up and running natively in under ten minutes. Thank you all for that :)

The only hurdle I encountered was the various configuration options, which unfortunately are not explained any further. By looking at the error messages I was quickly able to figure out the correct `diskpath` and `cdpath`, but I'm still unsure what `mediapath` is meant to point to (using the ISO from the LEGO Island Wiki). I have really looked around everywhere (not only in this GH project) but couldn't find any explanation, even though a `mediaPath` doesn't seem to be exclusive to this portable version.

So for one, I would really appreciate it if someone could explain to me what the media path is supposed to be. For another, I think it would be good to have short explanations or examples in the config GUI and/or the README for new users.

foxtacles:
`mediapath` has no function whatsoever. It's a leftover from the original decompilation. We should probably remove it from config and maybe from the INI file altogether.

gwuen:
Sounds like a good idea, thanks for the quick response!

What do you think of adding tooltips to the Disk Path and CD Path options? I could also (later) open a PR for that if you're not already on it. Other options like the Direct 3D Devices (that's the different backends for the d3drm translation layer, right?) could also use some kind of explanation, but I guess it's not that important as the default works ootb.

foxtacles:
Tooltips would be great - I don't think anybody is working on it right now. There's also other config related issues: https://github.com/isledecomp/isle-portable/issues/339

> (that's the different backends for the d3drm translation layer, right?)

Yes

VoxelTek:
Now that #427 has been merged, I'd say this issue is complete and can be marked as resolved.

Issue #339, opened 2025-06-19
Add fullscreen toggle to config application


AJenbo:

VoxelTek:
Now that #427 has been merged, I believe this issue can be marked as resolved.

AJenbo:
Thanks for the notice

PR #427, opened 2025-06-25
Revamp config tool, add tooltips and consolidate options.

VoxelTek:
- In the config tool, the "Media Path" entry was removed, as it doesn't do anything in isle-portable as far as I know. 
- The "Disk Path" and "CD Path" options were merged into one directory editing thing, where CD Path is specified and Disk Path is then derived from that. 
- A checkbox to enable or disable full-screen was added. 
- Tooltips were added to the config app, so now users can more easily tell what a setting in the config tool does.

Continuation of #333, primarily addressing #398 and #339.

The config tool now looks like this:
image


CONFIG/config.cpp
```diff
@@ -166,10 +165,13 @@ bool CConfigApp::ValidateSettings()
 {
 	BOOL is_modified = FALSE;
 
+	/*	Forces full screen if not IsPrimaryDriver(). Not sure exactly why,
```
AJenbo: Probably because a Voodoo card would be the secound GPU in the system and cannot be used in window mode as it's 3D exclusive. Similar to how it needed a cursor. It was basically the only real consumer GPU at that point.

You can just remove the code imo. It's still in git history in the worst case.

CONFIG/config.cpp
```diff
@@ -203,10 +205,13 @@ bool CConfigApp::ValidateSettings()
 			m_3d_video_ram = TRUE;
 			is_modified = TRUE;
 		}
+		/*	Forces full screen whenever m_flip_surfaces is enabled, and it almost always is.
+		/*	This isn't relevant to isle-portable, so we can just comment it out.
 		if (!m_full_screen) {
 			m_full_screen = TRUE;
 			is_modified = TRUE;
 		}
+		*/
```
AJenbo: It makes sens because you can't flip the page if not in full screen, but it's all mostly automated by the compositor at this point so you can just remove it.

```suggestion
```

foxtacles (approved):

CONFIG/config.cpp
```diff
@@ -166,10 +165,13 @@ bool CConfigApp::ValidateSettings()
 {
 	BOOL is_modified = FALSE;
 
+	/*	Forces full screen if not IsPrimaryDriver(). Not sure exactly why,
+	/*	but it's not relevant, so we can just comment it out.
 	if (!IsPrimaryDriver() && !m_full_screen) {
 		m_full_screen = TRUE;
 		is_modified = TRUE;
 	}
+	*/
```
AJenbo:
```suggestion
```

VoxelTek:
It's also worth mentioning that this PR also includes some minor validation of directories and such, so the `Data Path` and `Save Path` are only changed if the new directories actually exist.

AJenbo (approved):

PR #434, opened 2025-06-26
1.0 compatibility

### LINKED ISSUE #364: Regarding 1.0 Compatibility
Since the project is dependent on the version of the game being 1.1 and not 1.0, I do have concerns about versions like the Japanese version, which only have a 1.0 version.

My concern is with certain animations playing at incorrect times and causing potentially game-breaking stuff.
I have the Japanese version of the game installed and I tested it with isle-portable, and this is the result.

https://github.com/user-attachments/assets/186b3bff-06a9-4e73-a932-4c659e2d25ff

At first it may look normal, but once I put in my name, it does not play the usual cutscene it's supposed to (the one where the Infomaniac is talking about jetskiing, flying and a bunch of other things). The cutscenes for exiting the game are also incorrect.

This isn't just a problem with isle-portable either, isle.pizza also has the same issue and has a Japanese language option (1.0).
I'd like to prevent further problems with people trying to use 1.0 copies of the game with this by bringing up this issue.
Is it possible to implement a flag that checks if the version of the game is 1.0/a flag that the user sets if the version is 1.0?

Thanks!

foxtacles:
- [X] Different dialogue: After entering a name in the registration book in v1.0, the Infomaniac says "Good job! Welcome to Lego Island! Okay, let's get started...". In the September build, this first sentence is cut and the Infomaniac instead starts at "Okay, let's get started!".
- [x] Building: During any of the four building sequences, Bill Ding instructs you to click the triangle button when you are finished. However in v1.1, the sequence will automatically end itself once you place the last brick, rendering the triangle button largely unnecessary. v1.0's behavior is accurate to the explanation; you must click the triangle button to finish even if you've placed all the bricks. It's generally assumed testers would miss the opening explanation and become confused about what to do once they'd completed their builds (hence the change in v1.1), But many prefer v1.0's approach since it allows them to continue making changes even after the last brick is placed.
- [x] Window title: When playing in windowed mode, the August build's window title reads "Lego Island" while the September build's reads "LEGO®". This may be the only time the game uses sentence case for the word "LEGO" instead of capitalizing it as usual.
- [x] Character selection: In v1.1, clicking a character's icon selects that character even if the icon isn't dragged to a location (v1.0 requires the icon to be dragged somewhere to set the character). This allows players to switch characters and exit the Information Center through other means, such as the back door. However it also necessitates a change in the Infomaniac's animation. In v1.0, selecting Mama or Papa Brickolini ends with him flying offscreen before the game transitions to the selected location, but since players of v1.1 can select a character without actually leaving the Information Center, his animation has been crudely modified to position him back on the floor rather than flying offscreen.

Close #364


LEGO1/lego/legoomni/src/main/legomain.cpp
```diff
@@ -76,6 +76,7 @@ void LegoOmni::Init()
 	m_bkgAudioManager = NULL;
 	m_unk0x13c = TRUE;
 	m_transitionManager = NULL;
+	m_is10 = FALSE;
```
AJenbo: isV10 or even Version10, it's really hard to think this is a bool that checks that some input was equal to 10

PR #435, opened 2025-06-26
Add Launch Game button to config tool, rename executable to isle-config

VoxelTek:
Adds a button to start `isle` from the `config` tool.

Additionally, renames the executable from `config` to `isle-config`

NOTE: This assumes that the `isle-config` executable is in the same directory as `isle` (which it pretty much always will be)

image

### DIFF
--- a/CONFIG/MainDlg.cpp
+++ b/CONFIG/MainDlg.cpp
@@ -173,17 +173,12 @@ void CMainDialog::launch()
 		QMessageBox::Close
 	);
 
-#ifdef _WIN32
-	if (!QProcess::startDetached("./isle.exe")) {
-		msgBox.exec();
-	}
-#else
 	if (!QProcess::startDetached("./isle")) {   // Check in isle-config directory
 		if (!QProcess::startDetached("isle")) { // Check in $PATH
 			msgBox.exec();
 		}
 	}
-#endif
+
 	QDialog::accept();
 }
 


kernaltrap8:
Is there a check to error-out if `isle` cannot be found? Perhaps you could also have it check PATH.

VoxelTek:
Hm, good point. I might add a simple popup if it can't find `isle`, and I'll have it try and run from the PATH if it can't find a local version.

VoxelTek:
image

ItzSwirlz (approved):
Maybe clarify that the "launch game" will or will not save changes? Other than that, thanks for renaming config to `isle-config` :)

VoxelTek:
> Maybe clarify that the "launch game" will or will not save changes? Other than that, thanks for renaming config to `isle-config` :)
@ItzSwirlz

I've added tooltips to each button, if you think that's suitable.

**Save and Exit:** Save configuration and close the config tool.
**Launch Game:** Save configuration and launch LEGO Island.
**Exit without saving:** Discard changed settings and close the config tool.

I could also rename the buttons, if you think that'd be better.

jpeisach:
Personally I would go with `Save and Launch Game` but also that's a lot of words.. what do other people think?

By the way, we should look into isle-config translations 😄

StevenSYS:
I think "Save and Launch" would be good

VoxelTek:
> Personally I would go with `Save and Launch Game` but also that's a lot of words.. what do other people think?

How about this?
image
DAMMIT @StevenSYS you posted your comment right as I was about to upload this image, beat me to it



> By the way, we should look into isle-config translations 😄

Qt provides a lot of support for that, I believe. Shouldn't be hard to do so, provided we have someone to translate the text.

AJenbo (approved):

CONFIG/MainDlg.cpp
```diff
@@ -155,6 +158,35 @@ void CMainDialog::accept()
...
+
+	QMessageBox msgBox = QMessageBox(
+		QMessageBox::Warning,
+		QString("Error!"),
+		QString("Unable to locate isle executable!"),
+		QMessageBox::Close
+	);
+
+#ifdef _WIN32
+	if (!QProcess::startDetached("./isle.exe")) {
```
AJenbo: Windows will actually launch any executable that matches the name with out the extension, you can argue that it's relevant because of .\ but you wrote ./

VoxelTek: The compiler got unhappy when I used a backslash, for some reason. And given that Linux wouldn't launch the local executable when just `isle` was used, I figured I'd do the `./` with both it and Windows

madebr: Doesn't qt have helper functions to get the folder of the running process and append files to it?

VoxelTek: ...? I'm using a function to get the directory of the `isle-config` app, and then running the `isle` executable in the same folder. If that doesn't exist, it checks the $PATH (i.e. instead of running `./isle`, it runs `isle`. I did the same `./` thing for the Windows launch command, just cause, but I can remove that if you think I should.

VoxelTek: > Windows will actually launch any executable that matches the name with out the extension

OH I didn't fully understand, so I can probably drop the `#ifdef` stuff?

AJenbo: Right, that was my point, but my teasing muddled it a bit :) `./isle` works on Windows and most other platforms.

VoxelTek: Cool, I'll modify that.

PR #437, opened 2025-06-27
Replace a lot of sprintf calls with snprintf.

MaddTheSane:
They are slightly more secure, and don't throw warnings on macOS/Clang.


AJenbo:
I would really suggest moving to something more Modern like fmt

foxtacles:
We are using SDL.

Furthermore, security "out of principle" is not in scope for this project. The code base has already diverged enough as-is from its roots. Changing code like this doesn't buy us anything in terms of portability, and only further complicates keeping the code base up-to-date with the decomp.

PR #440, opened 2025-06-28
Add experimental WebGL support to Emscripten

foxtacles:
Open GL ES 2.0 renderer works on Emscripten out of the box, except for 2D rendering:

image
image

The main modification here is that it should be possible to provide a static description for a device we know is "always" available (such as WebGL in browsers). It is not possible to enumerate devices and test renderers in Emscripten the way we do it on other platforms. Even if we knew the `SDL_Window` at the time of enumeration, this still won't work since contexts can't be created multiple times, i.e. once a OpenGL ES context is established it's not possible to go back to software renderer anymore.

The interface modification as seen in this PR essentially turns the enumeration for Emscripten into providing a static list of available devices (Software, OpenGL ES 2.0).

I'm not sure if fixing the 2D rendering would necessitate creating a completely new renderer backend, since I have no idea why it isn't working. In that case we could change this again later.


foxtacles:
> Skulde i look into 2d rendering it seems like it fails in your screen shot.

Sure, that would be great. I don't really know where to start looking. 3D seems to work perfectly.

foxtacles:
One thing I've already tried is enabling `-sFULL_ES2=1`: https://emscripten.org/docs/porting/multimedia_and_graphics/OpenGL-support.html#opengl-support-opengl-es2-0-emulation

but it didn't make a difference (except now the Infocenter background is black instead of blue)

Use of `glDrawArrays` etc. is present in the current OpenGL ES renderer so maybe we need it though.

AJenbo:
2d rendering ended up being implemented using different rendering modes for most drivers as it was easier to position vertex like you would 2d surfaces, but that's not possible for sdl_gpu so I did have to implement it using 3d transformation there, I can likely just port that over to es.

Regarding the descriptors I'm doing the same for 3ds but much less invasive so I would like to see if that's possible here as well. Really nice that you got it up and running though.

foxtacles:
> Regarding the descriptors I'm doing the same for 3ds but much less invasive so I would like to see if that's possible here as well.

Sounds great, feel free to open a PR on top of this one, or a new one. I'll leave this PR open for now.

ISLE/isleapp.cpp
```diff
@@ -89,6 +91,21 @@ SDL_Window* window;
 
 extern const char* g_files[46];
 
+#ifdef __EMSCRIPTEN__
```
foxtacles: I'd say let's leave out Fullscreen support until they fixed the bug in SDL. I'd rather not have this hack and I think we can go without FS for now (you can have "windowed" Fullscreen anyway with just using CSS)

AJenbo: oh didn't mean to push that here, it's all still very rough.

AJenbo:
Things looks grate now when this is enabled, mostly it was just missing proper texture initialization which WebGL is very strict about, except for missing transitions.
The issue is that `glReadPixels()` isn't able to read from the canvas in WebGL, using an FBO works fine, but when switching to FBO we start seeing rendering bugs specifically on WebGL (not OpenGL ES 2.0), I have no idea why they are happening it seems to be culling random things even outside of the renderer, my best guess is a race condition, I saw similar issues with DirectX9 and RTX Remix.

miniwin/CMakeLists.txt
```diff
@@ -44,14 +44,20 @@ else()
...
+  find_library(OPENGL_ES2_LIBRARY NAMES GLESv2)
+  if(OPENGL_ES2_LIBRARY)
+    message(STATUS "Found OpenGL: enabling OpenGL ES 2.x renderer")
+    target_sources(miniwin PRIVATE src/d3drm/backends/opengles2/renderer.cpp)
+    target_compile_definitions(miniwin PRIVATE USE_OPENGLES2)
+    target_link_libraries(miniwin PRIVATE OpenGL::GL)
+  else()
+    message(STATUS "🧩 OpenGL ES 2.x support not enabled")
+  endif()
 endif()
```
AJenbo:
```cmake
find_library(OPENGL_ES2_LIBRARY NAMES GLESv2)
if(EMSCRIPTEN OR OPENGL_ES2_LIBRARY)
  message(STATUS "Found OpenGL: enabling OpenGL ES 2.x renderer")
  target_sources(miniwin PRIVATE src/d3drm/backends/opengles2/renderer.cpp)
  target_compile_definitions(miniwin PRIVATE USE_OPENGLES2)
  if(OPENGL_ES2_LIBRARY)
    target_link_libraries(miniwin PRIVATE OpenGL::GL)
  endif()
else()
  message(STATUS "🧩 OpenGL ES 2.x support not enabled")
endif()
```

PR #441, opened 2025-06-28
Add transition type to ini

foxtacles:
This allows configuring the transition type through an ini setting. The default setting is mosaic (same as the game's default without a setting).

Some transition types seem to be broken that should actually be working (like windows, or wipe down). I haven't looked into whether this is a bug in the decomp or in portable (possibly renderer-related). I think there's a chance nobody tried this before, although in the original game these transitions did work so I'll open an issue: https://github.com/isledecomp/isle-portable/issues/442


ISLE/isleapp.cpp
```diff
@@ -853,7 +852,13 @@ bool IsleApp::LoadConfig()
...
 
 	m_flipSurfaces = iniparser_getboolean(dict, "isle:Flip Surfaces", m_flipSurfaces);
+
+#ifdef __EMSCRIPTEN__
+	m_fullScreen = FALSE;
+#else
 	m_fullScreen = iniparser_getboolean(dict, "isle:Full Screen", m_fullScreen);
+#endif
+
 	m_wideViewAngle = iniparser_getboolean(dict, "isle:Wide View Angle", m_wideViewAngle);
```
jpeisach, line 862: FYI for 3ds this (Wide view angle) might be disabled to increase performance

Honestly I wonder if it might make sense to just abstract configs into their own functions in their own files: so we have a Emscripten_GetConfigDefaults which has this information and can prevent a hundred ifdefs in the code

ItzSwirlz (approved):
looks good

ISLE/isleapp.cpp
AJenbo: it's still going to require the same amount of fill rate just a few less triangles will be the src, so don't think it will amake any meaningful difference.

AJenbo:
> Some transition types seem to be broken that should actually be working (like windows, or wipe down). I haven't looked into whether this is a bug in the decomp or in portable (possibly renderer-related). I think there's a chance nobody tried this before, although in the original game these transitions did work so I'll open an issue: #442

It was likely me breaking them when translating things to support 32bit, I didn't know there was a way to actually test them.

PR #453, opened 2025-06-29
ci: combine ci and release workflow

madebr:


.github/workflows/ci.yml
```diff
@@ -148,3 +202,30 @@ jobs:
               LEGO1/omni/src/video/flic.cpp \
               $action_headers \
             --path LEGO1/omni LEGO1/lego/legoomni
+
+  release:
+    name: 'Release'
+    if: ${{ github.event_name == 'release' }}
```
foxtacles: Does this run automatically on master updates?

madebr: No. I skipped the trigger of `release.yml`. This has to be
```suggestion
    if: ${{ github.event_name == 'push' }}
```

.github/workflows/ci.yml
```diff
@@ -148,3 +202,30 @@ jobs:
               LEGO1/omni/src/video/flic.cpp \
               $action_headers \
             --path LEGO1/omni LEGO1/lego/legoomni
+
+  release:
+    name: 'Release'
+    if: ${{ github.event_name == 'push' }}
```
foxtacles: doesn't it need a check whether we are on the master branch as well? otherwise it would trigger on all branch pushes?

madebr:
The continuous releases will contain a lot more binaries, including mingw binaries and msvc 32-bit binaries (requiring an original d3drm.dll)

If necessary, we can remove these in the `release` job.

.github/workflows/ci.yml
madebr: Yes :)
```suggestion
    if: ${{ github.event_name == 'push' && github.ref_name == 'master' }}
```

foxtacles:
> The continuous releases will contain a lot more binaries, including mingw binaries and msvc 32-bit binaries (requiring an original d3drm.dll)
> 
> If necessary, we can remove these in the `release` job.

Looks like it's fewer https://github.com/isledecomp/isle-portable/releases/tag/continuous

Issue #399, opened 2025-06-22
Allow any number of joysticks to be used (make `joystickIndex` and `useJoystick` obsolete)


foxtacles:
Instead of relying on a `joystickIndex` and mapping that into the SDL available devices, I think we should follow the model I used for touch navigation: https://github.com/isledecomp/isle-portable/blob/master/LEGO1/lego/legoomni/src/input/legoinputmanager.cpp#L552

This queries all available input devices at runtime and applies any of their input states. It shoud allow all devices to be used as inputs, at any time, even when they are connected later (hotplugging)

Ramen2X:
Can look into this in a couple days

jpeisach:
While we're at this, can we also do gamepad support, I know it's a slightly larger task but it will be needed for other platforms

AJenbo:
Should be separate tasks, gamepad will benefit from this but adding a control scheme on top bloats this task a lot.

AJenbo:
Should be solved now that gamepad has been implemented.

foxtacles:
@AJenbo I don't think this is solved? The code still uses joystick index: https://github.com/isledecomp/isle-portable/blob/master/LEGO1/lego/legoomni/src/input/legoinputmanager.cpp#L165

And I don't think it supports more than one input device either.

foxtacles:
Although to be fair I guess having multiple joystick inputs doesn't make much sense here, and it seems to have a falback if the index is invalid, so I guess it's ok

PR #458, opened 2025-06-30
legoinputmanager: Don't set default value for numJoysticks

JPeisach:
Fixes joysticks not working

--------

For some reason it will find numJoysticks to be 0... even though that's not the case? I found disabling the default value (I guess the time it takes for `SDL_GetJoysticks` to return is too long) to fix this, though I expect compilers to not be happy with me...

If this doesn't work I'll move this to a static or class variable


foxtacles:
This doesn't make any sense

AJenbo:
Should do this instead: https://github.com/isledecomp/isle-portable/issues/399

The game should also listen for joystick connection events if it doesn't already.

PR #468, opened 2025-07-01
Add support for POT-only GPUs, fix VBO

AJenbo:
A few general fixes for OpenGL 1.1 that is needed for PSP support.
- Some systems can only use power of two textures
- VBO support broke in https://github.com/isledecomp/isle-portable/pull/446
- Correctly set wrapping and filtering
- Remove `glGetTexLevelParameterfv` requirement

### DIFF
--- a/miniwin/src/d3drm/backends/opengl1/actual.cpp
+++ b/miniwin/src/d3drm/backends/opengl1/actual.cpp
@@ -42,12 +42,26 @@ void GL11_DestroyTexture(GLuint texId)
 	glDeleteTextures(1, &texId);
 }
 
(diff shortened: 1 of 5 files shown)


foxtacles (approved):

AJenbo:
yes, so fare we are only targeting modern stuff, like OpenGL 1.1 - from 1997

AJenbo:
p.s. all textures are power of two, it's only the UI that isn't. For 3DS we resize it to screen size and place it on a larger texture, but for OpenGL I don't think we can quite make that assumption about the screen size. At least not if we don't check if we are running in full screen mode.

PR #478, opened 2025-07-02
Use sprites to render things on screen

AJenbo:
Avoid using the CPU to do scaling, blitting and clearing of the frame buffer. Instead it can be much faster to create a sprite in size of the change and pushing it to the blt/bltfast. This is especially true for low spec devices where we now only push a 320x240 texture from the CPU to the GPU when rendering video and leave the GPU to scale it to full screen.

@foxtacles can you please check that this still work correctly with DX5?

- Intro video
- Opening the registration book
- Pushing the button in the elevator
- Seeing the indicator for the elevator change


foxtacles:
https://github.com/user-attachments/assets/6eb74168-a073-44cb-8b64-8774996397bf

AJenbo:
Does that trigger everytime the book wiggles? Does everything else look correct?

foxtacles:
> Does that trigger everytime the book wiggles?

Yes. The same glitch happens in the elevator when you press a button

AJenbo:
Is it different with flip surface on/off

foxtacles:
I've tested with flip surfaces off.

Flip surfaces on is broken, doesn't launch:

```
ERROR: MxDirectDraw error: [0x887600e1] CreateSurface for front/back fullScreen buffer failed
ERROR: MxDirect3D::Create failed
ERROR: Failed to create LegoVideoManager
```

May have always been like this with DX5 though.

AJenbo:
@foxtacles please test the following on Dx5:
-  Intro video
- Opening the registration book

foxtacles:
Intro OK, rest:

https://github.com/user-attachments/assets/962dc7f7-5af0-4b89-bf9d-ea0471093cc2

LEGO1/omni/src/video/mxdisplaysurface.cpp
```diff
@@ -399,86 +399,78 @@ void MxDisplaySurface::VTable0x28(
...
-			data += stride;
-		}
+	RECT destRect;
+	if (m_videoParam.Flags().GetF1bit3()) {
+		destRect = {p_right, p_bottom, p_right + p_width * 2, p_bottom + p_height * 2};
 	}
+	else {
+		destRect = {p_right, p_bottom, p_right + p_width, p_bottom + p_height};
+	}
+	m_ddSurface2->Blt(&destRect, tempSurface, NULL, DDBLT_WAIT, NULL);
```
AJenbo: @foxtacles my only idea is try and use `m_ddSurface1`, could you give that a quick test?

foxtacles: https://github.com/user-attachments/assets/c3827568-422f-4793-8dac-c629354825a9

That produces a different set of issues: no intro screen, no book wiggling

AJenbo: Thanks, it's sort of what I expected, a bit of a long shot really.

foxtacles:
Intro not playing, book animation not playing

foxtacles (approved):
Works on dx5. Did not test other renderers

AJenbo:
I gave it a test in SDL3 GPU :)

PR #480, opened 2025-07-02
Make `draw cursor` feature work for modern platforms

Helloyunho:
This pr makes `draw cursor` option to actually work for modern platforms, potentially for console ports.

CleanShot 2025-07-03 at 01 28 29@2x


ISLE/stb_image.h
```diff
@@ -0,0 +1,7988 @@
+/* stb_image - v2.30 - public domain image loader - http://nothings.org/stb
```
foxtacles: This is a 3rd party dependency and should be located in the `3rdparty` folder.

I'd strongly encourage exploring alternatives to pulling in another dependency, i.e. using SDL, or using alternative methods. I'd rather not pull in a new dependency for a fringe feature such as this.

LEGO1/lego/legoomni/src/video/legovideomanager.cpp
```diff
@@ -272,15 +272,41 @@ void LegoVideoManager::MoveCursor(MxS32 p_cursorX, MxS32 p_cursorY)
...
+	if (640 < p_cursorX) {
+		m_cursorX = 640;
+	}
+
+	if (480 < p_cursorY) {
+		m_cursorY = 480;
+	}
+}
+
+void LegoVideoManager::SetCursorBitmap(const MxU8* p_cursorBitmap, MxS32 p_x, MxS32 p_y, MxS32 p_channels)
```
foxtacles: New functions should be implemented at the end of the unit

LEGO1/omni/src/video/mxdisplaysurface.cpp
```diff
@@ -1097,6 +1097,67 @@ LPDIRECTDRAWSURFACE MxDisplaySurface::CreateCursorSurface()
 	return NULL;
 }
 
+LPDIRECTDRAWSURFACE MxDisplaySurface::CreateCursorSurface(
```
foxtacles: New functions should be implemented at the end of the unit

foxtacles (changes requested):

ISLE/stb_image.h
Helloyunho: Got it, I'll look for alternatives.

AJenbo: SDL has build in support for loading BMP files, there is no need for extra code for this. All graphic in the game is also coming from BMP files so the game has custom code for it was well

ISLE/isleapp.cpp
```diff
@@ -656,6 +663,47 @@ MxResult IsleApp::SetupWindow()
 	m_cursorBusy = SDL_CreateSystemCursor(SDL_SYSTEM_CURSOR_WAIT);
 	m_cursorNo = SDL_CreateSystemCursor(SDL_SYSTEM_CURSOR_NOT_ALLOWED);
 	SDL_SetCursor(m_cursorCurrent);
+	if (g_isle->GetDrawCursor()) {
+		SDL_HideCursor();
+		SDL_IOStream* arrow_stream = SDL_IOFromMem(arrow_bmp, arrow_bmp_len);
```
madebr: You can use `SDL_IOFromConstMem`, which avoids a mem copy.

ISLE/res/arrow_bmp.h
```diff
@@ -0,0 +1,356 @@
+unsigned char arrow_bmp[] = {
```
madebr: Needed for `SDL_IOFromConstMem`, also for other headers.
```suggestion
const unsigned char arrow_bmp[] = {
```

AJenbo (changes requested):
Since these images gets embedded as C headers can we please convert the originals to png (easier to view, takes less space) and on top of that I noticed that they are black and white so lets encode them as bit flags instead of massive 32bit data blobs. This would reduce them from 4234b to 256b per cursor.

madebr:
> Since these images gets embedded as C headers can we please convert the originals to png (easier to view, takes less space) and on top of that I noticed that they are black and white so lets encode them as bit flags instead of massive 32bit data blobs.

SDL3 does not have a png decoder. You'd need SDL3_image, stb_image or lodepng for that.

AJenbo:
> SDL3 does not have a png decoder. You'd need SDL3_image, stb_image or lodepng for that.

From what I can tell the bmp files are never loaded they are just there for future convenience and viewing. The actual data being loaded is a byte array, the cursors trivial monocrome images with 1bit transparency so there isn't really a need for an image format.

AJenbo:
13kb -> 0.5kb
src png: [cursor.zip](https://github.com/user-attachments/files/21026323/cursor.zip)

79kb -> 6kb
headers: [generated-headers.zip](https://github.com/user-attachments/files/21026343/generated-headers.zip)

Python script for converting png to haders: [png2h.zip](https://github.com/user-attachments/files/21026351/png2h.zip)
(this PR is lacking instructions for how to convert .bmp to .h)

Function for loading the images:
```cpp
SDL_Surface* CreateCursorSurface(const uint8_t* data, const uint8_t* mask, int width, int height)
{
	SDL_Surface* surface = SDL_CreateSurface(width, height, SDL_PIXELFORMAT_ARGB8888);
	if (!surface) {
		return nullptr;
	}

	const SDL_PixelFormatDetails* format = SDL_GetPixelFormatDetails(surface->format);
	uint32_t whitePixel = SDL_MapRGBA(format, nullptr, 255, 255, 255, 255);
	uint32_t blackPixel = SDL_MapRGBA(format, nullptr, 0, 0, 0, 255);
	uint32_t transparentPixel = SDL_MapRGBA(format, nullptr, 255, 255, 255, 0);

	SDL_LockSurface(surface);
	uint32_t* pixels = static_cast<uint32_t*>(surface->pixels);

	for (int y = 0; y < height; ++y) {
		for (int x = 0; x < width; ++x) {
			int bitIndex = y * width + x;
			int byteIndex = bitIndex / 8;
			int bitOffset = 7 - (bitIndex % 8);

			bool isOpaque = (mask[byteIndex] >> bitOffset) & 1;
			bool isBlack = (data[byteIndex] >> bitOffset) & 1;

			uint32_t pixel;
			if (!isOpaque) {
				pixel = transparentPixel;
			}
			else {
				pixel = isBlack ? blackPixel : whitePixel;
			}

			pixels[y * surface->w + x] = pixel;
		}
	}

	SDL_UnlockSurface(surface);
	return surface;
}
```

madebr:
> > SDL3 does not have a png decoder. You'd need SDL3_image, stb_image or lodepng for that.
> 
> From what I can tell the bmp files are never loaded they are just there for future convenience and viewing. The actual data being loaded is a byte array, the cursors trivial monocrome images with 1bit transparency so there isn't really a need for an image format.

Right. I misread you comment a bit. You're right about the cursors being monochrome and a 1-bit bitmap being enough.
That's fine with me, with the condition that a build script is committed to convert from the png/bmp to header.

LEGO1/lego/legoomni/include/legovideomanager.h
```diff
@@ -6,6 +6,8 @@
 #include "legophonemelist.h"
 #include "mxvideomanager.h"
 
+#include <SDL3/SDL.h>
```
foxtacles:
```suggestion
#include <SDL3/SDL_surface.h>
```

ISLE/isleapp.cpp
```diff
@@ -1043,12 +1098,22 @@ void IsleApp::SetupCursor(Cursor p_cursor)
 		break;
 	}
 
-	if (m_cursorCurrent != NULL) {
-		SDL_SetCursor(m_cursorCurrent);
-		SDL_ShowCursor();
+	if (g_isle->GetDrawCursor()) {
+		if (m_cursorCurrentBitmap == NULL) {
+			VideoManager()->SetCursorBitmap(0);
```
foxtacles:
```suggestion
			VideoManager()->SetCursorBitmap(NULL);
```

LEGO1/omni/include/mxdisplaysurface.h
```diff
@@ -5,6 +5,8 @@
 #include "mxcore.h"
 #include "mxvideoparam.h"
 
+#include <SDL3/SDL.h>
```
foxtacles:
```suggestion
#include <SDL3/SDL_surface.h>
```

LEGO1/omni/src/video/mxdisplaysurface.cpp
```diff
@@ -10,6 +10,7 @@
 
 #include <SDL3/SDL_log.h>
 #include <assert.h>
+#include <string.h>
```
foxtacles: don't think this is needed?

Helloyunho:
@AJenbo
Should I bundle the python converter in `ISLE/res/`?
And should I keep `.cur` files?
Also I might be able to make CMake to generate header files

AJenbo:
I'm the wrong person to ask that.

LEGO1/lego/legoomni/src/video/legovideomanager.cpp
```diff
@@ -836,3 +835,35 @@ void LegoVideoManager::DrawTextToSurface32(
 		++p_text;
 	}
 }
+
+void LegoVideoManager::SetCursorBitmap(
+	MxS32 p_width,
+	MxS32 p_height,
+	const MxU8* p_cursorBitmap,
+	const MxU8* p_cursorMask
```
AJenbo: maybe just take `const CursorBitmap&` as input?

LEGO1/omni/src/video/mxdisplaysurface.cpp
```diff
@@ -1296,3 +1296,83 @@ LPDIRECTDRAWSURFACE MxDisplaySurface::FUN_100bc8b0(MxS32 p_width, MxS32 p_height
...
+	memset(&ddsd, 0, sizeof(ddsd));
+	ddsd.dwSize = sizeof(ddsd);
+
+	if (draw->GetDisplayMode(&ddsd) != DD_OK) {
+		return NULL;
+	}
+
+	MxU32 whitePixel = RGB8888_CREATE(0xff, 0xff, 0xff, 0xff);
+	MxU32 blackPixel = RGB8888_CREATE(0, 0, 0, 0xff);
+	MxU32 transparentPixel = RGB8888_CREATE(0, 0, 0, 0);
```
AJenbo: need RGB555 when in 16bit mode in case we want to support Voodoo again.

AJenbo:
Starting to look good i think.

LEGO1/omni/src/video/mxdisplaysurface.cpp
Helloyunho: Does `RGB555` support alpha channel?

AJenbo: No, it needs to use color key instad.

https://github.com/isledecomp/isle-portable/blob/605163b755273fd46fad57194bddd186b221b259/LEGO1/omni/src/video/mxdisplaysurface.cpp#L1085

AJenbo: You should also replace the old code with the new way of handeling a cursor.

AJenbo: No, technically you can set a range, but I both do not understand how that would work and SDL doesn't support it.
During decoding you set it to `RGB555_CREATE(0x1f, 0, 0x1f)` and then at the end you set purple as the color key for the surface.

Helloyunho: > You should also replace the old code with the new way of handeling a cursor.

You mean like delete old code? I'm keeping the old one for compatibility

AJenbo:
Looks good to me

CMakeLists.txt
```diff
@@ -529,6 +529,15 @@ if (ISLE_BUILD_APP)
...
     )
   endif()
+  if(Python3_FOUND)
+    execute_process(
+      COMMAND ${Python3_EXECUTABLE} tools/png2h.py ISLE/res/arrow.png ISLE/res/busy.png ISLE/res/no.png
+      WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+      RESULT_VARIABLE PNG2H_RESULT
+      OUTPUT_QUIET
+      ERROR_QUIET
+    )
```
madebr: This needs to be `add_custom_command`, the input and output files should be explicitly enumerated, and the generated headers should be part of `add_executable(isle)`.
Otherwise, the headers are always generated, causing `isle` to rebuild on every `cmake` invocation.

tools/png2h.py
```diff
@@ -0,0 +1,80 @@
...
+            f.write("\n")
+            f.write(
+                f"static const CursorBitmap {input_file_name}_cursor = {'{'} {width}, {height}, {input_file_name}_data, {input_file_name}_mask {'}'};\n"
+            )
+
+        print(f"Written {output_file} with cursor data.")
+
+
+if __name__ == "__main__":
+    main()
```
madebr:
```suggestion
    raise SystemExit(main())
```

tools/png2h.py
```diff
@@ -0,0 +1,80 @@
...
+    array_str = "\n    ".join(lines)
+    return f"static const unsigned char {name}[] = {{\n    {array_str}\n}};\n"
+
+
+def main():
+    if len(sys.argv) == 1:
+        print(f"Usage: {sys.argv[0]} [...input.png]")
+        sys.exit(1)
+
+    input_files = sys.argv[1:]
```
madebr: Use [`argparse`](https://docs.python.org/3/library/argparse.html)

```suggestion
    parser = argparse.ArgumentParser(allow_abbrev=False)
    parser.add_argument("inputs", nargs="+", help="PNG images")
    args = parser.parse_args();

    input_files = args.inputs
```

tools/png2h.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env python3
+import sys
```
madebr:
```suggestion
import argparse
import sys
```

madebr (changes requested):

CMakeLists.txt
madebr: Also not sure this should be part of CMake because `PIL` is not part of the Python standard library.

Helloyunho: I explicitly choose execute process since some machines(especially CI) don't have Pillow installed

madebr: It will silently fail on CI indeed. But it will succeed on my system, and every time I (or my IDE) runs CMake, isle will rebuild. That's not wanted.

Helloyunho: I don't think of better solution than removing png2h from cmake then...

madebr: To do this correctly, you might need to do a feature test. Something like:
```cmake
if(NOT DEFINED PYTHON3_PILLOW_AVAILABLE)
  execute_process(COMMAND ${Python3_EXECUTABLE} -c "import PIL"
    RESULT_VARIABLE result_import_pillow
  )
  if(result_import_pillow EQUAL 0)
    set(a 1)
  else()
    set(a 0)
  endif()
  set(PYTHON3_PILLOW_AVAILABLE ${a} CACHE BOOL "Is Python3 Pillow available?")
endif()
if(PYTHON3_PILLOW_AVAILABLE)
  add_custom_command(...)
endif()
```

(+ tests for availability of `Python3_EXECUTABLE`)

Helloyunho: Oh, didn't know that's possible. I'll fix it then

CMakeLists.txt
```diff
@@ -529,6 +532,36 @@ if (ISLE_BUILD_APP)
...
+    if(PYTHON_PIL_AVAILABLE)
+      add_custom_command(
+        OUTPUT
+          ${CMAKE_SOURCE_DIR}/ISLE/res/arrow_bmp.h
+          ${CMAKE_SOURCE_DIR}/ISLE/res/busy_bmp.h
+          ${CMAKE_SOURCE_DIR}/ISLE/res/no_bmp.h
+        COMMAND ${Python3_EXECUTABLE} tools/png2h.py ISLE/res/arrow.png ISLE/res/busy.png ISLE/res/no.png
+        WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+        DEPENDS
+          ${CMAKE_SOURCE_DIR}/ISLE/res/arrow.png
```
madebr, line 562: The python script is also a dependency.
```suggestion
          ${CMAKE_SOURCE_DIR}/tools/png2h.py
          ${CMAKE_SOURCE_DIR}/ISLE/res/arrow.png
```

tools/png2h.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python3
```
madebr: I'd rename this python script, since it targets png cursors only.
Something like `png2cursorh.py` (I'm bad at naming things)

Helloyunho: How about `cur2header.py`? or `cursor2header.py`?

madebr: or `pngcur2h.py`?

Helloyunho: `curpng2h.py`? since it's `cursorname.png`

madebr: That's fine with me :)

CMakeLists.txt
```diff
@@ -529,6 +532,38 @@ if (ISLE_BUILD_APP)
...
+        RESULT_VARIABLE PIL_RESULT
+        OUTPUT_VARIABLE PIL_OUTPUT
+        ERROR_QUIET
+        OUTPUT_STRIP_TRAILING_WHITESPACE
+      )
+      if(PIL_RESULT EQUAL 0 AND PIL_OUTPUT STREQUAL "pil")
+        set(PYTHON_PIL_AVAILABLE TRUE)
+      else()
+        message(STATUS "Python PIL not found, using pre-generated headers.")
+        set(PYTHON_PIL_AVAILABLE FALSE)
```
madebr: `PYTHON_PIL_AVAILABLE` should be a cache variable. (See my suggestion above)

Helloyunho: May I ask you why? Does it still refresh when user installs Pillow later the project was configured?

madebr: We want reconfigures to be fast.
CMake (and all programs in general) cache things to make things fast.
The default behavior of `find_package`, `find_library`, ... is also to cache its result.

You can always force (re)detection by configuring with `--fresh`, `-DPYTHON_PIL_AVAILABLE=1`, or `-UPYTHON_PIL_AVAILABLE`

Helloyunho: Okay, then I'll make it cache the result

CMakeLists.txt
```diff
@@ -529,6 +532,38 @@ if (ISLE_BUILD_APP)
...
+        OUTPUT_VARIABLE PIL_OUTPUT
+        ERROR_QUIET
+        OUTPUT_STRIP_TRAILING_WHITESPACE
+      )
+      if(PIL_RESULT EQUAL 0 AND PIL_OUTPUT STREQUAL "pil")
+        set(PYTHON_PIL_AVAILABLE TRUE CACHE BOOL "Is Python3 Pillow available?")
+      else()
+        message(STATUS "Python PIL not found, using pre-generated headers.")
+        set(PYTHON_PIL_AVAILABLE FALSE CACHE BOOL "Is Python3 Pillow available?")
+      endif()
```
madebr: - Do a single `set(cache)` to avoid code duplication.
- `TRUE/FALSE` is valid too, but `1/0` is the CMake default for cached check functions.
```suggestion
        set(found 1)
      else()
        message(STATUS "Python PIL not found, using pre-generated headers.")
        set(found 0)
      endif()
        set(PYTHON_PIL_AVAILABLE ${found} CACHE BOOL "Is Python3 Pillow available?")
```

tools/curpng2h.py
```diff
@@ -0,0 +1,79 @@
...
+    for input_file in input_files:
+        data, mask, width, height = encode_cursor(input_file)
+
+        input_file_name = input_file.stem
+        output_file = input_file.with_name(f"{input_file_name}_bmp.h")
+
+        with output_file.open("w") as f:
+            f.write(f"// Generated from {input_file}\n")
+            f.write(f"// Dimensions: {width}x{height}\n\n")
+            f.write(f"#pragma once\n")
```
madebr: - Include guard is usually first in a header.
- Also open the file with `newline="\n"` to avoid `\r\n` on Windows
```suggestion
        with output_file.open("w", newline="\n") as f:
            f.write(f"#pragma once\n")
            f.write(f"// Generated from {input_file}\n")
            f.write(f"// Dimensions: {width}x{height}\n\n")
```

tools/curpng2h.py
```diff
@@ -0,0 +1,79 @@
...
+            f.write(f'#include "cursor.h"\n\n')
```
madebr, line 62: Also print a auto-generated warning.
```
f.write("// This file is auto-generated, do not edit it.\n")
```

tools/curpng2h.py
```diff
@@ -0,0 +1,79 @@
...
+    lines = []
+    line = ""
+    for i, b in enumerate(data):
+        line += f"0x{b:02X}, "
+        if (i + 1) % 12 == 0:
+            lines.append(line)
+            line = ""
+    if line:
+        lines.append(line)
+    array_str = "\n    ".join(lines)
```
madebr: This can be simplified.
See `write_uint8_array` of `miniwin/src/d3drm/backends/sdl3gpu/shaders/gencshadersource.py`

madebr (approved):

PR #489, opened 2025-07-02
Render surfaces in a single blit

AJenbo:
This function will split surfaces along the horizontal line where it's intersected by animated sprites and render things in regions instead of a single blit. This could have been a nice optimization on software rendering if it only rendered to the previous position of the sprite, but it always renders all regions leaving the whole process mute and simply adds overhead, and especially on low resolution devices this can lead to rounding issues where the background or other sprites wobble when things move around on top.

@foxtacles things to check on DX5:
- Does the intro render at scale
- Does things look right as you move around the portraits


foxtacles:
https://github.com/user-attachments/assets/390a4bff-5d0a-4b17-964c-97be4987961c

AJenbo:
ok ... so half of it worked.

It is nice to see that it also renders the sky in the background here :)

LEGO1/omni/src/video/mxvideopresenter.cpp
```diff
@@ -240,97 +240,36 @@ void MxVideoPresenter::PutFrame()
...
+		dest.right = dest.left + GetWidth();
+		dest.bottom = dest.top + GetHeight();
+
+		switch (PrepareRects(src, dest)) {
+		case 0:
+			ddSurface->Blt(&dest, m_unk0x58, &src, DDBLT_KEYSRC, NULL);
+			break;
+		case 1:
+			ddSurface->BltFast(dest.left, dest.top, m_unk0x58, &src, DDBLTFAST_SRCCOLORKEY | DDBLTFAST_WAIT);
 		}
```
AJenbo: @foxtacles could you test this:

```cpp
		ddSurface->Blt(&dest, m_unk0x58, &src, DDBLT_NONE, NULL);
```
and
```cpp
		ddSurface->Blt(NULL, m_unk0x58, NULL, DDBLT_NONE, NULL);
```

foxtacles: which line should I replace with these

AJenbo: The ones i marked (255-260)

foxtacles: First line:
image
image
image

Second line:
image

AJenbo: I prefer the first one :D

AJenbo: @foxtacles I think the issue should be solved now, could you please check the latest version of the PR?

foxtacles: image

AJenbo: Could you give it another go ... I think it should work this time.

foxtacles: Same:

image

commit [f36167b](https://github.com/isledecomp/isle-portable/commit/f36167bf408c386122b305029931e480b108516e), fresh build

AJenbo: @foxtacles alright this time i got it, I think.

foxtacles: will test tomorrow morning

foxtacles: Looks good

foxtacles (approved):

PR #490, opened 2025-07-03
Fix saving of default textures

### LINKED ISSUE #483: Broken textures/colours on the racecar
_In the web version (isle.pizza):_

The textures on the racecar seem to mess up when building the default car:
Image
Image

Steps to reproduce:
1. Choose any character
2. Go to race track
3. Build the racecar without changing any colours or decals

Expected result:
The decals and colours of the built car outside match those form the defaults chosen in the build area.

Actual result:
Some textures and colours are plain white, while others work just fine.

Extra info:
The second screenshot shows me going back to rebuild the car, and the decals are still broken.

foxtacles:
Close #483

@AJenbo do we need to remove the `DDLOCK_WRITEONLY` flag?

### DIFF
--- a/LEGO1/lego/legoomni/src/common/legoutils.cpp
+++ b/LEGO1/lego/legoomni/src/common/legoutils.cpp
@@ -708,7 +708,7 @@ void WriteDefaultTexture(LegoStorage* p_storage, const char* p_name)
 		memset(&desc, 0, sizeof(desc));
 		desc.dwSize = sizeof(desc);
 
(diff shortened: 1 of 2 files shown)


AJenbo:
Yes I think so, but doing so here shouldn't be an issue since it looks to be a plain surface and not a framebuffer.

foxtacles:
Tested on software render, works fine

AJenbo:
Fwiw the software renderer doesn't really care if the surface is read or write since it's main memory.

PR #497, opened 2025-07-03
Added QT DLLs (i think?)

BagelSketches:
implemented windeployqt

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,10 +35,10 @@ jobs:
       matrix:
         include:
           - { name: 'Linux',          os: 'ubuntu-latest',  dx5: false, config: true,  linux: true, werror: true,   clang-tidy: true }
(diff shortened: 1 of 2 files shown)


BagelSketches:
this previously ddidnt work in the earlier one because i put bin instead of build, this should work now.

.github/workflows/ci.yml
```diff
@@ -103,6 +107,8 @@ jobs:
       - name: Build (CMake)
         run: cmake --build build --verbose
 
+
+
```
AJenbo:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -138,6 +144,11 @@ jobs:
           cd build
           mkdir dist
           mv *.3dsx dist/
+      - name: Pack stuff(windows)
```
AJenbo:
```suggestion

      - name: Pack stuff(windows)
```

PR #503, opened 2025-07-03
3DS: Account for sprite scaling when resizing UI

AJenbo:
This fixes scaling of cached UI elements that will be rendered at a non 1:1 scale in relation to the virtual canvas. This mostly serves as an optimization for 3DS.

Technically other platforms can benefit from this to, but the other renderers are generic and doesn't take the specific screen size in to account.

This also generates the fake Mosaic at 64x48 instead of 640x480 (the blocks are 10x10 pixels), allowing the 3DS to process it at full FPS


AJenbo:
No not really. It's mostly to avoid scaling first scaling things up to 640x480 before scaling them down to the 3ds resolution. There can be some rounding differences but they should fall in favor of quality.

PR #511, opened 2025-07-03
Add changing the transition type to the config tool

### LINKED ISSUE #510: Config removes "Transition Type" from ini
If you use the config application the "Transition Type" option will be removed from your ini

StevenSYS:


foxtacles:
> Closes https://github.com/isledecomp/isle-portable/issues/510

I'd say that the fix should be that the config doesn't remove extra .ini keys that it isn't aware of, in general

StevenSYS:
I would try to make it not remove extra keys, but I don't think I understand iniparser enough to do that

AJenbo:
At least t his PR does add the option to change it in config.

AJenbo:
Uh is the transition option offscreen somewhere?

image

StevenSYS:
This is what it looks like for me

Skärmbild från 2025-07-04 09-17-56

AJenbo:
It's because I haven't realized config was renamed to isle-config and I'm still using an old build because of that.

PR #513, opened 2025-07-03
Allow Web port to be playable in Firefox Private

### LINKED ISSUE #444: Web port not compatible with Firefox Private Mode
Mozilla has not implemented OPFS in private mode. A workaround is possible but non-trivial

foxtacles:
Close #444 

for isle.pizza, this will be shown and config be read-only /disabled in Firefox Privatge:

image


AJenbo:
Still no full screen high resolution hacks :smiling_face_with_tear:

PR #335, opened 2025-06-18
Build config during CI for MSVC x64

AJenbo:

PR #520, opened 2025-07-04
Added building config.exe with qt dlls on msvc 64-bit builds

BagelSketches:
after a while trying to figure out how cpack works, i got the qt dlls working in the msvc 64 build of the config. i made some slight adjustments to the original pr from @AJenbo with the same goal as mine, see :  #335 
the only thing i really did was get the install(directory) command in cmakelists to work(get cpack to package the dlls with the binary)


BagelSketches:
did most of this without access to a code editor so i used githubs edit file function lol

CMakeLists.txt
```diff
@@ -680,6 +680,24 @@ install(TARGETS isle ${install_extra_targets}
...
+              "--dir"
+              "QTLibs"
+              "--no-compiler-runtime"
+              "--no-opengl-sw"
+              "--no-system-d3d-compiler"
+              "--no-translations"
+              "--no-quick-import"
+      COMMENT "Running windeployqt with minimal dependencies"
+    )
+    install(DIRECTORY "Build/QTLibs/" DESTINATION "${CMAKE_INSTALL_BINDIR}")
```
madebr: Running `windeployqt` on every build is not correct.
It should be only run at install time.
Something as following (untested):
```suggestion
    install(CODE "message(STATUS \"Running windeployqt with minimal dependencies\")
      execute_process(COMMAND \"${WINDEPLOYQT_EXECUTABLE}\"
        \"$<TARGET_FILE:isle-config>\"
        --dir QTLibs
        --no-compiler-runtime
        --no-opengl-sw
        --no-system-d3d-compiler
        --no-translations
        --no-quick-import
      )"
    )
    install(DIRECTORY "Build/QTLibs/" DESTINATION "${CMAKE_INSTALL_BINDIR}")
```

CMakeLists.txt
```diff
@@ -680,6 +680,24 @@ install(TARGETS isle ${install_extra_targets}
   LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}"
 )
 if (ISLE_BUILD_CONFIG)
+  find_program(WINDEPLOYQT_EXECUTABLE windeployqt)
+  if(WINDEPLOYQT_EXECUTABLE)
```
madebr: This probably needs a `if(WIN32)`

CMakeLists.txt
```diff
@@ -680,6 +680,24 @@ install(TARGETS isle ${install_extra_targets}
...
+              "--no-compiler-runtime"
+              "--no-opengl-sw"
+              "--no-system-d3d-compiler"
+              "--no-translations"
+              "--no-quick-import"
+      COMMENT "Running windeployqt with minimal dependencies"
+    )
+    install(DIRECTORY "Build/QTLibs/" DESTINATION "${CMAKE_INSTALL_BINDIR}")
+  else()
+    message(WARNING "windeployqt not found.")
```
madebr: A warning is too harsh imho, and we want to make it possible to install `isle-config` without qt binaries (e.g. msys2)
```suggestion
    message(STATUS "windeployqt not found: Qt binaries will not be installed")
```

madebr (changes requested):

CMakeLists.txt
madebr: Can you do this?
This message should not be printed on non-Windows platforms.

CMakeLists.txt
```diff
@@ -680,6 +680,23 @@ install(TARGETS isle ${install_extra_targets}
...
+        --no-opengl-sw
+        --no-system-d3d-compiler
+        --no-translations
+        --no-quick-import
+      )"
+    )
+    install(DIRECTORY "Build/QTLibs/" DESTINATION "${CMAKE_INSTALL_BINDIR}")
+  else()
+    message(STATUS "windeployqt not found: Qt binaries will not be installed")
+  endif()
```
AJenbo: This should satisfy the last review comment:
```suggestion
  if(WIN32)
    find_program(WINDEPLOYQT_EXECUTABLE windeployqt)
    if(WINDEPLOYQT_EXECUTABLE)
      install(CODE "message(STATUS \"Running windeployqt with minimal dependencies\")
        execute_process(COMMAND \"${WINDEPLOYQT_EXECUTABLE}\"
          \"$<TARGET_FILE:isle-config>\"
          --dir QTLibs
          --no-compiler-runtime
          --no-opengl-sw
          --no-system-d3d-compiler
          --no-translations
          --no-quick-import
        )"
      )
      install(DIRECTORY "Build/QTLibs/" DESTINATION "${CMAKE_INSTALL_BINDIR}")
    else()
      message(STATUS "windeployqt not found: Qt binaries will not be installed")
    endif()
  endif()
```

madebr (approved):
We can probably reduce the file size of the msvc x64 archive, but looks good to me.

(I have not tested the executable)

AJenbo:
@BagelSketches thanks for figuring this out

PR #524, opened 2025-07-04
Add DXBC for Xbox ONE support

AJenbo:


Helloyunho:
As discussed in Matrix, the space parameters in shader doesn't work on shader model 5.0, which is the version that dx11 supports

madebr (changes requested):
I'm getting Vulkan validation errors when trying this PR.
Afterwards, the address sanitizer detects a double free.

miniwin/src/d3drm/backends/sdl3gpu/shaders/src/Common.hlsl
```diff
@@ -18,3 +18,15 @@ struct SceneLight {
...
+    #define REGISTER_B0_SPACE1 register(b0)
+    #define REGISTER_B1_SPACE3 register(b0)
+    #define REGISTER_T0_SPACE2 register(t0)
+    #define REGISTER_S0_SPACE2 register(s0)
+#else
+    #define REGISTER_B0_SPACE1 register(b0, space1)
+    #define REGISTER_B1_SPACE3 register(b0, space3)
+    #define REGISTER_T0_SPACE2 register(t0, space2)
+    #define REGISTER_S0_SPACE2 register(s0, space2)
+#endif
```
madebr, line 32: Wouldn't the following be more usable in the future?
```suggestion
#ifdef NO_REGISTER_SPACE
    #define REGISTER_SPACE(REG, SPACE) register(REG)
#else
    #define REGISTER_SPACE(REG, SPACE) register(REG, SPACE)
#endif
```
and use `REGISTER_SPACE(b0, space1)` in the shaders?

AJenbo:
Need SM 5.0 is SM 5.1

PR #525, opened 2025-07-05
Fix uninitialized presenter

AJenbo:
This was inherited from the original and probably leads to random crashes, so time to fix.


LEGO1/lego/legoomni/src/worlds/registrationbook.cpp
```diff
@@ -671,14 +671,14 @@ MxLong RegistrationBook::HandlePathStruct(LegoPathStructNotificationParam& p_par
 MxBool RegistrationBook::CreateSurface()
 {
 	MxCompositePresenterList* presenters = m_checkmark[0]->GetList();
-	MxStillPresenter *presenter, *uninitialized;
+	MxStillPresenter *presenter;
```
AJenbo, line 674:
```suggestion
	MxStillPresenter* presenter;
```

Helloyunho (approved):
LGTM

foxtacles (approved):

PR #526, opened 2025-07-05
Add Xbox One/Series X|S port

Helloyunho:
## Description

This PR adds Xbox One/Series X|S port. Game is playable with minor bugs.

## Bugs

- You cannot exit the game normally, probably due to SDL_assert failure (the game does save though).
- Controller does not work when it's connected after the game has launched.

## Not tested

- Didn't go through the ending yet.
- Didn't check what happens if you connect more than 1 controllers.


ISLE/dummy_controller_map.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
```
foxtacles: Let's please use consistent include guards (the one original isle uses)

LEGO1/lego/legoomni/src/input/legoinputmanager.cpp
```diff
@@ -191,8 +189,13 @@ MxResult LegoInputManager::GetJoystickState(MxU32* p_joystickX, MxU32* p_joystic
 			return FAILURE;
 		}
 
+#ifdef WINDOWS_STORE
```
foxtacles: we should not be using platform specific code in LEGO1.  If SDL doesn't abstract this properly for us (is that expected?) we need to add our own interface abstraction

LEGO1/lego/legoomni/src/race/carrace.cpp
```diff
@@ -386,7 +386,7 @@ MxLong CarRace::HandleType0Notification(MxNotificationParam&)
 void CarRace::FUN_10017820(MxS32 p_param1, MxS16 p_param2)
 {
 	MxS32 local4;
-	MxStillPresenter* presenter;
+	MxStillPresenter* presenter = NULL;
```
foxtacles, line 389: What's this change for (and the others like it?)

ISLE/dummy_controller_map.h
```diff
@@ -0,0 +1,8 @@
...
+
+#define ISLE_BUTTON_SPACE 0
+#define ISLE_BUTTON_ESCAPE 0
```
foxtacles: Rather than defining "dummy" values (which I assume won't fulfill any functionality), let's ifdef the entire functionality that depends on it

foxtacles (changes requested):

LEGO1/lego/legoomni/src/race/carrace.cpp
Helloyunho: without that, MSVC cries for potentially uninitialized variable

ISLE/dummy_controller_map.h
Helloyunho: it is possible, but I thought later we could fill this with default values(maybe same as xbox mapping for xinput..?)

ISLE/dummy_controller_map.h
Helloyunho: the ifndef and define method? sure

LEGO1/lego/legoomni/src/input/legoinputmanager.cpp
Helloyunho: I don't think of better way than doing this hmmm....

miniwin/CMakeLists.txt
```diff
@@ -190,4 +194,4 @@ if(ISLE_COMPILE_SHADERS)
       "--shader-jsons" ${shader_jsons}
     DEPENDS "${py_gencshadersource}" ${shader_headers} ${shader_jsons})
 endif()
-target_sources(miniwin PRIVATE "${index}" "${index_cpp}")
+target_sources(miniwin PRIVATE "${index}" "${index_cpp}")
```
AJenbo: Make sure files end on a new line

ISLE/isledebug.cpp
```diff
@@ -183,7 +183,7 @@ void IsleDebug_Init()
 		}
 		g_videoPalette =
 			SDL_CreateTexture(g_debugRenderer, SDL_PIXELFORMAT_RGBX32, SDL_TEXTUREACCESS_STREAMING, 16, 16);
-#if SDL_VERSION_ATLEAST(3, 3, 0)
+#if SDL_VERSION_ATLEAST(3, 3, 0) && !defined(WINDOWS_STORE)
```
AJenbo: This seems really strange

ISLE/dummy_controller_map.h
foxtacles: Either we do it right away or we don't, but I think a stop gap of having dummy values is more confusing than anything

LEGO1/lego/legoomni/src/input/legoinputmanager.cpp
foxtacles: We should add a new module on the top level, similar to `miniwin`, that provides platform-specific abstractions as we need them.

i.e. instead of using `SDL_GetJoystickAxis` we would use something like `Isle_GetJoystickAxis` (which could have the same interface, in the beginning). This function can then be defined on a per-platform basis in this module (and probably have a default fallback to the regular SDL call).

There's already something very simple I added for Emscripten here: https://github.com/isledecomp/isle-portable/blob/master/util/compat.h#L22 which we should refactor into that module as well.

If you'd like one of the established contributors can propose/add a setup for this.

CC @madebr @AJenbo

LEGO1/lego/legoomni/src/input/legoinputmanager.cpp
```diff
@@ -153,9 +153,7 @@ MxResult LegoInputManager::GetJoystick()
 	}
 
 	MxS32 numJoysticks = 0;
-	if (m_joyids == NULL) {
-		m_joyids = SDL_GetJoysticks(&numJoysticks);
-	}
+	m_joyids = SDL_GetJoysticks(&numJoysticks);
```
AJenbo: If you need to update the id each time then you have to free the old one first.
```suggestion
	if (m_joyids) {
		SDL_free(m_joyids);
	}
	m_joyids = SDL_GetJoysticks(&numJoysticks);
```

ISLE/dummy_controller_map.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+#define ISLE_BUTTON_SPACE 0
```
AJenbo: maybe better to call it interrupti instead of space?

ISLE/xbox_one_series/xbox_controller_map.h
```diff
@@ -0,0 +1,30 @@
...
+#define SDL_XBOX_BUTTON_START 7
+#define SDL_XBOX_BUTTON_LEFT_THUMB 8
+#define SDL_XBOX_BUTTON_RIGHT_THUMB 9
+#define SDL_XBOX_BUTTON_DPAD_UP 10
+#define SDL_XBOX_BUTTON_DPAD_DOWN 11
+#define SDL_XBOX_BUTTON_DPAD_LEFT 12
+#define SDL_XBOX_BUTTON_DPAD_RIGHT 13
+
+#define SDL_XBOX_AXIS_LEFT_Y 0
+#define SDL_XBOX_AXIS_LEFT_X 1
```
AJenbo: What's preventing you from using the normal `SDL_GAMEPAD_AXIS_LEFTX`?

ISLE/isleapp.cpp
```diff
@@ -896,6 +992,7 @@ bool IsleApp::LoadConfig()
 	m_useJoystick = iniparser_getboolean(dict, "isle:UseJoystick", m_useJoystick);
 	m_joystickIndex = iniparser_getint(dict, "isle:JoystickIndex", m_joystickIndex);
 	m_drawCursor = iniparser_getboolean(dict, "isle:Draw Cursor", m_drawCursor);
+	m_mouseSensitivity = iniparser_getdouble(dict, "isle:Mouse Sensitivity", m_mouseSensitivity);
```
AJenbo: Since this doesn't affect the actual mouse sensitivitet maybe "cursor", "virtual mouse", or "controller" would be better.

LEGO1/lego/legoomni/src/input/legoinputmanager.cpp
foxtacles: Another thought: maybe we should be using the Game Controller API of SDL instead? https://wiki.libsdl.org/SDL3/CategoryGamepad

It seems this API would save us doing a bunch of our own joystick/gamepad abstraction.

AJenbo: For DevilutionX we only use the joystick API under SDL 1.2 and only to do what the Gampead API does in SDL 2/3.

ISLE/dummy_controller_map.h
AJenbo: Lets not use xinput, use SDL_Gamepad instead, it abstracts xinput as well as many other gamepad protocols.

LEGO1/lego/legoomni/src/input/legoinputmanager.cpp
foxtacles: It would probably be a good idea to integrate gamepad API / controls as a separate PR, and then rebase this port on top of it. That way we have two clean efforts.

Helloyunho: I'll make a new PR with SDL's Gamepad API

ISLE/isledebug.cpp
Helloyunho: SDL3 UWP port doesn't support `SDL_SCALEMODE_PIXELART`

ISLE/isleapp.cpp
Helloyunho: `cursor` sounds good to me

ISLE/xbox_one_series/xbox_controller_map.h
madebr: Also please don't pollute the `SDL_` namespace.

CMakeLists.txt
```diff
@@ -532,6 +545,11 @@ if (ISLE_BUILD_APP)
       ISLE/3ds/config.cpp
     )
   endif()
+  if(WINDOWS_STORE)
+    target_sources(isle PRIVATE
+      ISLE/xbox_one_series/config.cpp
+    )
```
madebr, line 551: UWP != xbox series

Does this also work as a Windows Store app on a desktop computer?
If so, then the subfolder can be renamed to uwp or windowsstore.

ISLE/isledebug.cpp
madebr: This should be fixed in the UWP port imho.

The UWP port should fall back to `SDL_SCALEMODE_NEAREST`, [like the software renderer does](https://github.com/libsdl-org/SDL/blob/7c11a8cb9f66a2cac63f9a24ab2f49f6d4bf12a0/src/render/software/SDL_render_sw.c#L427-L429)

ISLE/isleapp.cpp
```diff
@@ -675,7 +766,7 @@ MxResult IsleApp::SetupWindow()
 	SDL_SetNumberProperty(props, SDL_PROP_WINDOW_CREATE_HEIGHT_NUMBER, g_targetHeight);
 	SDL_SetBooleanProperty(props, SDL_PROP_WINDOW_CREATE_FULLSCREEN_BOOLEAN, m_fullScreen);
 	SDL_SetStringProperty(props, SDL_PROP_WINDOW_CREATE_TITLE_STRING, WINDOW_TITLE);
-#if defined(MINIWIN) && !defined(__3DS__)
+#if defined(MINIWIN) && !defined(__3DS__) && !defined(WINDOWS_STORE)
```
jpeisach, line 808: we should maybe make an ifdef for "is using openGL", but since that can be changed at runtime idk what the best way of checking for that is

AJenbo: it would probably be good to add a flag for if OpenGL support should be build, but I don't know how that should be handled from Isle's side. @foxtacles

foxtacles: Assuming there is no way to get rid of it without rewriting a whole bunch of other stuff, I'd suggest we add a new interface to miniwin, something akin to:

`Miniwin_CreateWindowWithProperties`

which can replace `SDL_CreateWindowWithProperties`. Then miniwin can handle all the renderer-related, platfom-specific setup for the window before its creation. I realize that would be the first API of its kind, but I think that would be cleaner than leaking more information about graphics backends into isle. miniwin already has all that information (`USE_OPENGL1`, `USE_OPENGLES2`) and by definition always will have all the information for future backends as well.

If we want to avoid the notion of miniwin creating a window itself we could also have an API like `Miniwin_AugmentWindowProperties` which just adds whatever necessary to `props`. That could be as simple as a macro.

ISLE/isledebug.cpp
Helloyunho: Fixed by updating the SDL port

CMakeLists.txt
Helloyunho: It does run... with software renderer and it doesn't register mouse click

Helloyunho: Also, that config is meant to be applied only on Xbox...

ISLE/isleapp.cpp
```diff
@@ -457,6 +491,63 @@ SDL_AppResult SDL_AppEvent(void* appstate, SDL_Event* event)
...
+		}
+		else if (event->gaxis.axis == ISLE_MOUSE_JOYSTICK_Y) {
+#ifdef WINDOWS_STORE
+			g_lastJoystickMouseY =
+				-((float) event->gaxis.value) / SDL_JOYSTICK_AXIS_MAX * g_isle->GetMouseSensitivity();
+#else
+			g_lastJoystickMouseY = ((float) event->gaxis.value) / SDL_JOYSTICK_AXIS_MAX * g_isle->GetMouseSensitivity();
+#endif
+		}
+		else if (event->gaxis.axis == ISLE_MOUSE_CLICK_AXIS) {
```
madebr: Imho, using `ISLE_MOUSE_CLICK_AXIS` does not make the code clearer.
Let's just use `SDL_GAMEPAD_AXIS_LEFTX` instead.

ISLE/isleapp.cpp
```diff
@@ -457,6 +491,63 @@ SDL_AppResult SDL_AppEvent(void* appstate, SDL_Event* event)
...
+		if (event->gaxis.axis == ISLE_MOUSE_JOYSTICK_X) {
+			g_lastJoystickMouseX = ((float) event->gaxis.value) / SDL_JOYSTICK_AXIS_MAX * g_isle->GetMouseSensitivity();
+		}
+		else if (event->gaxis.axis == ISLE_MOUSE_JOYSTICK_Y) {
+#ifdef WINDOWS_STORE
+			g_lastJoystickMouseY =
+				-((float) event->gaxis.value) / SDL_JOYSTICK_AXIS_MAX * g_isle->GetMouseSensitivity();
+#else
+			g_lastJoystickMouseY = ((float) event->gaxis.value) / SDL_JOYSTICK_AXIS_MAX * g_isle->GetMouseSensitivity();
+#endif
```
madebr: This should be addressed by the sdl3-uwp fork: there's no reason that fork returns negative values whereas SDL3 would return positive values.

ISLE/isleapp.cpp
```diff
@@ -457,6 +491,63 @@ SDL_AppResult SDL_AppEvent(void* appstate, SDL_Event* event)
...
+			if (event->gbutton.button == ISLE_BUTTON_ESCAPE) {
+				if (InputManager()) {
+					InputManager()->QueueEvent(c_notificationKeyPress, SDLK_ESCAPE, 0, 0, SDLK_ESCAPE);
+				}
+			}
+		}
+		break;
+	}
+	case SDL_EVENT_JOYSTICK_AXIS_MOTION: {
+		if (event->gaxis.axis == ISLE_MOUSE_JOYSTICK_X) {
```
madebr:
```suggestion
		if (event->gaxis.axis == SDL_GAMEPAD_AXIS_LEFTX) {
```

ISLE/isleapp.cpp
```diff
@@ -457,6 +491,63 @@ SDL_AppResult SDL_AppEvent(void* appstate, SDL_Event* event)
...
+			g_lastJoystickMouseX = ((float) event->gaxis.value) / SDL_JOYSTICK_AXIS_MAX * g_isle->GetMouseSensitivity();
+		}
+		else if (event->gaxis.axis == ISLE_MOUSE_JOYSTICK_Y) {
```
madebr: You've defined both `ISLE_MOUSE_JOYSTICK_X` and `ISLE_MOUSE_JOYSTICK_Y` as 0, which is `SDL_GAMEPAD_AXIS_LEFTX`.

That's unlikely to be correct?

```suggestion
		else if (event->gaxis.axis == SDL_GAMEPAD_AXIS_LEFTX) {
```

madebr (changes requested):

ISLE/isleapp.cpp
AJenbo: I'm not opposed to it

CMakeLists.txt
AJenbo: Fwiw DevilutionX doesn't distinguish between UWP and Xbox. But it does call it UWP and not WINDOWS_STORE.

AJenbo:
Looks like resolving the conflict will address most of the issues pointed out :)

CMakeLists.txt
madebr: I like DevilutionX's approach. Let's use UWP everywhere, including the c++ code.

ISLE/xbox_one_series/xbox_controller_map.h
```diff
@@ -0,0 +1,30 @@
...
+#define SDL_XBOX_AXIS_RIGHT_X 3
+#define SDL_XBOX_AXIS_LEFT_TRIGGER 4
+#define SDL_XBOX_AXIS_RIGHT_TRIGGER 5
+
+#define ISLE_BUTTON_SPACE SDL_XBOX_BUTTON_A
+#define ISLE_BUTTON_ESCAPE SDL_XBOX_BUTTON_START
+
+#define ISLE_MOUSE_JOYSTICK_X SDL_XBOX_AXIS_RIGHT_X
+#define ISLE_MOUSE_JOYSTICK_Y SDL_XBOX_AXIS_RIGHT_Y
+#define ISLE_MOUSE_CLICK_AXIS SDL_XBOX_AXIS_RIGHT_TRIGGER
```
AJenbo: This can all be removed now right?

Helloyunho: yeah just removed it 😅

ISLE/xbox_one_series/config.h
```diff
@@ -0,0 +1,5 @@
+#pragma once
```
foxtacles: isle style include guards please

Helloyunho: oh right forgot about that sorry

ISLE/isleapp.cpp
foxtacles: https://github.com/isledecomp/isle-portable/issues/534

CMakeLists.txt
Helloyunho: Though again that config is only for xbox... What should be the way to distinguish between windows and xbox..?

foxtacles (approved):
LGTM. Would like approval from @madebr and @AJenbo as well

foxtacles:
> Controller does not work when it's connected after the game has launched

Let's create an issue for the missing joystick hot plug functionality in case this is still an issue

Helloyunho:
It is still a problem so I'll make an issue about that

Helloyunho:
...made an issue at #535

CMakeLists.txt
AJenbo: That uwp is not intended for Windows :shrug:

AJenbo (approved):

CMakeLists.txt
Helloyunho: it certainly not, I'll mark this as resolved for now

miniwin/src/d3drm/backends/sdl3gpu/shaders/generated/ShaderIndex.cpp
```diff
@@ -143,4 +179,4 @@ const SDL_GPUShaderCreateInfo* GetFragmentShaderCode(FragmentShaderId id, SDL_GP
   }
   SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "Could not find fragment shader code for id=%d and formats=0x%x", id, formats);
   return nullptr;
-}
+}
```
madebr: Final newline

.github/workflows/ci.yml
```diff
@@ -143,6 +144,15 @@ jobs:
           cd build
           mkdir dist
           mv *.3dsx dist/
+      
+      - name: Package (Xbox One)
```
madebr: This can be removed: you added install instructions to the cmake script.

Helloyunho: cpack contains not only msix and deps but also exe which is unnecessary
if it's fine then i'll change it

madebr: Installing the raw binary does not make sense for more platforms.
e.g. on n3ds, the elf is also useless (ignoring debugging)

We want to use CPack here, and not the manual packaging to avoid creating a continuous release with all xbos files separate: they should be grouped.

So I'd be fine with doing something as:
```cmake
if(NOT (NINTENDO_3DS OR WINDOWS_STORE))
  install(TARGETS isle ...)
endif()
```

Helloyunho: got it, I'll also apply that to prevent raw bins

madebr: The `Xbox One` archive looks empty on ci.

miniwin/src/ddraw/ddraw.cpp
```diff
@@ -357,7 +357,7 @@ HRESULT DirectDrawImpl::CreateDevice(
 		DDRenderer = new Citro3DRenderer(DDSDesc.dwWidth, DDSDesc.dwHeight);
 	}
 #endif
-#ifdef _WIN32
+#if defined(_WIN32) && !defined(WINDOWS_STORE)
```
jpeisach, line 360: I'm assuming that we won't be distributing UWP on desktop computers, that would be my only concern here if we were. Otherwise LGTM

Helloyunho: yeah UWP is not for Windows desktop

.github/workflows/ci.yml
Helloyunho: oh wait I'll check

Helloyunho: @madebr Fixed it

madebr (approved):

PR #527, opened 2025-07-05
Support gamepad

Helloyunho:
This PR adds support for gamepads and replaces Joystick API to SDL's Gamepad API as talked in [here](https://github.com/isledecomp/isle-portable/pull/526#discussion_r2186899706).

## Button Mapping (Based on Xbox, should work for every gamepads)
- A: skip cutscenes
- Menu: exit game
- Right thumbstick: move virtual cursor
- Left thumbstick: move character

## Tested Platforms
- macOS with Xbox Controller
- Steam Deck
- Nintendo 3DS (see the conversation below)

### DIFF
--- a/ISLE/isleapp.cpp
+++ b/ISLE/isleapp.cpp
@@ -386,7 +386,7 @@ SDL_AppResult SDL_AppIterate(void* appstate)
 			g_mousemoved = FALSE;
 		}
 
-		MoveVirtualCursorViaJoystick();
+		g_isle->MoveVirtualMouseViaJoystick();
 	}
 
 	return SDL_APP_CONTINUE;
@@ -1276,7 +1276,7 @@ IDirect3DRMMiniwinDevice* GetD3DRMMiniwinDevice()
 	return d3drmMiniwinDev;
 }
 
-void MoveVirtualCursorViaJoystick()
+void IsleApp::MoveVirtualMouseViaJoystick()
 {
 	if (g_lastJoystickMouseX != 0 || g_lastJoystickMouseY != 0) {
 		g_mousemoved = TRUE;
--- a/ISLE/isleapp.h
+++ b/ISLE/isleapp.h
@@ -60,6 +60,7 @@ public:
 	MxResult ParseArguments(int argc, char** argv);
 	MxResult VerifyFilesystem();
 	void DetectGameVersion();
+	void MoveVirtualMouseViaJoystick();
 
 private:
 	char* m_hdPath;              // 0x00
@@ -106,7 +107,5 @@ extern IsleApp* g_isle;
 extern MxS32 g_closed;
 
 extern IDirect3DRMMiniwinDevice* GetD3DRMMiniwinDevice();
-
-void MoveVirtualCursorViaJoystick();
 
 #endif // ISLEAPP_H


ISLE/isleapp.cpp
```diff
@@ -90,6 +90,11 @@ MxS32 g_targetDepth = 16;
 // GLOBAL: ISLE 0x410064
 MxS32 g_reqEnableRMDevice = FALSE;
 
+float g_lastJoystickMouseX = 0;
```
jpeisach: MxFloat maybe for these constants?

ISLE/isleapp.cpp
```diff
@@ -893,9 +979,10 @@ bool IsleApp::LoadConfig()
 	m_wideViewAngle = iniparser_getboolean(dict, "isle:Wide View Angle", m_wideViewAngle);
 	m_use3dSound = iniparser_getboolean(dict, "isle:3DSound", m_use3dSound);
 	m_useMusic = iniparser_getboolean(dict, "isle:Music", m_useMusic);
-	m_useJoystick = iniparser_getboolean(dict, "isle:UseJoystick", m_useJoystick);
-	m_joystickIndex = iniparser_getint(dict, "isle:JoystickIndex", m_joystickIndex);
+	m_useGamepad = iniparser_getboolean(dict, "isle:UseJoystick", m_useGamepad);
```
jpeisach: Update the string; also I think we need to convey the clarification that Gamepad implies Joystick to the end user

ISLE/isleapp.cpp
Helloyunho: oh you're right

ISLE/isleapp.cpp
Helloyunho: it will break old configs though...

ISLE/isleapp.cpp
```diff
@@ -379,6 +385,27 @@ SDL_AppResult SDL_AppIterate(void* appstate)
...
+					c_notificationMouseMove,
+					g_mousedown ? LegoEventNotificationParam::c_lButtonState : 0,
+					g_lastMouseX,
+					g_lastMouseY,
+					0
+				);
+			}
+
+			if (g_isle->GetDrawCursor()) {
+				VideoManager()->MoveCursor(Min((MxS32) g_lastMouseX, 639), Min((MxS32) g_lastMouseY, 479));
```
jpeisach: This is a small nitpick that doesn't have to be addressed in this PR. We should use `g_targetWidth - 1` and `g_targetHeight - 1` where applicable

ISLE/isleapp.cpp
```diff
@@ -457,6 +484,63 @@ SDL_AppResult SDL_AppEvent(void* appstate, SDL_Event* event)
 		}
 		break;
 	}
+	case SDL_EVENT_GAMEPAD_BUTTON_DOWN: {
+		{
+			if (event->gbutton.button == SDL_GAMEPAD_BUTTON_SOUTH) {
```
jpeisach, line 470: On 3DS this maps to the B button (I just tested). I'm not sure what the best way of ifdefing or setting a value on 3DS to map to the A button would be. I guess B button is fine, people could easily figure it out and get used to it, but, /shrug

ISLE/isleapp.cpp
Helloyunho: should i fix it in this pr?

ISLE/isleapp.cpp
jpeisach: Good point. It's an internal value anyways.

ItzSwirlz (approved):
Works on 3DS, glad that the start button maps to escape to the infocenter. Circle pad works as expected.

ISLE/isleapp.cpp
jpeisach: No, just something I want to speak up about. You can mark this as resolved :)

AJenbo: i mean it wouldn't hurt to fix it

ISLE/isleapp.cpp
```diff
@@ -113,8 +118,8 @@ IsleApp::IsleApp()
 	m_drawCursor = FALSE;
 	m_use3dSound = TRUE;
 	m_useMusic = TRUE;
-	m_useJoystick = TRUE;
-	m_joystickIndex = 0;
+	m_useGamepad = TRUE;
```
foxtacles: We should not rename the variables, function names etc from Joystick to Gamepad. Let's keep what we had. This will facilitate easier merging and avoid confusion. Right now half of the code uses gamepad terminology, the other joystick. Let's settle on the original joystick wording.

foxtacles (changes requested):

ISLE/isleapp.cpp
Helloyunho: I'll leave this for another pr since IDK how to do it so far 😅

ISLE/isleapp.cpp
AJenbo: Use https://wiki.libsdl.org/SDL3/SDL_GamepadType to detect if it should be one or the other.

ISLE/isleapp.cpp
Helloyunho: Got it, I'll rename those back to joystick

ISLE/isleapp.cpp
Helloyunho: IDK where 3DS belongs to...

AJenbo: I don't think you need to be concerned about that for this PR, it will likely be SDL_GAMEPAD_TYPE_NINTENDO_SWITCH_PRO, SDL_GAMEPAD_TYPE_STANDARD or SDL_GAMEPAD_TYPE_UNKNOWN. A 3DS matineerne should come in afterwards and add a build time overwrite if it's not detected automatically.

But it would be nice if you can lead by doing the basic detection using that enum.

Helloyunho: Got it, I'm gonna mark this as resolved

ISLE/isleapp.cpp
```diff
@@ -379,6 +385,27 @@ SDL_AppResult SDL_AppIterate(void* appstate)
 		if (g_mousemoved) {
 			g_mousemoved = FALSE;
 		}
+
+		if (g_lastJoystickMouseX != 0 || g_lastJoystickMouseY != 0) {
```
foxtacles: Is there a reason not to put this code in the event handler directly? Otherwise it seems this would fire every tick as long as these values aren't zero?

Helloyunho: SDL_AppEvent gives event only when the joystick value is changed and not when it's reporting the same value
So for example if you hold your joystick to the most top and stay like that it will not fire the event

foxtacles: Ok, but why do we need to fire an event for something that hasn't changed here? Obtaining the current state of the joystick/gamepad is meant to be handled in `LegoInputManager::GetJoystickState` (same thing for touch, keyboard etc.).

Helloyunho: If you check the code, it doesn't fire when the mouse doesn't have to move(well, it does when mouse goes to the end of the screen and you try to go further than that)

foxtacles: I see, so this is only for simulating mouse motion events based on the state of the joystick?

Let's please take this piece of code and put it into its own function at the bottom of `isleapp.cpp`, and add a comment that it is doing that. This is the only instance where we generate an event for isle that doesn't directly correlate to an event we received from SDL so it is a bit special.

Helloyunho: Got it, what should be the function name? `MoveVirtualCursorViaJoystick`?

foxtacles: Sounds good

ISLE/isleapp.h
```diff
@@ -105,4 +107,6 @@ extern MxS32 g_closed;
 
 extern IDirect3DRMMiniwinDevice* GetD3DRMMiniwinDevice();
 
+void MoveVirtualCursorViaJoystick();
```
foxtacles: Can you move it into the `IsleApp` class as a member function?

Helloyunho: sure just in sec

ISLE/isleapp.cpp
```diff
@@ -1207,3 +1275,27 @@ IDirect3DRMMiniwinDevice* GetD3DRMMiniwinDevice()
 	}
 	return d3drmMiniwinDev;
 }
+
+void MoveVirtualCursorViaJoystick()
```
foxtacles: Maybe an even better name would be `MoveVirtualMouseViaJoystick`. The cursor is only moved when `GetDrawCursor` is active.

Helloyunho: got it

foxtacles (approved):

Issue #430, opened 2025-06-25
OpenGL: Transition glitched on NVIDIA GPUs (Proprietary Driver)


StevenSYS:
On an NVIDIA GPU, the transition effect looks glitched with the OpenGL 1.5 and OpenGL ES 2.0 renderers

https://github.com/user-attachments/assets/c8f1fd31-0158-4a7f-b2c7-581874806b74

OS: Arch Linux (6.12.34-1-lts)
CPU: AMD FX-4300 (Yes I know this CPU is terrible)
GPU: NVIDIA RTX 3050 8GB (nvidia-lts 575.64)

AJenbo:
As best I can tell this happens because it is presenting the back buffer without waiting for a call to `SDL_GL_SwapWindow()`. If you remove the m_dirty guard you get the same effect on other system. This could be because double buffering is disabled at a driver level but I'm not sure.


```cpp
void OpenGL1Renderer::Flip()
{
	if (m_dirty) {
		SDL_GL_SwapWindow(DDWindow);
		m_dirty = false;
	}
}
```

Does this also happen in windowed mode? It would be good if someone on Windows can also test this.

StevenSYS:
> Does this also happen in windowed mode?

In the video, I have it in windowed mode

kernaltrap8:
Also happening on AMD.
```
GPU0:
        apiVersion         = 1.4.305
        driverVersion      = 25.0.5
        vendorID           = 0x1002
        deviceID           = 0x73df
        deviceType         = PHYSICAL_DEVICE_TYPE_DISCRETE_GPU
        deviceName         = AMD Radeon RX 6750 XT (RADV NAVI22)
        driverID           = DRIVER_ID_MESA_RADV
        driverName         = radv
        driverInfo         = Mesa 25.0.5-2
        conformanceVersion = 1.4.0.0
        deviceUUID         = 00000000-0800-0000-0000-000000000000
        driverUUID         = 414d442d-4d45-5341-2d44-525600000000
```

https://github.com/user-attachments/assets/99c1020f-a754-4d3e-acb8-7dcd6ef89870

AJenbo:
Interesting since it's not happening on Radeon 7800xt using radv

aroenai:
Happens on macOS too (OpenGL 1.1, M2 Macbook Pro)

https://github.com/user-attachments/assets/b8bc56eb-7496-40ff-9b90-027bbe384e87

kernaltrap8:
[...]

PR #530, opened 2025-07-06
Faster and cleaner mosaic transition

AJenbo:
This download the buffer once, samples the relevant colors for the 64x48 effect. Then on subsequent steps in the transition it loads them gradually into a 64x48 texture leaving the GPU to scale it to full screen.

This has the follow benifits:
- Makes it light enough that it can run on low end devices like PSP
- Avoids some potential issues with multi buffers causing flickering
- Retain a high quality image instead of down scaling to 640x480 during the effect
- Avoid smearing and drifting tiles caused by scaling precision when not running in 640x480

Probably solves: #430

New:

https://github.com/user-attachments/assets/f540aec4-e364-46d8-b148-122a6d4762ed

image

**Examples of issue with the old implementation:**

Resolution drop during transition:

image

Pixel drift from resize rounding (PSP at 480x272, can also happen at high resolution), and flickering from multi buffering, poor performance because of slow download/upload: 

https://github.com/user-attachments/assets/27a3a433-0fab-4b5a-87cd-101ec26bb4f8


foxtacles:
broken on dx5


https://github.com/user-attachments/assets/1521300c-c243-4709-804f-8fbe81acfd9d

AJenbo:
@foxtacles please give it another try

foxtacles (approved):
works on dx5

PR #533, opened 2025-07-06
Some quick dpad mouse control

AJenbo:
Not every gamepad has two joysticks, but might have a dpad.

### DIFF
--- a/ISLE/isleapp.cpp
+++ b/ISLE/isleapp.cpp
@@ -95,6 +95,11 @@ MxFloat g_lastJoystickMouseY = 0;
 MxFloat g_lastMouseX = 320;
 MxFloat g_lastMouseY = 240;
 
+bool g_dpadUp = false;
+bool g_dpadDown = false;
+bool g_dpadLeft = false;
+bool g_dpadRight = false;
+
 // STRING: ISLE 0x4101dc
 #define WINDOW_TITLE "LEGO®"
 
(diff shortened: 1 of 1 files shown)


ISLE/isleapp.cpp
```diff
@@ -392,6 +392,11 @@ SDL_AppResult SDL_AppIterate(void* appstate)
 	return SDL_APP_CONTINUE;
 }
 
+static bool dpadUp = false;
```
foxtacles: let's keep those at the top of the file as regular isle globals

ISLE/isleapp.cpp
```diff
@@ -1276,13 +1337,35 @@ IDirect3DRMMiniwinDevice* GetD3DRMMiniwinDevice()
 	return d3drmMiniwinDev;
 }
 
+const float DPAD_MOUSE_SPEED = 3.0f;
```
foxtacles: same

Helloyunho:
Tested on macOS with Xbox Controller and it works!

foxtacles (approved):

ISLE/isleapp.cpp
```diff
@@ -95,6 +95,12 @@ MxFloat g_lastJoystickMouseY = 0;
 MxFloat g_lastMouseX = 320;
 MxFloat g_lastMouseY = 240;
 
+bool g_dpadUp = false;
+bool g_dpadDown = false;
+bool g_dpadLeft = false;
+bool g_dpadRight = false;
+const float g_dpadMouseSpeed = 3.0f;
```
Helloyunho: Maybe use the same cursor sensitivity variable?

AJenbo: Good catch

Helloyunho:
Just a quick thoughts of later work: Add a toggle key to change between dpad to move character <-> dpad to move mouse so you can use SNES style controller as well

AJenbo:
lets wait till someone tries to target that :D

This was for PSP as the main target.

PR #544, opened 2025-07-07
Fix clear screen on buffered devices

### LINKED ISSUE #465: Flickering on credits screen
When using the OpenGL ES 2.0 or 1.1 HAL rendering device, the background of the credits screen flashes. I suspect this is to do with the credits screen not being built for higher framerates. I am on Endeavour OS (based on Arch Linux) on Wayland with an Nvidia GPU and 575.64 drivers. This happens both in fullscreen and in windowed mode.

Note: Do not watch the video if you are prone to seizures from flashing lights.

https://github.com/user-attachments/assets/ac81a2e2-a0d6-4c9c-a27c-f31f93cefca0

AJenbo:
Fixes #465

### DIFF
--- a/LEGO1/omni/src/video/mxdisplaysurface.cpp
+++ b/LEGO1/omni/src/video/mxdisplaysurface.cpp
@@ -52,7 +52,7 @@ void MxDisplaySurface::ClearScreen()
 	DDSURFACEDESC desc;
 
 	if (!m_videoParam.Flags().GetFlipSurfaces()) {
-		backBuffers = 1;
+		backBuffers = 2;
 	}
 	else {
 		backBuffers = m_videoParam.GetBackBuffers() + 1;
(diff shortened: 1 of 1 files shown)


LEGO1/omni/src/video/mxdisplaysurface.cpp
```diff
@@ -52,7 +52,7 @@ void MxDisplaySurface::ClearScreen()
 	DDSURFACEDESC desc;
 
 	if (!m_videoParam.Flags().GetFlipSurfaces()) {
-		backBuffers = 1;
+		backBuffers = 2;
```
foxtacles, line 55: Could you explain this? How will this make a difference in the loop below?

AJenbo: clearing one buffer is not enough on double buffered systems.

foxtacles: Right, it's just confusing because `i` isn't being used in the loop logic. So I'm wondering what exactly happens.

foxtacles (approved):

PR #545, opened 2025-07-07
Align FakeMosaicTransition with MosaicTransition

AJenbo:
Should make it work on Dx5 and generally be more compatible ... still not transparent on WebGL though (works on OpenGL ES and all other backends)


foxtacles:
Seems to work on dx5 (if the colors are expected that way):

https://github.com/user-attachments/assets/cf43931b-f286-44fe-8854-92ee9b5c4f7e

foxtacles (approved):

AJenbo:
The palette is a little off but that wasn't totally unexpected.

PR #541, opened 2025-07-06
Add PSVita port

olebeck:
psvita port using gxm
some todos for this:

- [x] paf based config app
[?] download cd in config app maybe
- [ ] mipmaps
- [x] msaa (if enough fragment performance left over)
- [ ] something in game to tell the player what the gamepad layout is
- [x] flickering on the credits https://github.com/isledecomp/isle-portable/pull/544

PR #548, opened 2025-07-07
Android port

AJenbo:
Read this PR per commit, the first one is just the 6000 lines of the default SDL3 Android project folder.

Todo:
- [ ] Backgrounds not loading
- [ ] Touch scaling https://github.com/isledecomp/isle-portable/pull/541
- [x] Icon
- [x] Default config


madebr:
The Java sources should be considered part of SDL3, and are only compatible with the same version.
We should avoid any solution that copies the SDL java sources to the isle-portable repo.

Let's use the android-devel package from SDL3 instead: see [the release page of SDL3](https://github.com/libsdl-org/SDL/releases/tag/release-3.2.16).
When using this Android aar-chive, it becomes easy to switch SDL3 version.
[SDL_helloworld](https://github.com/libsdl-org/SDL_helloworld) contains a minimum android gradle project.

AJenbo:
Sorry but can you describe how that is done in a bit more detail? I see your name all over that repo, maybe it's something you know how to do easily?

madebr:
> Sorry but can you describe how that is done in a bit more detail? I see your name all over that repo, maybe it's something you know how to do easily?

Download the `SDL3-devel-3.2.16-android.zip` archive and read its `INSTALL.md`.
The zip contains the java classes + pre-built SDL3 binary.

You need to add this to the gradle project:
```gradle
android {
    /* ... */
    buildFeatures {
        prefab true
    }
}
dependencies {
    implementation files('libs/SDL3-3.2.16.aar')
    /* ... */
}
```

Alternatively, we use the java classes from the SDL3 git submodule and somehow instruct gradle to use them.

@tahg , sorry to bother you.
But do you know how to configure a gradle project to use java sources that are not in a subdirectory?
We want a gradle project at `android-project/app/build.gradle` to use java classes at `external/SDL/android-project/app/src/main`.
If this requires changes in the SDL repo, that's no problem.

android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
```diff
@@ -0,0 +1,2228 @@
...
+
+
+/**
+    SDL Activity
+*/
+public class SDLActivity extends Activity implements View.OnSystemUiVisibilityChangeListener {
+    private static final String TAG = "SDL";
+    private static final int SDL_MAJOR_VERSION = 3;
+    private static final int SDL_MINOR_VERSION = 3;
+    private static final int SDL_MICRO_VERSION = 0;
```
madebr, line 64: This is the reason why I want to avoid vendoring the java sources.
The version in the java classes must remain in sync with the SDL git submodule.

The compatible SDL version is declared here ...

android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
```diff
@@ -0,0 +1,2228 @@
...
+        if (!mBrokenLibraries) {
+            String expected_version = String.valueOf(SDL_MAJOR_VERSION) + "." +
+                                      String.valueOf(SDL_MINOR_VERSION) + "." +
+                                      String.valueOf(SDL_MICRO_VERSION);
+            String version = nativeGetVersion();
+            if (!version.equals(expected_version)) {
+                mBrokenLibraries = true;
+                errorMsgBrokenLib = "SDL C/Java version mismatch (expected " + expected_version + ", got " + version + ")";
+            }
+        }
```
madebr, line 404: ... and checked here.

android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
AJenbo: Yeah, for DevilutionX we have a script we run to update the Java src when ever we update SDL. But we also don't run on bleeding edge ... with continuous release.

android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
AJenbo: yep

android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
madebr: That would work too as it needs no extra steps.

AJenbo:
I don't think I have the will to setup the rest of this, there is still the crash, CI and loose ends to tie up. I mostly just needed to know if OpenGL ES 2.0 would work on new Androids devices, they do. But if anyone wants to work on an android port then feel free to pick the bits from this PR if it saves you time.

Issue #519, opened 2025-07-04
macOS - OpenGL 1.1 HAL option not available on recent builds


aroenai:
Builds after 6/27/25 no longer have the OpenGL 1.1 HAL as a selectable option using the same build environment on macOS with Apple Silicon processors.

aroenai:
Appears to have started after e2ff65c

AJenbo:
Hm unfortunately I don't have a Mac to test with, does this issue still happen if you don't make the test window hidden? Can the game still be made to use opengl by editing the config?

aroenai:
> Hm unfortunately I don't have a Mac to test with, does this issue still happen if you don't make the test window hidden? Can the game still be made to use opengl by editing the config?

If you edit the isle.ini file manually for OpenGL, the game uses the SDL3 GPU HAL (credits flicker, which doesn't happen on OpenGL). If you do a quick and dirty "git revert e2ff65c" it works normally (there's conflicts on miniwin/src/d3drm/backends/opengles2/renderer.cpp but GLESv2 doesn't work on Mac anyway).

AJenbo:
Ok, then someone on macOS should look in to what causes it to fail to initialize after the change since it's not simply a question about the test window which is what that commit aimed to address.

AJenbo:
My best guess is that this happens because because macOS doesn't allow OpenGL to use a window after it's been interacted with by Metal. Can you try and add a return before this line and see if that allows you to set OpenGL as the renderer.

https://github.com/isledecomp/isle-portable/blob/c6b94dabcf1a75c1b757547e2325a6b4c7ba0f48/miniwin/src/internal/d3drmrenderer_sdl3gpu.h#L123

> SDL3 GPU HAL (credits flicker)

This has been addressed by the way

aroenai:
> My best guess is that this happens because because macOS doesn't allow OpenGL to use a window after it's been interacted with by Metal. Can you try and add a return before this line and see if that allows you to set OpenGL as the renderer.
> 
> [isle-portable/miniwin/src/internal/d3drmrenderer_sdl3gpu.h](https://github.com/isledecomp/isle-portable/blob/c6b94dabcf1a75c1b757547e2325a6b4c7ba0f48/miniwin/src/internal/d3drmrenderer_sdl3gpu.h#L123)
> 
> Line 123 in [c6b94da](/isledecomp/isle-portable/commit/c6b94dabcf1a75c1b757547e2325a6b4c7ba0f48)
> 
[...]

PR #552, opened 2025-07-08
Use minimal test for SDL_GPU on macOS

AJenbo:
This attempts to address #519 

Apparently Metal does not properly restore Window state after shutdown which prevents OpenGL from functioning properly. This avoids the issue by only checking for a functional device.

### DIFF
--- a/miniwin/src/internal/d3drmrenderer_sdl3gpu.h
+++ b/miniwin/src/internal/d3drmrenderer_sdl3gpu.h
@@ -120,11 +120,19 @@ private:
 
 inline static void Direct3DRMSDL3GPU_EnumDevice(LPD3DENUMDEVICESCALLBACK cb, void* ctx)
 {
+#ifdef __APPLE__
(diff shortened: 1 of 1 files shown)


foxtacles (approved):

miniwin/src/internal/d3drmrenderer_sdl3gpu.h
```diff
@@ -120,11 +120,17 @@ class Direct3DRMSDL3GPURenderer : public Direct3DRMRenderer {
 
 inline static void Direct3DRMSDL3GPU_EnumDevice(LPD3DENUMDEVICESCALLBACK cb, void* ctx)
 {
-	Direct3DRMRenderer* device = Direct3DRMSDL3GPURenderer::Create(640, 480);
+#ifdef __APPLE__
+	if (!SDL_CreateGPUDevice(SDL_GPU_SHADERFORMAT_MSL, false, nullptr)) {
+		return;
+	}
```
madebr: You need to destroy the device if it succeeds.

madebr (approved):

PR #559, opened 2025-07-08
Disable offscreen canvases in case of no WebGL support

### LINKED ISSUE #556: WebGL: fallback to software not working if WebGL context cannot be created
On devices that cannot create a WebGL context (i.e. Chromium inside WSL2), the game cannot currently fall back to Software Renderer. I think the only way to solve this would probably be to fix this in Emscripten, since the transfer to offscreen of the canvas happens at boot time already.

```
SDL_GL_CreateContext: Could not create webgl context
...
Direct3D driver name="Miniwin driver" description="Miniwin Emulation"
isle.js:1 Uncaught (in promise) DOMException: Failed to execute 'getContext' on 'HTMLCanvasElement': Cannot get context from a canvas that has transferred its control to offscreen.
    at Object.createContext (http://localhost:6931/isle.js:1:86191)
    at 402840 (http://localhost:6931/isle.js:1:179565)
    at __emscripten_receive_on_main_thread_js (http://localhost:6931/isle.js:1:39041)
    at http://localhost:6931/isle.wasm:wasm-function[2951]:0x144af4
    at http://localhost:6931/isle.wasm:wasm-function[2953]:0x144bfb
    at http://localhost:6931/isle.wasm:wasm-function[2955]:0x144c84
    at http://localhost:6931/isle.wasm:wasm-function[1247]:0x5cf34
    at http://localhost:6931/isle.wasm:wasm-function[2957]:0x144ca9
    at http://localhost:6931/isle.wasm:wasm-function[1247]:0x5cf34
    at http://localhost:6931/isle.wasm:wasm-function[2943]:0x144574
```

### LINKED ISSUE #558: WebGL: Firefox performance in Infocenter worse than Software
In Firefox, when using WebGL, the performance in the Infocenter seems really poor. Animations are very choppy, worse than when using the Software renderer. The performance on the island is good.

Chrome is perfectly smooth, however.

edit: looks like this can be solved with `Flip Surfaces=ON`. @AJenbo any idea what the major difference could be? Should we enable this by default? (currently it's default `OFF`)

foxtacles:
Close #556 
Close #558 

This sets `disableOffscreenCanvases` for browsers that have no WebGL support. It will then fall back to Software.

It will also set `Flip Surfaces` to `ON` for all Emscripten builds. Performance seems to be much better in WebGL, especially in Firefox.

### DIFF
--- a/ISLE/emscripten/config.cpp
+++ b/ISLE/emscripten/config.cpp
@@ -1,0 +1,17 @@
+#include "config.h"
+
+#include "filesystem.h"
+
+#include <SDL3/SDL_log.h>
+#include <iniparser.h>
+
+void Emscripten_SetupDefaultConfigOverrides(dictionary* p_dictionary)
+{
+	SDL_Log("Overriding default config for Emscripten");
+
+	iniparser_set(p_dictionary, "isle:diskpath", Emscripten_bundledPath);
+	iniparser_set(p_dictionary, "isle:cdpath", Emscripten_streamPath);
+	iniparser_set(p_dictionary, "isle:savepath", Emscripten_savePath);
+	iniparser_set(p_dictionary, "isle:Full Screen", "false");
+	iniparser_set(p_dictionary, "isle:Flip Surfaces", "true");
+}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -532,6 +532,7 @@ if (ISLE_BUILD_APP)
   endif()
   if(EMSCRIPTEN)
     target_sources(isle PRIVATE
+      ISLE/emscripten/config.cpp
       ISLE/emscripten/events.cpp
       ISLE/emscripten/filesystem.cpp
       ISLE/emscripten/messagebox.cpp
--- a/ISLE/emscripten/config.h
+++ b/ISLE/emscripten/config.h
@@ -1,0 +1,8 @@
+#ifndef EMSCRIPTEN_CONFIG_H
+#define EMSCRIPTEN_CONFIG_H
+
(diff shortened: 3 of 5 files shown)


ISLE/emscripten/config.cpp
```diff
@@ -0,0 +1,17 @@
...
+
+void Emscripten_SetupDefaultConfigOverrides(dictionary* p_dictionary)
+{
+	SDL_Log("Overriding default config for Emscripten");
+
+	iniparser_set(p_dictionary, "isle:diskpath", Emscripten_bundledPath);
+	iniparser_set(p_dictionary, "isle:cdpath", Emscripten_streamPath);
+	iniparser_set(p_dictionary, "isle:savepath", Emscripten_savePath);
+	iniparser_set(p_dictionary, "isle:Full Screen", "false");
+	iniparser_set(p_dictionary, "isle:Flip Surfaces", "true");
```
AJenbo, line 16: Doesn't the game switch off flip surface if full screen is false?

foxtacles: Apparently not

foxtacles: I think a lot of code is never actually called, like `MxVideoManager::Create` and subsequently `MxDisplaySurface::Create`.  It is superseded by `LegoVideoManager::Create`

foxtacles: image

AJenbo: > I think a lot of code is never actually called

Yeah I think this game has potential for some major dead code removal :D

PR #562, opened 2025-07-09
Add support for HD textures

AJenbo:
Place the images in `<isle:diskpath>/textures/<texturename>.bmp`

Start pack/examples with most of the island ground textures + jukebox and Infomaniac shirt:
[lego_island_hd_part1.zip](https://github.com/user-attachments/files/21134638/lego_island_hd_part1.zip)
(22 textures)

List of valid texture names:
[texture-names.txt](https://github.com/user-attachments/files/21134648/texture-names.txt)
(120 textures)

Screen shots
image
image
image

Note this does not cover 2D UI elements as they are loaded through a different path.

### DIFF
--- a/LEGO1/lego/legoomni/src/common/legotextureinfo.cpp
+++ b/LEGO1/lego/legoomni/src/common/legotextureinfo.cpp
@@ -56,16 +56,87 @@ LegoTextureInfo* LegoTextureInfo::Create(const char* p_name, LegoTexture* p_text
 		strcpy(textureInfo->m_name, p_name);
 	}
 
-	LPDIRECTDRAW pDirectDraw = VideoManager()->GetDirect3D()->DirectDraw();
-	LegoImage* image = p_texture->GetImage();
-
 	DDSURFACEDESC desc;
 	memset(&desc, 0, sizeof(desc));
 	desc.dwSize = sizeof(desc);
 	desc.dwFlags = DDSD_PIXELFORMAT | DDSD_WIDTH | DDSD_HEIGHT | DDSD_CAPS;
+	desc.ddsCaps.dwCaps = DDSCAPS_TEXTURE | DDSCAPS_SYSTEMMEMORY;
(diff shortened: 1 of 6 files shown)


LEGO1/lego/legoomni/src/common/legotextureinfo.cpp
```diff
@@ -56,16 +56,87 @@ LegoTextureInfo* LegoTextureInfo::Create(const char* p_name, LegoTexture* p_text
...
 
-	LPDIRECTDRAW pDirectDraw = VideoManager()->GetDirect3D()->DirectDraw();
-	LegoImage* image = p_texture->GetImage();
-
 	DDSURFACEDESC desc;
 	memset(&desc, 0, sizeof(desc));
 	desc.dwSize = sizeof(desc);
 	desc.dwFlags = DDSD_PIXELFORMAT | DDSD_WIDTH | DDSD_HEIGHT | DDSD_CAPS;
+	desc.ddsCaps.dwCaps = DDSCAPS_TEXTURE | DDSCAPS_SYSTEMMEMORY;
+	desc.ddpfPixelFormat.dwSize = sizeof(desc.ddpfPixelFormat);
```
foxtacles, line 64: If we are to accept extensions to core functionality it should be done in a minimally invasive way. `LEGO1` should remain as close as possible to the original decomp. Adding logic like that directly into `LEGO1` for non-portability reasons breaks that contract, and at that point we might as well accept arbitrary functionality change contributions. However, as the only active maintainer this is not something I'm willing to support.

I could imagine "extensions" implemented outside `LEGO1` that are optional and integrated through a minimal interface. The ini file could control the extensions:

```
[extensions]
Texture Loader=true
```

This would provide a clear pathway to add new functionality, while at the same time keeping `LEGO1` relatively unmodified. I realize adding this adds a little bit of an abstraction overhead but it's the only way that let's me still maintain this properly.

PR #566, opened 2025-07-09
Add iOS Port

Helloyunho:
## Description

This port was possible mostly because of macOS port

Put the assets to `Files -> On My iPhone -> Isle Portable -> isle` and you're good to go

## Bug
- (Not an actual bug, but more like iOS behavior) when you update the app, the directory path internally changes, requires you to remove(or edit) config
- (Not sure it's intended) left and right rotating is only possible when you touch the bottom area

## Notes
- LEGO Island on a touch device is not a good idea i must say... maybe we need a virtual button to skip dialog, exit game, etc.


madebr (approved):

AJenbo:
iOS is just much nicer to port to then Android, the java layer really sucks

Helloyunho:
yeah it was just a few touch from macos port

PR #567, opened 2025-07-09
Dynamically on/off virtual cursor

### LINKED ISSUE #549: Turn cursor on/off dynamically based on GamePad / Mouse input
Some devices offer both mouse an input, and it isn't clear when either is the preferred or only option.

The in-game mouse is both delayed and unresponsive during segments so doesn't work well as the default for actual mouse use.

Switching the cursor on when ever an Gamepad right stick / depad even is encountered and off when ever a mouse / touch event is recived would make this work dynamically without a need to change the ini (which can be hard on some systems or for some users)

Helloyunho:
Inspired by #549, so closes #549

It hides cursor when you move virtual cursor with joystick, and show the cursor when you move your system cursor with mouse

It also remembers where the cursor was at so you can do continuous movement

Touch devices are not tested yet


Helloyunho:
since it hides virtual cursor by default it fixed that issue lmao

PR #569, opened 2025-07-09
put ifdefs around every d3drm backend so that any can be disabled

olebeck:
split out from psvita port, needed to disable software renderer so tried making that a bit cleaner

### DIFF
--- a/miniwin/src/d3drm/d3drm.cpp
+++ b/miniwin/src/d3drm/d3drm.cpp
@@ -7,24 +7,6 @@
 #include "d3drmmesh_impl.h"
 #include "d3drmobject_impl.h"
 #include "d3drmrenderer.h"
-#ifdef USE_OPENGL1
-#include "d3drmrenderer_opengl1.h"
-#endif
-#ifdef USE_OPENGLES2
-#include "d3drmrenderer_opengles2.h"
-#endif
-#ifdef USE_CITRO3D
-#include "d3drmrenderer_citro3d.h"
-#endif
-#ifdef USE_DIRECTX9
-#include "d3drmrenderer_directx9.h"
-#endif
-#ifdef USE_SDL_GPU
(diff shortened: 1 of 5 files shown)


AJenbo:
> split out from psvita port, needed to disable software renderer so tried making that a bit cleaner

why? I guess it saves on code size, but other then that?

miniwin/src/d3drm/d3drm.cpp
```diff
@@ -146,12 +150,18 @@ HRESULT Direct3DRMImpl::CreateDeviceFromSurface(
 	DDSDesc.dwSize = sizeof(DDSURFACEDESC);
 	surface->GetSurfaceDesc(&DDSDesc);
 
-	if (SDL_memcmp(&guid, &SDL3_GPU_GUID, sizeof(GUID)) == 0) {
+	if (false) {
+	}
```
AJenbo: ???
```suggestion
```

miniwin/src/ddraw/ddraw.cpp
```diff
@@ -343,9 +350,13 @@ HRESULT DirectDrawImpl::CreateDevice(
 	DDSDesc.dwSize = sizeof(DDSURFACEDESC);
 	pBackBuffer->GetSurfaceDesc(&DDSDesc);
 
-	if (SDL_memcmp(&guid, &SDL3_GPU_GUID, sizeof(GUID)) == 0) {
+	if (false) {
+	}
```
AJenbo: is this some sort of compiler hack?
```suggestion
```

olebeck: I'm not sure if there's a prettier way to do it, I did it that way since every condition can be ifdefd out so can't have the first one be the start of the if

AJenbo: Move it to a function that can return, that way we can also get rid of the duplication.

PR #570, opened 2025-07-09
Add extensions, `TextureLoader`

### LINKED ISSUE #568: Implement extensions framework
The goal is to allow extensions to the game, under the following constraints:

* Minimal interfacing with `LEGO1`; no modifications to existing logic
* Sustained in a module outside `LEGO1`
* Optional

This should allow adding functionality as seen in #562 while at the same time being able to cleanly maintain the original experience/code.

I have some ideas on how to structure this so I will come up with a PoC soon.

foxtacles:
Close #568 

This adds a `extensions` module that allows adding code that can directly integrate with `LEGO1`, but using minimal interfacing with the original code. This enables developing additional functionality without sacrificing the maintainability of `LEGO1`.

I've added `TextureLoader` as an initial extension based on #562. It can be enabled by adding the following to the `.ini` file:

```
[extensions]
Texture Loader=true
```

### DIFF
--- a/extensions/src/textureloader.cpp
+++ b/extensions/src/textureloader.cpp
@@ -2,6 +2,8 @@
 
 using namespace Extensions;
 
+bool TextureLoader::enabled = false;
+
 bool TextureLoader::PatchTexture(LegoTextureInfo* p_textureInfo)
 {
 	SDL_Surface* surface = FindTexture(p_textureInfo->m_name);
--- a/extensions/src/textureloader.cpp
+++ b/extensions/src/textureloader.cpp
@@ -1,0 +1,96 @@
+#include "extensions/textureloader.h"
+
+using namespace Extensions;
+
+bool TextureLoader::PatchTexture(LegoTextureInfo* p_textureInfo)
+{
+	SDL_Surface* surface = FindTexture(p_textureInfo->m_name);
+	if (!surface) {
+		return false;
+	}
+
+	const SDL_PixelFormatDetails* details = SDL_GetPixelFormatDetails(surface->format);
+
+	DDSURFACEDESC desc;
+	memset(&desc, 0, sizeof(desc));
+	desc.dwSize = sizeof(desc);
+	desc.dwFlags = DDSD_PIXELFORMAT | DDSD_WIDTH | DDSD_HEIGHT | DDSD_CAPS;
+	desc.ddsCaps.dwCaps = DDSCAPS_TEXTURE | DDSCAPS_SYSTEMMEMORY;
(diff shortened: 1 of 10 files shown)


foxtacles:
> My only concern is `Extension::Call` doing a linear string search, but this is not a performance issue at the moment.

Refactored it a bit, got rid of the linear search and it's a bit cleaner now I think.

extensions/src/textureloader.cpp
```diff
@@ -0,0 +1,98 @@
...
+{
+	SDL_Surface* surface;
+	MxString path = MxString(MxOmni::GetHD()) + texturePath + p_name + ".bmp";
+
+	path.MapPathToFilesystem();
+	if (!(surface = SDL_LoadBMP(path.GetData()))) {
+		path = MxString(MxOmni::GetCD()) + texturePath + p_name + ".bmp";
+		path.MapPathToFilesystem();
+		surface = SDL_LoadBMP(path.GetData());
+	}
```
madebr, line 95: Do you intend to add a dependency to SDL3_Image to load other (compressed) formats?

foxtacles: That would probably make sense. I'm not planning on working on this extension myself at this time (I mostly copy pasted @AJenbo code from #562 ), this PR is mainly to pave the way for others who are interested to do so

Helloyunho:
Just a quick thought about this: maybe we can create extensions folder and when user adds a dll(or dylib, so, etc) it enables that extension
And in API side we give them a way to hook all the lego1 functions as they want kinda like a game mods

extensions/src/textureloader.cpp
AJenbo: with only 120 textures it's not like it takes up a crazy amount of space. you can probably make most texture 256 colors to have them even smaller. All mine where 32bit because of some early issues with loading them correctly. SDL_Image or libpng could be added for for better on disk formats, but I don't think it's worth it for the added complexity.

madebr: > but I don't think it's worth it for the added complexity.

:100: Building all decoder libraries for all target platforms will not make things easier.

AJenbo: On DevilutionX we have a minimalisic version of SDL_Images that only supports PNG

madebr: The CMake of SDL_image is able to build all dependencies (and disable them) , but it's imho bad practice to vendor everything. CI times will be monstrous.

But as you said, let's not get into this mess (yet).

PR #579, opened 2025-07-11
Add support for Discord RPC and allow enabling extensions via config app

matu6968:
Ever since the pull request #570 was merged, i have **decided to implement a Discord RPC extension with the help of AI** that exposes "playing LEGO Island" to a compatible Discord client and it shows how much you played, what character you are playing with and what act you are on.

It looks like this within the Discord client:

(excuse for the blank icon, due to Discord caching it won't show up immediately)
Screenshot from 2025-07-11 14-42-56

This feature only works on desktop platforms like Windows, Mac OS and Linux and will be disabled if compiling for a different platform like Emscripten but i have not tested it yet on other platforms other then Linux to see if it would be affected.

Also since the extensions feature isn't easily accessible to the user, i have decided to also add a descrete option the in config application to be able to enable them:

Screenshot from 2025-07-11 14-48-13

This should now be able for users to easily enable the extensions via the config UI.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,7 +70,7 @@ add_library(Isle::iniparser INTERFACE IMPORTED)
 
 if (DOWNLOAD_DEPENDENCIES)
   # FetchContent downloads and configures dependencies
-  message(STATUS "Fetching SDL3 and iniparser. This might take a while...")
+  message(STATUS "Fetching SDL3, iniparser and discord-rpc (on desktop platforms). This might take a while...")
   include(FetchContent)
   if (WINDOWS_STORE)
     FetchContent_Declare(
@@ -89,6 +89,10 @@ if (DOWNLOAD_DEPENDENCIES)
   endif()
   FetchContent_MakeAvailable(SDL3)
 
+  # Disable iniparser tests before fetching it
+  set(INIPARSER_BUILD_TESTS OFF CACHE BOOL "Disable iniparser tests" FORCE)
+
+  # iniparser dependency
   FetchContent_Declare(
     iniparser
     GIT_REPOSITORY "https://gitlab.com/iniparser/iniparser.git"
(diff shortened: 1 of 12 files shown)


CMakeLists.txt
```diff
@@ -70,7 +70,7 @@ add_library(Isle::iniparser INTERFACE IMPORTED)
 
 if (DOWNLOAD_DEPENDENCIES)
   # FetchContent downloads and configures dependencies
-  message(STATUS "Fetching SDL3 and iniparser. This might take a while...")
+  message(STATUS "Fetching SDL3, iniparser and discord-rpc (on desktop platforms). This might take a while...")
```
foxtacles, line 73: We should not pull in dependencies for extensions like that.

Instead, that should create separate CMakeLists.txt inside the extensions folder, that can do this individually for each extension.

Also, we need to change the `ISLE_EXTENSION` configuration option to be a list of keys instead of a Boolean. My initial thought was that extensions would be lightweight for now but compiling extra dependencies by default is not acceptable, so something like this extension needs to be opt-in at build time.

Subsequently we will need macros defined for each compiled extension like `EXTENSION_TEXTURE_LOADER`, `EXTENSION_DISCORD_RPC` etc. to control available extensions etc . in the code.

There might be better ways to do this.

CC @madebr

ISLE/isleapp.cpp
```diff
@@ -1441,3 +1457,42 @@ void IsleApp::MoveVirtualMouseViaJoystick()
 		}
 	}
 }
+
+void IsleApp::InitializeDiscordRPC()
```
foxtacles, line 1461: The idea of extensions is to have the code in the extension and call it through Extension::Call. None of this code should be in isleapp.cpp

extensions/src/discordrpc.cpp
```diff
@@ -0,0 +1,189 @@
...
+#include <discord_rpc.h>
+
+namespace DiscordRPC
+{
+bool enabled = false;
+
+// Discord RPC configuration
+// This includes a already provided application ID, but you can change it if you want
+static const RPCConfig rpcConfig = {
+    "1392967803421589544", // If needed, replace with your Discord application ID
```
foxtacles, line 17: We almost definitely don't want to hardcore this

foxtacles (changes requested):

foxtacles:
> with the help of AI

Looks like it since it needs a lot of cleanup. I've provided some guidance but I'd generally recommend to try to understand how this works first because this costs us a lot of extra time.

extensions/include/extensions/extensions.h
```diff
@@ -8,7 +8,7 @@
 
 namespace Extensions
 {
-constexpr const char* availableExtensions[] = {"extensions:texture loader"};
+constexpr const char* availableExtensions[] = {"extensions:texture loader", "extensions:discord rpc"};
```
jpeisach, line 11: We should actually find a way to ifdef this so we aren't loading or offering this extension on platforms that don't have discord

extensions/src/discordrpc.cpp
```diff
@@ -0,0 +1,189 @@
...
+        return "Act 3: The Final Showdown";
+    default:
+        return "Unknown Act";
+    }
+}
+
+std::string GetAreaName(int area)
+{
+    switch (area) {
+    case 1: return "LEGO Island";
```
jpeisach, line 109: If possible use the LegoGameState enums

extensions/src/discordrpc.cpp
```diff
@@ -0,0 +1,189 @@
...
+    case 48: return "Jet Ski";
+    default: return "Unknown Area";
+    }
+}
+
+std::string GetActorName(int actorId)
+{
+    switch (actorId) {
+    case 0: return "Pepper Roni";
+    case 1: return "Mama";
```
jpeisach, line 165:
```suggestion
    case 1: return "Mama Brickolini";
```

extensions/src/discordrpc.cpp
```diff
@@ -0,0 +1,189 @@
...
+    case 2: return "Papa";
+    case 3: return "Nick";
```
jpeisach, line 167:
```suggestion
    case 3: return "Nick Brick";
```

extensions/src/discordrpc.cpp
```diff
@@ -0,0 +1,189 @@
...
+    default: return "Unknown Area";
+    }
+}
+
+std::string GetActorName(int actorId)
+{
+    switch (actorId) {
+    case 0: return "Pepper Roni";
+    case 1: return "Mama";
+    case 2: return "Papa";
```
jpeisach, line 166:
```suggestion
    case 2: return "Papa Brickolini";
```

extensions/src/discordrpc.cpp
```diff
@@ -0,0 +1,189 @@
...
+    case 3: return "Nick";
+    case 4: return "Laura";
```
jpeisach, line 168:
```suggestion
    case 4: return "Laura Brick";
```

extensions/src/discordrpc.cpp
```diff
@@ -0,0 +1,189 @@
...
+    default: return "Unknown Character";
+    }
+}
+
+std::string GetActivityDescription(const GameStateInfo& gameState)
```
jpeisach, line 173: I recommend leveraging LegoGameState to be able to include actions such as:
"Racing"
"Roaming"
"Building"

extensions/src/discordrpc.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+std::string GetAreaName(int area)
+{
+    switch (area) {
+    case 1: return "LEGO Island";
+    case 2: return "Information Center";
+    case 3: return "Information Door";
+    case 4: return "Elevator Bottom";
+    case 5: return "Elevator Ride";
+    case 6: return "Elevator Ride 2";
```
jpeisach, line 114: Probably no need to include extra stuff like this, and the "Exited"

jpeisach:
Can't build on macOS:
```
CMake Error at CMakeLists.txt:148 (add_library):
  add_library cannot create target "discord-rpc" because another target with
  the same name already exists.  The existing target is a shared library
  created in source directory
  "/Users/jpeisach/isle-portable/build/_deps/discord-rpc-src/src".  See
  documentation for policy CMP0002 for more details.
```

VoxelTek:
I'd recommend reducing the focus of this just to the Discord RPC extension, as I'm working on a more modular and expandable `isle-config` extensions manager, especially considering extensions may have many of their own settings (which the Texture Loader now does).

PR #582, opened 2025-07-11
Fix ConvertEventToRenderCoordinates for mouse button events

foxtacles:
I'm also wondering what the reason is for converting to `int` during the calculation? Aren't we needlessly losing precision?


AJenbo:
What's broken?

AJenbo:
> I'm also wondering what the reason is for converting to `int` during the calculation?

I think they where int in SDL2 i didn't know that SDL3 changed this. I'm not sure anything else is setup to handle the extra precision though.

PR #590, opened 2025-07-13
Allow extensions to define options

foxtacles:


AJenbo:
I don't think I can make a meaningful review here, but I don't see anything bad as such.

PR #596, opened 2025-07-14
Add haptic feedback (rumble)

foxtacles:
This adds a basic rumble effect, supported on certain gamepads. It also adds an `.ini` option `Haptic`, which is on by default since I think it's a natural fit.

Currently, the only game event that triggers a rumble is when you run into a Lego character (crash), but I'm planning to add it to other places (i.e. bumping into race cars)

We should also integrate SDL's Haptic API, however since it is not supported on Emscripten I have no device available that supports this API. Someone with iOS/iPad/NDS could give it a shot.

### DIFF
--- a/ISLE/isleapp.cpp
+++ b/ISLE/isleapp.cpp
@@ -791,6 +791,11 @@ SDL_AppResult SDL_AppEvent(void* appstate, SDL_Event* event)
 			SDL_Log("Game started");
 		}
 	}
+	else if (event->user.type == g_legoSdlEvents.m_hitActor) {
+		if (InputManager()) {
+			InputManager()->HandleRumbleEvent();
+		}
+	}
 
 	return SDL_APP_CONTINUE;
 }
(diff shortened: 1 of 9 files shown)


VoxelTek:
@foxtacles Should I add this as an option to the config tool? Additionally, I've been thinking, should I add a setting to select *what* controller to use in the config tool? 

EDIT: Should I do the same for touch control? I can imagine there are some users on, say, a Surface Pro, that may want to play with their computer's touch screen....for some reason.

foxtacles:
> @foxtacles Should I add this as an option to the config tool? Additionally, I've been thinking, should I add a setting to select _what_ controller to use in the config tool?
> 
> EDIT: Should I do the same for touch control? I can imagine there are some users on, say, a Surface Pro, that may want to play with their computer's touch screen....for some reason.

Adding haptic setting: yes please 

I'd like to avoid settings that require you to select a specific input device. I'd much prefer if all connected devices can automatically be used for control without having to pick them individually. I don't really see a reason to limit it to a single joystick for instance. If you don't want to use one of them...just don't use it

VoxelTek:
> I'd like to avoid settings that require you to select a specific input device. I'd much prefer if all connected devices can automatically be used for control without having to pick them individually. I don't really see a reason to limit it to a single joystick for instance. If you don't want to use one of them...just don't use it

Sounds good, I'll do that.

AJenbo:
What if a connect Device had drift or is a virtual device that activates on keyboard input

PR #598, opened 2025-07-14
Add touch control options, add rumble checkbox, clean up layout.

VoxelTek:
Screenshot_20250714_141112

- Combo box added for changing Touch Control Scheme
- Checkbox added for rumble toggle
- Transition type combo box cleaned up
- `Low` Model Quality marked as broken
- Misc other changes and fixes


foxtacles:
Maybe you can add a quick check box for activating texture loader? No need to go all the way with a generic extension abstraction, but would be good to have for now.

VoxelTek:
> Maybe you can add a quick check box for activating texture loader? No need to go all the way with a generic extension abstraction, but would be good to have for now.

Yeah, ok, lemme figure out the best place to put it

PR #604, opened 2025-07-14
Fix no black border on virtual cursor

Helloyunho:
TL;DR
CleanShot 2025-07-14 at 23 53 08@2x
to
Configure LEGO Island 2025-07-14 23 50 57

With #595, the black border on virtual cursor disappears, which makes it hard to see on white background.
This pr fixes that by setting different color to key-out(idk that's the right term)


AJenbo:
This will create a purple halo around the mouse on low resolution devices, instead color keying should only be set if alpha is not enabled, if alpha is enabled transparent black should be used.

PR #607, opened 2025-07-14
Integrate SDL Haptic API

### LINKED ISSUE #597: Integrate SDL Haptic API
https://github.com/isledecomp/isle-portable/pull/596

Using the Haptic API will allow us to also have rumble on phones etc

foxtacles:
Close #597 

This PR does a few things:

* Integrate SDL Haptic API
* Track all connected joysticks, mice and haptic devices in `LegoInputManager`
* Track last used input device. Rumble (and possibly other effects in the future) will be run on the last used input device only.
* Remove `m_useJoystick` and `m_joystickIndex` and corresponding `.ini` settings; instead, allow any number of connected devices to be used

Unfortunately I don't have a haptic device I can test this with. My gamepad (Xbox 360 controller) doesn't register as a haptic device. If anyone could test this with a haptic device that would be great (i.e. iOS port should probably work; Android port if we had it; or any physical haptic device connected)

The way to test this would be to run into a Lego minifigure and it should play a rumble effect.


Helloyunho:
No haptic feedback on iOS...

foxtacles:
> No haptic feedback on iOS...

Looks like SDL doesn't support iOS after all https://github.com/libsdl-org/SDL/tree/main/src/haptic

Issue #578, opened 2025-07-11
Aspect Ratio selection


AdminProductions:
I recently downloaded the Windows version of isle-portable and noticed that the fullscreen mode makes the game run at 16:9 (I have a 16:9 monitor).
This is also true when running the game in windowed mode and maximizing the window.
While this looks good when running around the island, transition effects and 2D images are still locked to 4:3.
Would it be possible to implement an aspect ratio toggle?

foxtacles:
I was actually about to implement this for the Web port. However, having it a platform-independent option would certainly be better, and also save me from using a bunch of Emscripten-specific stuff.

@AJenbo is that something you could work on, how much effort would it be? Seems easy enough to calculate the target portion of the window for rendering the game based on the 4:3 aspect ratio, but I'm not sure how to make SDL change its frame of reference to the rendered part of the window instead of all of it (otherwise (0, 0) starts at the black bar, and not at the game, requiring patching up all coordinate events)

Seems like `SDL_RenderSetLogicalSize` would do the trick but only when having a `SDL_Renderer`, which we do not in most backends. So only way is manual transformation of event values? CC @madebr

AJenbo:
> This is also true when running the game in windowed mode and maximizing the window.

Not quite since UI will alter the aspect a bit. You can resize the window to any aspect you like.

> I'm not sure how to make SDL change its frame of reference to the rendered part of the window instead of all of it (otherwise (0, 0) starts at the black bar, and not at the game, requiring patching up all coordinate events)

Minwin already calculates the offset and scale for the 4:3 game context and adjust sdl_events accordingly, that's why it works with the high resolution window provided by isle.

> Seems like `SDL_RenderSetLogicalSize` would do the trick but only when having a `SDL_Renderer`, which we do not in most backends. So only way is manual transformation of event values?

This is already done by miniwin.

> how much effort would it be?

I still suggest doing this by implementing exclusive Fullscreen mode as an option. It should probably not be the default since some systems can't do 640x480, but you could additionally provide a resolution option so they can pick another 4:3 resolution if available.

This would also satisfy the players that prefer 4:3 stretched to 16:9 (I heard Lego island players specifically say they prefer this) as they can set the monitor or GPU to stretch the resolution the same way they are used to.

And it can help performance for systemets with weak GPUs as they can run the game at less than native resolution.

Lastly this would also fix Fullscreen for dx5.

> [@AJenbo](https://github.com/AJenbo) is that something you could work on

[...]

PR #608, opened 2025-07-15
Rework of isle-config, add resolution adjustment and max framerate settings

VoxelTek:
image
image

- Resolution adjustment has been added. Allows for setting the initial startup resolution in windowed mode.
- Max Framerate setting has been added. Allows for customising the maximum framerate to be other than the default maximum of 100fps.
- `isle-config` has been reworked to be a smaller window, with each category of settings getting a different tab.

## NOTES:
- The Aspect Ratio setting, when set to either `Standard`, `Widescreen` or `Square`, simply means that the horizontal resolution cannot be manually changed, and is instead relative to the vertical resolution. The Aspect Ratio setting has no purpose beyond this, and does not influence the game itself in any way.
- While maximum framerates below 100fps have been tested and confirmed to work, potentially raising the limit above 100fps has not been tested, and may cause unexpected results. 

Due to the introduction of tabs to the UI, perhaps it might be best to re-enable resizing of the `isle-config` window?


ISLE/isleapp.h
```diff
@@ -108,6 +108,8 @@ class IsleApp {
 	MxTransitionManager::TransitionType m_transitionType;
 	LegoInputManager::TouchScheme m_touchScheme;
 	MxBool m_haptic;
+	int m_xRes;
+	int m_yRes;
```
foxtacles: let's use MxS32

ISLE/isleapp.cpp
```diff
@@ -124,6 +124,8 @@ extern const char* g_files[46];
 // FUNCTION: ISLE 0x401000
 IsleApp::IsleApp()
 {
+	m_xRes = 640;
+	m_yRes = 480;
```
foxtacles: Add these at the end of the function

ISLE/isleapp.h
VoxelTek: Done

ISLE/isleapp.cpp
VoxelTek: Done

AJenbo:
Could you add an option for exclusive full screen?

When checked it should also call: https://wiki.libsdl.org/SDL3/SDL_SetWindowFullscreenMode

(you can use https://wiki.libsdl.org/SDL3/SDL_GetFullscreenDisplayModes to find the closets valid option to what has been entered)

This could take care of https://github.com/isledecomp/isle-portable/issues/578

VoxelTek:
Hmm, in that case, I might rework the `Display` tab, gimme a sec.

VoxelTek:
well uhh, probably shouldn't merge, something's gone wrong somewhere cause
image

For some reason, it seems setting the window resolution to something other than 640x480 in fullscreen mode breaks the game. Odd that it doesn't happen for windowed, @AJenbo any idea why?

In the meantime, I'll keep working on the fullscreen thing, and see if I can fix this issue I've caused.

AJenbo:
> For some reason, it seems setting the window resolution to something other than 640x480 in fullscreen mode breaks the game. Odd that it doesn't happen for windowed, @AJenbo any idea why?

It uses the initial size as the virtual canvas size for 2D rendering:

https://github.com/isledecomp/isle-portable/blob/19edb0340eac58bc9647f178f49f464d3310d434/miniwin/src/ddraw/ddraw.cpp#L304

https://github.com/isledecomp/isle-portable/blob/19edb0340eac58bc9647f178f49f464d3310d434/miniwin/src/ddraw/ddraw.cpp#L274

https://github.com/isledecomp/isle-portable/blob/19edb0340eac58bc9647f178f49f464d3310d434/miniwin/src/d3drm/d3drm.cpp#L141

So if you initialize the renderer with a size of 1280x768 then rendering 2D content at a size of 640x480 won't take up the full screen.

ISLE/isleapp.cpp
```diff
@@ -839,6 +840,17 @@ MxResult IsleApp::SetupWindow()
...
 
 	window = SDL_CreateWindowWithProperties(props);
+
+	if (m_fullScreen) {
+		SDL_DisplayMode closestMode;
+		SDL_DisplayID displayID = SDL_GetDisplayForWindow(window);
+		bool findModeSuccess =
+			SDL_GetClosestFullscreenDisplayMode(displayID, m_xRes, m_yRes, m_frameRate, true, &closestMode);
+		if (findModeSuccess) {
+			bool setModeSuccess = SDL_SetWindowFullscreenMode(window, &closestMode);
```
AJenbo: This should only be called if the user chose exclusive fullscreen, not when it's the default border lesss fullscreen

VoxelTek: Ah, right, you wanted a setting for that. Would the term "Exclusive Fullscreen" work?

AJenbo: Yeah that sounds good

ISLE/isleapp.cpp
```diff
@@ -1085,6 +1102,13 @@ bool IsleApp::LoadConfig()
 		(MxTransitionManager::TransitionType) iniparser_getint(dict, "isle:Transition Type", m_transitionType);
 	m_touchScheme = (LegoInputManager::TouchScheme) iniparser_getint(dict, "isle:Touch Scheme", m_touchScheme);
 	m_haptic = iniparser_getboolean(dict, "isle:Haptic", m_haptic);
+	m_xRes = iniparser_getint(dict, "isle:Horizontal Resolution", m_xRes);
+	m_yRes = iniparser_getint(dict, "isle:Vertical Resolution", m_yRes);
+	if (!m_fullScreen) { // 2D elements break otherwise, not sure why
```
AJenbo: When fullscreen is enabled isle will call SetDisplayMode on DirectX, if not it will derive the resolutionfrom the window size.

VoxelTek: I'm...not quite sure what you're saying. Could you rephrase?

VoxelTek:
> Could you add an option for exclusive full screen?
> 
> When checked it should also call: https://wiki.libsdl.org/SDL3/SDL_SetWindowFullscreenMode
> 
> (you can use https://wiki.libsdl.org/SDL3/SDL_GetFullscreenDisplayModes to find the closets valid option to what has been entered)
> 
> This could take care of #578

Turns out, there was a more suitable function, `SDL_GetClosestFullscreenDisplayMode`. I've used that, and it seems to be working well.

I had envisioned simply showing all the possible fullscreen modes in the config tool, and allowing the user to select one, however I ran into a problem in that the `DisplayID` may not be the same between reboots, or even between `isle-config` and `isle` itself. As such, I decided to handle it all in `isleapp.cpp`.

Windowed mode, fullscreen mode, and exclusive fullscreen mode all appear to work now. However, I would like @foxtacles to test out the web build, as this wouldn't be the first time I've broken it in a fun and interesting way. I'd do so myself, but I wasn't able to figure out how to make a custom build instead of just using the official repo.

ISLE/isleapp.cpp
AJenbo: > 2D elements break otherwise, not sure why

Because it only calls `SetDisplayMode()` `if (m_fullScreen)`

VoxelTek: Ah yeah, I understand that bit now, you explained it above. I guess I could change that.

AJenbo:
> I wasn't able to figure out how to make a custom build instead of just using the official repo.

Click the one for your system:

image

Go to the summery page, and download the one for your system at the bottom of the page

VoxelTek:
> > I wasn't able to figure out how to make a custom build instead of just using the official repo.
> 
> Click the one for your system:

No I mean the Emscripten (web) version. While I understand it compiles as part of the CI, I could never get it to work for me. I might look into it more this afternoon.

foxtacles (approved):
Looks good to me. Web port uses its own fullscreen logic so it isn't affected by these changes

VoxelTek:
> Web port uses its own fullscreen logic so it isn't affected by these changes

Fair, but I did once break it cause I made windowed mode resizable.

AJenbo:
Loooks like you just have to resolve a merge conflict now

AJenbo (approved):

PR #616, opened 2025-07-16
Use actual transparent pixel if surface supports alpha channel

Helloyunho:
follows [this comment](https://github.com/isledecomp/isle-portable/pull/604#issuecomment-3071263389)


AJenbo (approved):
Nice work

Issue #119, opened 2025-05-19
~ViewLODList(): Assertion `m_refCount == 0'


AJenbo:
This happens when shutting down the game. It's hard to know if the remaining leakage issues are related to it or not so probably a good place to start if anyone want to try and fix memory leakage.

```
isle: isle-portable/LEGO1/viewmanager/viewlodlist.h:209: virtual ViewLODList::~ViewLODList(): Assertion `m_refCount == 0' failed.
Aborted (core dumped)

```

foxtacles:
(happens in decomp too)

MaddTheSane:
Someone probably did a `delete aViewLODList;` instead of a `aViewLODList->Release()` (or equivalent). I will admit to being more knowledgeable of Objective-C, which uses reference counting on all non-singleton objects and direct calls to the destructor is considered a bad idea. Other than compiling for release (and thus ignoring the assert), fixing it should be trivial.

I know this project is currently trying to match the behavior of the original Lego Island as close as possible, so maybe this could be "fixed" in the next stage of the project?

AJenbo:
I really hope this project will fix bad stuff like that. We did already fix other bad dev assumptions in the code.

foxtacles:
We can fix bugs that aren't game altering in portable, but before doing so we should understand the complete nature of them. The assertion in this particular case seems reasonable, so we should figure out where the actual bug is and resolve that. (instead of just removing the assert, for instance)

foxtacles:
There appears to be another leak related to `LegoROI`. `LegoROI` adds a reference to a `ViewLODList`, and in certain circumstances (i.e. after having built the racecar, going back to the Infocenter, then existing through the green brick) a `LegoROI` is leaked somewhere which leads to a `ViewLODList` reference leak, which leads to this assertion.

Unfortunately the ownership model of `LegoROI` is very complicated so it's not obvious which component isn't cleaning up properly. Tracking this will require more digging and tracing of everything where `LegoROI*` is passed into (i.e. all `LegoEntity` instances)

Wintermute21:
this is from the latest master branch as of: 0c7a95f14377be205944ce358f7e508dec3c566b
`Assertion failed: (m_refCount == 0), function ~ViewLODList, file viewlodlist.h, line 211.`
on ARM macOS 26.3

calmsacibis995:
[...]

PR #619, opened 2025-07-17
properly release lodlist

olebeck:
fixes the ref count assert #119

### DIFF
--- a/LEGO1/viewmanager/viewlodlist.cpp
+++ b/LEGO1/viewmanager/viewlodlist.cpp
@@ -50,14 +50,10 @@ ViewLODListManager::~ViewLODListManager()
 			delete const_cast<ViewLOD*>(pLODList->PopBack());
 		}
 
-		delete pLODList;
-		// ??? for now
-		delete[] const_cast<char*>(rROIName);
+		pLODList->Release();
 	}
 
-	// ??? correct way of "emptying" map
-	m_map.erase(m_map.begin(), m_map.end());
-
+	m_map.clear();
 	assert(m_map.begin() == m_map.end());
 }
 


LEGO1/viewmanager/viewlodlist.cpp
```diff
@@ -50,14 +50,10 @@ ViewLODListManager::~ViewLODListManager()
 			delete const_cast<ViewLOD*>(pLODList->PopBack());
 		}
 
-		delete pLODList;
-		// ??? for now
-		delete[] const_cast<char*>(rROIName);
+		pLODList->Release();
```
foxtacles, line 53: This seems incorrect.

Release calls Destroy once the refCount reaches zero: https://github.com/isledecomp/isle-portable/blob/6683288c77114579d0e610023ffeaa62b01f2792/LEGO1/viewmanager/viewlodlist.cpp#L124 which will iterate over the list again, and remove the element. This will lead to iterator corruption since it invalidates the local iterator.

And if that worked, the map would be empty and clear would no longer be required after the loop. You should add an assertion to make sure the map is empty. If it is not, there is a resource leak somewhere else that is being masked by this change. If there is a second manager instance that shares the same LOD lists this might not be the case, but we should still make sure that no resources are left behind.

foxtacles: In other words, the current original implementation of the destructor is generally correct, given the precondition that all LODs have a refCount of exactly 0.  The manager should not be deleted unless that is the case, and the current assertion guards against that.  This means one or more calls to `Release()` are likely missing somewhere in the code base.

foxtacles:
Fixed it here: https://github.com/isledecomp/isle-portable/pull/623

PR #628, opened 2025-07-19
Fix bad merge

AJenbo:


foxtacles:
Hm, looks like we are not building the debug in the CI anymore. We probably should

PR #636, opened 2025-07-22
Upgrade OpenGL ES renderer to 3.0, add option for MSAA

foxtacles:
Works on WebGL, I don't have any other device where I could test this on.

I've added a mechanism to allow passing new rendering options, like MSAA, to miniwin. I've used the same DirectX interface approach that we've been using for the d3drm miniwin device, so this should work fine with DirectX 5 as well.

You can request MSAA in the `.ini`:

```
[isle]
MSAA=4
```


jpeisach:
This might sound ridiculous but is it possible we could support both ES2 and ES3 for platforms later down the line that may not have ES3?

AJenbo:
It's totally possible, just dig up the old version, but lets not do so before we have something that needs ES2. Android would be the first thing that comes to mind for ES but all version supported by the current SDK have ES3.

AJenbo:
This doesn't run at all for me.
```
isle-config: LEGO1/mxdirectx/mxdirectxinfo.cpp:236: BOOL MxDeviceEnumerate::EnumDirectDrawCallback(LPGUID, LPSTR, LPSTR): Assertion `videoParam' failed.
```

miniwin/src/d3drm/backends/opengles3/renderer.cpp
```diff
@@ -37,9 +37,14 @@ Direct3DRMRenderer* OpenGLES2Renderer::Create(DWORD width, DWORD height)
...
 	SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_ES);
-	SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 2);
+	SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
 	SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 0);
 
+	if (msaaSamples > 1) {
+		SDL_GL_SetAttribute(SDL_GL_MULTISAMPLEBUFFERS, 1);
+		SDL_GL_SetAttribute(SDL_GL_MULTISAMPLESAMPLES, msaaSamples);
+	}
+
```
AJenbo: I think this is only relevant if you where rendering directly to the framebuffer but this backend uses indirect rendering via FBO.

miniwin/src/d3drm/backends/opengles3/renderer.cpp
```diff
@@ -610,6 +625,52 @@ void OpenGLES2Renderer::Resize(int width, int height, const ViewportTransform& v
...
+	}
+	m_colorTarget = m_depthTarget = m_msaaColorRbo = m_msaaDepthRbo = 0;
+
+	GLint samples, maxSamples;
+	glGetIntegerv(GL_SAMPLES, &samples);
+	glGetIntegerv(GL_MAX_SAMPLES, &maxSamples);
+	m_useMsaa = samples > 1;
+
+	if (m_useMsaa) {
+		glBindFramebuffer(GL_FRAMEBUFFER, m_msaaFbo);
```
AJenbo: rather then having an FBO for msaa and one for non-msaa it would make more sens to have one FBO that is either MSAA or not, you won't ever use both at the same time so it's just taking up extra memory to have both.

foxtacles:
> This doesn't run at all for me.
> 
> ```
> isle-config: LEGO1/mxdirectx/mxdirectxinfo.cpp:236: BOOL MxDeviceEnumerate::EnumDirectDrawCallback(LPGUID, LPSTR, LPSTR): Assertion `videoParam' failed.
> ```

This should just be the config app. I haven't tested that since I'm not using it. I'll make an exception for it

AJenbo:
I guess we have to ask @VoxelTek if they can add this option to the config then.

AJenbo:
> This should just be the config app.

It also fails to run by the way so making an exception for this in config won't get it working.

foxtacles:
> > This should just be the config app.
> 
> It also fails to run by the way so making an exception for this in config won't get it working.

That's strange, works fine here

AJenbo:
But your only testing via WebGL 2.0 right, not ES 3.0?

foxtacles:
> But your only testing via WebGL 2.0 right, not ES 3.0?

Yeah but the error you shared has nothing to do with the renderer.

AJenbo:
It was the only one I got so fare :)

I see that it's actually not building the ES3 backend for me, I wonder what that's about.

miniwin/CMakeLists.txt
```diff
@@ -44,16 +44,16 @@ if(NOT WINDOWS_STORE)
...
   endif()
 
-  find_library(OPENGL_ES2_LIBRARY NAMES GLESv2)
-  if(EMSCRIPTEN OR OPENGL_ES2_LIBRARY)
-    message(STATUS "Found OpenGL: enabling OpenGL ES 2.x renderer")
-    target_sources(miniwin PRIVATE src/d3drm/backends/opengles2/renderer.cpp)
-    list(APPEND GRAPHICS_BACKENDS USE_OPENGLES2)
-    if(OPENGL_ES2_LIBRARY)
-      target_link_libraries(miniwin PRIVATE ${OPENGL_ES2_LIBRARY})
+  find_library(OPENGL_ES3_LIBRARY NAMES GLESv3)
```
AJenbo: This should fix it not building on Linux, I know it seems a bit odd ...
```suggestion
  find_library(OPENGL_ES3_LIBRARY NAMES GLESv2)
```

miniwin/src/d3drm/backends/opengles3/renderer.cpp
foxtacles: Yeah, makes sense. Somehow I was under the impression I'd need a separate FBO just for MSAA but doesn't look like it. Do you want to make that change?

miniwin/src/d3drm/backends/opengles3/renderer.cpp
foxtacles: Hmm, I think I do needed to set these for WebGL but let me try if it works without.

LEGO1/mxdirectx/mxdirectxinfo.cpp
```diff
@@ -216,12 +219,24 @@ BOOL MxDeviceEnumerate::EnumDirectDrawCallback(LPGUID p_guid, LPSTR p_driverDesc
...
 	}
 
+	result = lpDD->QueryInterface(IID_IDirect3DMiniwin, (void**) &miniwind3d);
+	if (result == DD_OK) {
+		MxVideoParam* videoParam = (MxVideoParam*) SDL_GetPointerProperty(
+			SDL_GetWindowProperties(reinterpret_cast<SDL_Window*>(m_hWnd)),
+			ISLE_PROP_WINDOW_CREATE_VIDEO_PARAM,
+			nullptr
+		);
+		assert(videoParam);
```
AJenbo, line 237: Fails for config, maybe:
```suggestion
		if (videoParam);
```

VoxelTek:
> I guess we have to ask @VoxelTek if they can add this option to the config then.

Hi yes

So the video driver list isn't a preset list or anything, it should automatically get all the available options. The file throwing an error is not one directly part of `isle-config` I believe, so I'm suspecting that there's a bug somewhere. I'll try and find and fix it if I can.

EDIT: I may be stupid, I think I may have misunderstood your message. I'll add MSAA support to `isle-config`.

miniwin/src/d3drm/backends/opengles3/renderer.cpp
AJenbo: Yeah I got it working now so I can start cleaning things up a bit and upgrading the shaders to make use of ES 3 as well.

AJenbo:
@VoxelTek i was more wondering if you would be willing to add UI for setting the new ini option for MSAA:

[isle]
MSAA                           = "8"

1 is off, 8 is pretty high, but I thin some hardware can do even 16. It is possible to query for the max, but I don't think that function has been exposed and it would be complicated to gather it from all renderers.

VoxelTek:
> This doesn't run at all for me.
> 
> ```
> isle-config: LEGO1/mxdirectx/mxdirectxinfo.cpp:236: BOOL MxDeviceEnumerate::EnumDirectDrawCallback(LPGUID, LPSTR, LPSTR): Assertion `videoParam' failed.
> ```

Kinda same here, trying to run `isle-config` throws this assertion error, and even when I change the assert to an `if` (as suggested by you above) it segfaults. So....



> @VoxelTek i was more wondering if you would be willing to add UI for setting the new ini option for MSAA:
> 
> [isle] MSAA = "8"
> 
> 1 is off, 8 is pretty high, but I thin some hardware can do even 16. It is possible to query for the max, but I don't think that function has been exposed and it would be complicated to gather it from all renderers.

yea ok so I did misunderstand, I'll get to work on this while also trying to understand why in the world it's segfaulting :(

VoxelTek:
@foxtacles @AJenbo To confirm, are the possible values for `MSAA` 1 through to 8, (e.g. 1, 2, 3, 4) or the doubles (e.g. 1, 2, 4, 8)?

LEGO1/mxdirectx/mxdirectxinfo.cpp
foxtacles: I think I'd like to make it required for the main app so that when someone accidentally forgets to set the property (or it breaks in some other way) we notice it, otherwise the new settings just will be quietly ignored.

For the config app it probably won't matter so excluding it just for the config app seems fine. Or we could just add the property to the window creation of the config app as well

AJenbo:
1,2,4,8,16

VoxelTek:
Right, that makes things somewhat harder, but not impossible

miniwin/src/d3drm/backends/opengles3/renderer.cpp
AJenbo: Arh I see you use it to store the configured samples and then read it out later using glGetIntegerv. A bit over kill maybe :)

LEGO1/mxdirectx/mxdirectxinfo.cpp
foxtacles: Should be working now for config

VoxelTek:
So, wait, is the max `8` or `16`? When `isle-config` starts, it prints this to the command line:
`MSAA is off. Requested samples: 0, active samples: 0, max samples: 8`

miniwin/src/d3drm/backends/opengles3/renderer.cpp
foxtacles: For WebGL I needed to set it otherwise MSAA would simply not work at all. Might be different for other platforms

foxtacles:
> So, wait, is the max `8` or `16`? When `isle-config` starts, it prints this to the command line: `MSAA is off. Requested samples: 0, active samples: 0, max samples: 8`

The max is whatever the renderer supports on any given the system. The log that you are seeing is from the OpenGL ES test window, you can ignore it for config. Querying the max value for each renderer would be a lot of work so I'd just have a list of default values as @AJenbo suggested. These can be framed as "requests", and won't be guaranteed to work if set

VoxelTek:
Right, I'll have it be 16 as the max.

AJenbo:
I just adjusted things so that even if set to more then supported the render will just use what ever it's max supported value is. This matches how we handle texture resolutions and some other options.

> So, wait, is the max 8 or 16?

Depends on your graphics card.

miniwin/src/d3drm/backends/opengles3/renderer.cpp
AJenbo: Can you please test to see if it works now.

miniwin/src/d3drm/backends/opengles3/renderer.cpp
```diff
@@ -1,8 +1,8 @@
-#include "d3drmrenderer_opengles2.h"
+#include "d3drmrenderer_opengles3.h"
 #include "meshutils.h"
 
-#include <GLES2/gl2.h>
 #include <GLES2/gl2ext.h>
```
AJenbo, line 4:
```suggestion
```

miniwin/src/d3drm/backends/opengles3/renderer.cpp
```diff
@@ -626,67 +622,36 @@ void OpenGLES3Renderer::Resize(int width, int height, const ViewportTransform& v
...
-		glRenderbufferStorageMultisample(GL_RENDERBUFFER, samples, GL_DEPTH24_STENCIL8, width, height);
-		glFramebufferRenderbuffer(GL_FRAMEBUFFER, GL_DEPTH_ATTACHMENT, GL_RENDERBUFFER, m_msaaDepthRbo);
-
-		if (glCheckFramebufferStatus(GL_FRAMEBUFFER) != GL_FRAMEBUFFER_COMPLETE) {
-			SDL_Log("MSAA Framebuffer is not complete! Disabling MSAA.");
-			m_useMsaa = false;
-		}
-	}
-
-	SDL_Log(
```
foxtacles: Can we keep some kind of logging? Otherwise it's hard to tell what is active and what isn't

foxtacles:
Transitions seem to be broken on WebGL now (mosaic)

miniwin/src/d3drm/backends/opengles3/renderer.cpp
foxtacles: Can't remove it because `GL_TEXTURE_MAX_ANISOTROPY_EXT` is defined in it

Couldn't find a v3 header that has it

AJenbo: yeah your right :)

miniwin/src/d3drm/backends/opengles3/renderer.cpp
foxtacles: MSAA seems to work, transitions are broken. But let's add back some logging please that tells us how MSAA is active or not, I don't like squinting at my screen to see what's going on

miniwin/src/d3drm/backends/opengles3/renderer.cpp
AJenbo: Done

AJenbo:
> Transitions seem to be broken on WebGL now (mosaic)

Same on ES 3.0, I'll go over the changes and see if I can find where there is a miss match.

VoxelTek:
Implemented `MSAA` stuff into the config tool, along with some of the Exclusive Fullscreen resolution options from a couple days ago. Will probably make a PR shortly.

[MSAA.webm](https://github.com/user-attachments/assets/016ea973-88c5-4eb0-b9f7-d7628b88f22f)

foxtacles:
Works on WebGL

AJenbo:
@foxtacles things looks good now in terms of MSAA, but there are still a few things that are decisively ES 2.0 and not ES 3.0 (like the shades) which I would like to correct as well.

foxtacles:
> @foxtacles things looks good now in terms of MSAA, but there are still a few things that are decisively ES 2.0 and not ES 3.0 (like the shades) which I would like to correct as well.

Sure, feel free to merge whenever you think it's ready

AJenbo:
@foxtacles if things work on your end then I think this is ready to merge.

foxtacles:
Getting an error now on WebGL:

> Uncaught TypeError: Failed to execute 'attachShader' on 'WebGL2RenderingContext': parameter 2 is not of type 'WebGLShader'.

No more info than this unfortunately

Also this:

`CompileShader (No supported SDL_GPU backend found!)`

AJenbo:
> CompileShader (No supported SDL_GPU backend found!)

that's just the SDL_GPU backend detecting that it's not supported by Emscripten. Not new.

foxtacles:
> > CompileShader (No supported SDL_GPU backend found!)
> 
> that's just the SDL_GPU backend detecting that it's not supported by Emscripten. Not new.

I have not seen it before. This is what isn't new:

> ERROR: SDL_CreateGPUDevice failed (No supported SDL_GPU backend found!)

Those extra lines with `CompileShader` below it are new

image

AJenbo:
> No more info than this unfortunately

I added proper logging for shader compilation errors, please check now.

foxtacles:
image

AJenbo:
> I have not seen it before. This is what isn't new:

Arh it's because it's the last error from SDL and the compile error isn't an SDL error so it's just printing the same message in the wrong context.

AJenbo:
> image

Can you get me a text version instead ...

foxtacles:
```
isle.js:1 Found 21 game files in /bundled
isle.js:1 Error enumerating files for path / (Can't open directory: No such file or directory)
isle.js:1 ERROR: HDC GetDC(HWND): Not implemented
isle.js:1 ERROR: int ReleaseDC(HWND, HDC): Not implemented
isle.js:1 ERROR: SDL_CreateGPUDevice failed (No supported SDL_GPU backend found!)
isle.js:1 Shader compile error: ERROR: 0:2: '
isle.js:1 ' : #version directive must occur on the first line of the shader
isle.js:1 ERROR: 0:5: 'in' : storage qualifier supported in GLSL ES 3.00 and above only
isle.js:1 ERROR: 0:6: 'in' : storage qualifier supported in GLSL ES 3.00 and above only
isle.js:1 ERROR: 0:7: 'in' : storage qualifier supported in GLSL ES 3.00 and above only
isle.js:1 ERROR: 0:13: 'out' : storage qualifier supported in GLSL ES 3.00 and above only
isle.js:1 ERROR: 0:14: 'out' : storage qualifier supported in GLSL ES 3.00 and above only

isle.js:1 ERROR: 0:15: 'out' : storage qualifier supported in GLSL ES 3.00 and above only
isle.js:1 Shader compile error: ERROR: 0:2: '
isle.js:1 ' : #version directive must occur on the first line of the shader
isle.js:1 ERROR: 0:14: 'in' : storage qualifier supported in GLSL ES 3.00 and above only
isle.js:1 ERROR: 0:15: 'in' : storage qualifier supported in GLSL ES 3.00 and above only
isle.js:1 ERROR: 0:16: 'in' : storage qualifier supported in GLSL ES 3.00 and above only
isle.js:1 ERROR: 0:23: 'out' : storage qualifier supported in GLSL ES 3.00 and above only
isle.js:1 ERROR: 0:65: 'texture' : no matching overloaded function found
isle.js:1 ERROR: 0:65: '=' : dimension mismatch
isle.js:1 ERROR: 0:65: '=' : cannot convert from 'const mediump float' to 'mediump 4-component vector of float'
isle.js:1 worker sent an error! http://localhost:6931/isle.js:1: Uncaught TypeError: Failed to execute 'attachShader' on 'WebGL2RenderingContext': parameter 2 is not of type 'WebGLShader'.
```

AJenbo:
See if that helped, apparently Emsripten doesn't think the first line is line 1

foxtacles:
> See if that helped, apparently Emsripten doesn't think the first line is line 1

Yes that worked. I guess WebGL is technically correct here, since the first line was `(` followed by a new line character

Good to merge from my end

VoxelTek:
Not sure if this is known, but `SDL GPU` doesn't appear to work with MSAA. The game starts and plays just fine, but there's no evidence of the effects.

OpenGL ES 3.0:
image


SDL GPU (likely Vulkan):
image


While it is somewhat subtle, the lack of MSAA is present and visible, especially when actually playing the game.

foxtacles:
https://github.com/isledecomp/isle-portable/issues/639

VoxelTek:
Oops, my mistake.

PR #641, opened 2025-07-23
Make Anisotropic filtering configurable

foxtacles:
Since we have the ability now, I've made AF an option similar to MSAA.

```
[isle]
Anisotropic=16
```

### DIFF
--- a/ISLE/isleapp.cpp
+++ b/ISLE/isleapp.cpp
@@ -183,6 +183,7 @@ IsleApp::IsleApp()
 	m_frameRate = 100.0f;
 	m_exclusiveFullScreen = FALSE;
 	m_msaaSamples = 0;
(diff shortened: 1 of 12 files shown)


AJenbo (approved):

foxtacles:
CC @VoxelTek another option to add to config. Values can be same as MSAA

VoxelTek:
I'll get to work on that shortly

VoxelTek:
Regarding MSAA and anisotropy, should the slider labels display their actual values (1, 2, 4, 8, 16) or a "level" (0, 1, 2, 3, 4)? I can imagine it would be a tad confusing, as users might look for a way to turn these "off", not realising that `1` is off.

VoxelTek:
Also, it'd be helpful if you could provide some tooltips to use.

Issue #642, opened 2025-07-23
(Web port) OpenGL ES 3.0 causes glitches (Chrome Android)


foxtacles:
Looks like on mobile devices (at least Chrome Android) the new ES 3.0 renderer causes some glitches. I had to revert the update on isle.pizza for now

This happens independent of MSAA, regardless of settings

Can only reproduce on mobile 

@AJenbo any ideas? I remember we had that shader issue in the past regarding binding a texture, could it be anything of that sort?

https://drive.google.com/file/d/1LHiCOflEm-9gRWakpZBlR42zZeqK7D6u/view?usp=drivesdk

foxtacles:
Identifier one of the issues: we need to use `highp` precision in the vertex shader.

The other glitch (briefly screen goes black / is cleared during transitions) is still unclear. It seems to happen pseudo-randomly only

foxtacles:
> The other glitch (briefly screen goes black / is cleared during transitions) is still unclear. It seems to happen pseudo-randomly only

I can fix this by reverting to the old Flip logic of the ES 2.0 renderer which draws to a quad.

~I can also fix it by having another "download" intermediate FBO for Download~

@AJenbo any ideas how we could fix this without reverting the logic?

This is what I have as a patch that fixes the problem: [renderer.patch](https://github.com/user-attachments/files/21397274/renderer.patch) but it basically reverts to the ES 2.0 texture render flip logic, still hoping for a better fix.

foxtacles:
It looks like this also fixes the flickering issue

<img width="511" height="459" alt="Image" src="https://github.com/user-attachments/assets/b083eed6-40bb-4611-8cfd-7476ac0f5d8b" />

`glMemoryBarrier(GL_FRAMEBUFFER_BARRIER_BIT)` isn't available in ES 3.0 unfortunately

foxtacles:
After some messing around, looks like this fixes the issue as well (keeping the current `Flip`):

(basically any physical write to the back buffer will do, but the following example should be minimally performance impacting - maybe there's better ways)

<img width="668" height="610" alt="Image" src="https://github.com/user-attachments/assets/716e175d-3daf-43a5-980f-bd90d032a466" />

I think this might be a driver bug

AJenbo:
Maybe add it with a comment

foxtacles:
I'm gonna do some more debugging with glGetError but if that yields nothing I will open a PR

AJenbo:
[...]

PR #643, opened 2025-07-23
Fix insufficient float precision

foxtacles:
On mobile WebGL systems, this causes textures to glitch, see #642


AJenbo:
Makes sense, I thought there was a chance of this but didn't see any issues when testing

PR #653, opened 2025-07-31
OG Xbox port

VoxelTek:
- [x] CMake'd
- [x] Get `miniaudio` working ([miniaudio#1023](https://github.com/mackron/miniaudio/issues/1023)) (now compiles, unsure if it works)
- [x] Builds
- [ ] Boots to FMVs
- [ ] Boots to game
- [ ] Runs on real hardware
- [ ] Add CI building

### DIFF
--- a/ISLE/xbox/dbg.c
+++ b/ISLE/xbox/dbg.c
@@ -3,5 +3,5 @@
 
 
 void my_output(void *userdata, int category, SDL_LogPriority priority, const char *message) {
-    DbgPrint("%s", message);
+    DbgPrint("%s\n", message);
 }


AJenbo:
@VoxelTek do you expect to be able to move forward with this port? You would likely have to write a new renderer for it since it's not able to use any of the existing ones and it's CPU isn't particularly good.

VoxelTek:
I believe I will have some assistance with writing a rendering backend, however I'm currently concerned about being stuck on some fairly major issues that I'm not sure how to even begin solving.

I do think I could move forward with the port, but only with assistance. If not, this port will likely go stale.

AJenbo:
Then I suggest mentioning the branch in the Wiki instead of having a long term unresolved PR

PR #656, opened 2025-08-05
Android port

Covkie:
Implemented with SDL Android archive (.aar)

Connect to device with MTP and place game data in `Android/data/org.legoisland.isle/files`

- [x] Crash on first launch due to missing isle.ini (race condition?)
- [x] File Icon
- [x] ? signing / apk name

~~- [ ] Support minimum GLES2 with runtime GLES3~~

Touch scheme defaults to Virtual mouse (0)


Covkie:
I'd also like some input regarding a native Android config UI. Could maybe function like the webport where it provides options before starting the SDL activity and returns when the game exits so the app doesn't just quit.

Alternatively a pure SDL(imgui ?) config UI might also be an idea so it's usable by other mobile/console ports

CMakeLists.txt
```diff
@@ -105,12 +107,50 @@ else()
...
+  FetchContent_GetProperties(SDL3)
+
+  set(SDL_ANDROID_ARCHIVE "${CMAKE_SOURCE_DIR}/android-project/app/libs/SDL3.aar")
+  add_custom_command(OUTPUT "${SDL_ANDROID_ARCHIVE}"
+    COMMAND ${Python3_EXECUTABLE} "${SDL3_SOURCE_DIR}/build-scripts/build-release.py"
+       --actions android
+       --fast
+       --android-ndk-home=${ANDROID_NDK_PATH}
+       --force
+       --root="${SDL3_SOURCE_DIR}"
```
madebr: The `build-release.py` script is not really meant to be used like this.
There aren't any promises about argument stability.

CMakeLists.txt
```diff
@@ -105,12 +107,50 @@ else()
...
+
+  add_custom_target(ANDROID_GRADLE_BUILD ALL
+    DEPENDS "${SDL_ANDROID_ARCHIVE}"
+    WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}/android-project"
+    COMMAND "${GRADLEW}" ${GRADLE_TASK} -PcmakeArgs="-DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} -DISLE_USE_DX5=${ISLE_USE_DX5} -DISLE_BUILD_CONFIG=${ISLE_BUILD_CONFIG} -DENABLE_CLANG_TIDY=${ENABLE_CLANG_TIDY} -DISLE_WERROR=${ISLE_WERROR} -DISLE_DEBUG=${ISLE_DEBUG} -Werror=dev"
+    VERBATIM
+    COMMENT "Building Android APK with Gradle"
+  )
+  message(STATUS "SDL AAR")
+  return()
```
madebr: The build procedure is a little perverted. You're actually doing cmake -> gradle -> cmake.
Don't infect CMake with gradle please :)

I'd just require users to https://github.com/libsdl-org/SDL_helloworld/blob/501e31c4d9deac135b63b7535c2220d7d71d4d92/.github/workflows/main.yml#L51-L52 before doing anything isle related.
This can be moved to a separate script, downloading a SDL3 release or building a SDL3 aar.

On ci, you can then just do:
- download, or build SDL3 aar
- run gradle

(and not do apt-get)

CMakeLists.txt
```diff
@@ -54,6 +54,7 @@ cmake_dependent_option(ISLE_COMPILE_SHADERS "Compile shaders" ON "SDL_SHADERCROS
 option(CMAKE_POSITION_INDEPENDENT_CODE "Build with -fPIC" ON)
 option(ENABLE_CLANG_TIDY "Enable clang-tidy")
 option(DOWNLOAD_DEPENDENCIES "Download dependencies" ON)
+set(ANDROID_NDK_PATH "" CACHE PATH "Path to Android NDK")
```
madebr: With my other comment about the general build procedure, I think this can be removed.

CMakeLists.txt
Covkie: Yeah I'm aware it wraps gradle like that, I just found it was the easiest solution to ensure the .aar exists.
Downloading a tagged prebuilt SDL3 aar doesnt exactly work as the rest of the project is built from the latest tag, So I wrote the aar build script integration.

CMakeLists.txt
Covkie: It's used in SDL's actions like this so I figured it was fine. What alternative do you suggest?

madebr: The scripts in `SDL/build-scripts` are internal-only.

I don't know how, but is it possible to let Gradle know that the cmake build step will make available java sources somewhere, which Gradle then can compile?

VoxelTek:
> I'd also like some input regarding a native Android config UI. Could maybe function like the webport where it provides options before starting the SDL activity and returns when the game exits so the app doesn't just quit.
> 
> Alternatively a pure SDL(imgui ?) config UI might also be an idea so it's usable by other mobile/console ports

Might be possible to port the Qt config tool to Android, although I'd imagine that probably wouldn't be a great experience. I think the Qt config tool is still the way to go on desktop, but I also agree that an SDL-based config tool should be written for other platforms. That does raise the issue of then having 3 individual config tools, but the backend ([`config.cpp`](https://github.com/isledecomp/isle-portable/blob/master/CONFIG/config.cpp)) can probably be kept the same between platforms, and perhaps the web version could be moved to this proposed SDL-based config tool.

AJenbo:
feel free to grab the icon and anything else useful from https://github.com/isledecomp/isle-portable/pull/548

CMakeLists.txt
Covkie: There likely is, let me do some research. Probably gonna be some compile step or gradle task shenanigans

Covkie: I've moved the .aar building to gradle but i didn't remove the use of the SDL provided build script. 
Using the script is the only realistic option as SDL has no gradle target to build the aar and recreating its functionality will bring more headaches than not. If the script changes and breaks the build then it can be fixed.

madebr: > I've moved the .aar building to gradle but i didn't remove the use of the SDL provided build script. Using the script is the only realistic option as SDL has no gradle target to build the aar and recreating its functionality will bring more headaches than not. If the script changes and breaks the build then it can be fixed.

That's reasonable and acceptable. SDL's gradle script is lacking in its aar-building capabilities.
It was [capable to create a aar until this PR](https://github.com/libsdl-org/SDL/pull/10450), but the aar was missing features that developers expect from a SDL3 archive, so we removed it.

.github/workflows/ci.yml
```diff
@@ -47,6 +47,8 @@ jobs:
           - { name: 'Emscripten',          os: 'ubuntu-latest',  generator: 'Ninja',                 dx5: false, config: false, emsdk: true, werror: true,   clang-tidy: false,  cmake-wrapper: 'emcmake' }
           - { name: 'Nintendo 3DS',        os: 'ubuntu-latest',  generator: 'Ninja',                 dx5: false, config: false, n3ds: true,  werror: true,   clang-tidy: false,  container: 'devkitpro/devkitarm:latest', cmake-args: '-DCMAKE_TOOLCHAIN_FILE=/opt/devkitpro/cmake/3DS.cmake' }
           - { name: 'Xbox One',            os: 'windows-latest', generator: 'Visual Studio 17 2022', dx5: false, config: false, msvc: true,  werror: false,  clang-tidy: false,  vc-arch: 'amd64', cmake-args: '-DCMAKE_SYSTEM_NAME=WindowsStore -DCMAKE_SYSTEM_VERSION=10.0.26100.0', xbox-one: true}
+          - { name: 'Android',             os: 'ubuntu-latest',  generator: 'Ninja',                 dx5: false, config: false, android: true, werror: true, clang-tidy: false,}
+          - { name: 'Android (Debug)',     os: 'ubuntu-latest',  generator: 'Ninja',                 dx5: false, config: false, android: true, werror: true, clang-tidy: false,  debug: true }
```
madebr: I doubt a debug android build is usable. To debug it, you'd need to know the exact git commit to fetch all sources (isle-portable and SDL3), and set a "substitute-path" so gdb/lldb can map the sources on your system to the sources on the GitHub runner.

.github/workflows/ci.yml
```diff
@@ -81,7 +83,7 @@ jobs:
           chmod a+x /opt/devkitpro/tools/bin/makerom
 
       - name: Install Linux dependencies (apt-get)
-        if: ${{ matrix.linux }}
+        if: ${{ matrix.linux || matrix.android }}
```
madebr: Is this still needed? You're no longer doing cmake -> gradle -> cmake.

.github/workflows/ci.yml
```diff
@@ -106,7 +108,34 @@ jobs:
...
+
+      - name: Build (Android)
+        if: ${{ matrix.android }}
+        run: |
+          cd android-project && \
+          ./gradlew assemble${{ matrix.debug == 'true' && 'Debug' || 'Release' }} \
+            --info \
+            -PcmakeArgs="-DCMAKE_BUILD_TYPE=Release 
+              -DISLE_USE_DX5=${{ !!matrix.dx5 }} \
+              -DISLE_BUILD_CONFIG=${{ !!matrix.config }} \
```
madebr, line 143: Depending on your Android experience and having some motivation, an Android-specific config app would be useful ( :wink:  :wink: )

CMakeLists.txt
```diff
@@ -55,6 +55,7 @@ cmake_dependent_option(ISLE_COMPILE_SHADERS "Compile shaders" ON "SDL_SHADERCROS
 option(CMAKE_POSITION_INDEPENDENT_CODE "Build with -fPIC" ON)
 option(ENABLE_CLANG_TIDY "Enable clang-tidy")
 option(DOWNLOAD_DEPENDENCIES "Download dependencies" ON)
+option(ANDROID_SDL "Download SDL for Android Archive" OFF)
```
madebr: I think all `ANDROID_SDL` changes can be removed?

.github/workflows/ci.yml
Covkie: I included it for the gradle debug target but thats probably not really useful to anyone so i'll remove it

.github/workflows/ci.yml
Covkie: build fails without

CMakeLists.txt
Covkie: it gates to just downloading the SDL3 sources so the aar script can be run

.github/workflows/ci.yml
madebr: What error?
We're installing x11, opengl and qt6 for Linux here, things Android does not need.

madebr: Can you share a link to a failing GitHub action?

.github/workflows/ci.yml
Covkie: I was gonna do a config on the java side but doing it in SDL would make it useful for other mobile ports so ill leave it for someone else

.github/workflows/ci.yml
Covkie: https://github.com/Covkie/isle-portable/actions/runs/16896391022/job/47866899164

madebr: I understand why it fails because of  your initial run with `-DSDL_ANDROID=ON`.
That should be avoided.

madebr: I propose the following:
1) modify `<root>/CMakeLists.txt` so it can use a vendored iniparser (through `FetchContent`) and use a system (pre-installed) SDL3 through `find_package`.
2) Create a minimum `CMakeLists.txt` that justs downloads SDL3, but not configures/builds it (e.g. using a `cmake -P` script). [Here's an example for the N-Gage SDK that downloads external modules](https://github.com/ngagesdk/ngage-toolchain/blob/main/setup/setup.cmake)
3) Then for Android, you need to download SDL3 using the script from step 2, pass its location to the gradle script and run its `build-release.py` script

.github/workflows/ci.yml
madebr: I think a universal config app is near impossible. iPhoneOS needs a swift app and I doubt we want a flutter or Electron config app.
A (minimal) Java/Kotlin app would be very welcome imho.

.github/workflows/ci.yml
```diff
@@ -124,14 +123,14 @@ jobs:
         if: ${{ matrix.android }}
         run: |
           cd android-project && \
-          ./gradlew assemble${{ matrix.debug == 'true' && 'Debug' || 'Release' }} \
+          ./gradlew assembleRelease \
```
madebr: I doubt `assembleRelease` will work on ci as it needs private keys. Until we figure this out, let's do `assembleDebug`?

Covkie: It works and builds an unsigned apk.

.github/workflows/ci.yml
VoxelTek: This was discussed above, and ideally we don't want to have 3 separate config tools. There are already two, the Qt app and the one at `isle.pizza`. However, I understand that neither one is particularly suitable for many of the ports of LEGO Island, and as such, I believe that it may make sense to create a touchscreen/controller-oriented config app, meant for non-desktop platforms (such as 3DS, iOS, Xbox, etc).

If I have spare time, I might start work on that myself.

VoxelTek: > I think a universal config app is near impossible. iPhoneOS needs a swift app and I doubt we want a flutter or Electron config app. A (minimal) Java/Kotlin app would be very welcome imho.
@madebr

Hm, why do you say that? To the best of my understanding, a config app could be created in SDL3 fairly easily. It's not like there's a need to go any lower-level than SDL3's functions.

Ideally, I'd want to stick with Qt for the desktop app. It's certainly much better than SDL when it comes to making UIs, and ideally I'd use it for everything. However, many platforms we've ported to/are porting to don't have Qt (Xbox, 3DS, PSP). As such, I believe that an alternate config application could be made for these platforms, while desktop sticks with the Qt app.

madebr: I was implicitly talking about native looking configuation tools. But you're probably right. ignore my comment then :)

Covkie: From the Libraries page on the SDL wiki; https://github.com/mikke89/RmlUi/ looks like a solution to the horror that would be doing the UI in pure SDL

Algiuxs:
Just wanted to say, that the APK can't Install if it's in release and not debug and unsigned.

Algiuxs:
Wanted to add, that you can sign it yourself for now.

Covkie:
I'm aware. I've been signing locally when testing the action build.

Covkie:
I think this is ready for merging. Runtime GLES3 & SDL Config UI can be later PRs

edit: The Android job wont complete on this repo without the needed secrets

Covkie:
> SDL3 is built twice: once for the aar and once for isle-portable. Apart from that, this looks good.

Seemingly unavoidable due to how gradle native cmake works. It actually clones and builds 5 times, the aar and the four architectures.

madebr:
> > SDL3 is built twice: once for the aar and once for isle-portable. Apart from that, this looks good.
> 
> Seemingly unavoidable due to how gradle native cmake works. It actually clones and builds 5 times, the aar and the four architectures.

I meant two times for each arch. The aar contains `libSDL3.so` for all 4 Android architectures. So the total is 8x.
Going to 4x requires changes to `CMakeLists.txt`, which can be addressed in a future PR.

Covkie:
CMake still reclones some other deps for every architecture (4 times). I don't know if this is a limitation/bug with FetchContent as I've set `DFETCHCONTENT_BASE_DIR` but it just gets overwritten.

ISLE/isleapp.cpp
```diff
@@ -1027,6 +1031,12 @@ bool IsleApp::LoadConfig()
 {
 #ifdef IOS
 	const char* prefPath = SDL_GetUserFolder(SDL_FOLDER_DOCUMENTS);
+#elif defined(ANDROID)
+	// SDL_GetAndroidExternalStoragePath() returns without a trailing / resulting in "filesisle.ini" :(
+	const char* androidPath = SDL_GetAndroidExternalStoragePath();
+	char* prefPath = new char[strlen(androidPath) + 2];
```
foxtacles: Where is this freed?

ISLE/isleapp.cpp
```diff
@@ -1027,6 +1031,12 @@ bool IsleApp::LoadConfig()
...
+	strcpy(prefPath, androidPath);
```
foxtacles: Let's use SDL versions of these functions.

ISLE/android/config.h
```diff
@@ -0,0 +1,9 @@
+
```
foxtacles:
```suggestion
```

ISLE/android/config.cpp
```diff
@@ -0,0 +1,30 @@
...
+#include <SDL3/SDL_log.h>
+#include <SDL3/SDL_system.h>
+#include <iniparser.h>
+
+void Android_SetupDefaultConfigOverrides(dictionary* p_dictionary)
+{
+	SDL_Log("Overriding default config for Android");
+
+	const char* data = SDL_GetAndroidExternalStoragePath();
+	char* savedata = new char[strlen(data) + strlen("/saves/") + 1];
```
foxtacles: Where is this freed? I'd suggest using MxString.

.github/workflows/ci.yml
```diff
@@ -106,7 +107,40 @@ jobs:
 
       - uses: actions/checkout@v4
 
+      - name: Setup Java (Android)
+        if: ${{ matrix.android }}
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: '17'
```
foxtacles, line 121:
```suggestion
          java-version: '17'

```

ISLE/isleapp.cpp
Covkie: wdym?

foxtacles: https://wiki.libsdl.org/SDL3/SDL_strlcpy
etc

Covkie: Is there a reason you want SDL_* functions here while using the normal functions elsewhere

ISLE/isleapp.cpp
Covkie: oops forgot I'd ifed it out

ISLE/android/config.cpp
Covkie: ios/config.cpp doesn't free it so I assumed it was handled.

ISLE/isleapp.cpp
foxtacles: We are using the SDL equivalent functions everywhere where available. strcpy and strcat have no direct equivalents, which is why they were left untouched. However this is new code and it's better to stick as much to SDL as possible.

ISLE/android/config.cpp
foxtacles: Might be incorrect in iOS/config.cpp as well. Haven't reviewed that code.

.github/workflows/ci.yml
```diff
@@ -106,7 +107,41 @@ jobs:
...
+
+      - name: Get CMake (Android)
+        if: ${{ matrix.android }}
+        uses: lukka/get-cmake@latest
+        with:
+          cmakeVersion: 3.30.5
```
madebr, line 127: Is this needed? CMake is already installed on the runners.

CMakeLists.txt
```diff
@@ -73,22 +73,35 @@ if (DOWNLOAD_DEPENDENCIES)
...
+    set(SDL3_SOURCE_DIR "${CMAKE_SOURCE_DIR}/android-project/build/_deps/sdl3-src")
+
+    add_library(SDL3 SHARED IMPORTED)
+    set_target_properties(SDL3 PROPERTIES
+      IMPORTED_LOCATION "${SDL3_SOURCE_DIR}/build-android/${CMAKE_ANDROID_ARCH_ABI}-build/libSDL3.so"
+      INTERFACE_INCLUDE_DIRECTORIES "${SDL3_SOURCE_DIR}/include"
     )
+    add_library(SDL3::SDL3 ALIAS SDL3)
+    add_library(SDL3::Headers ALIAS SDL3)
   else()
```
madebr: Can't this `if(ANDROID) blabla()` just be:
```cmake
if(ANDROID)
  find_package(SDL3 REQUIRED CONFIG COMPONENTS Shared)
```
Gradle creates, when prefab is enabled, cmake configure files from the json files inside the aar.
You're hard coding relative paths here that can break in the future.

.github/workflows/ci.yml
Covkie: The CMake version needs to match exactly with what's specified in build.gradle (default SDK cmake is too old). The version in the runner can change and the current version in the runner is seemingly incompatible with gradle's cmake script

CMakeLists.txt
Covkie: I tried this but it seems to have cache issues on clean runs. (looking in old paths etc)

android-project/app/build.gradle
```diff
@@ -0,0 +1,101 @@
...
+                if (project.hasProperty('cmakeArgs')) {
+                    project.cmakeArgs.split(" ").each {arg -> arguments arg}
+                }
+                abiFilters 'armeabi-v7a', 'arm64-v8a', 'x86', 'x86_64'
+            }
+        }
+    }
+    externalNativeBuild {
+        cmake {
+            version "3.30.5"
```
madebr, line 29: Is this line required?
[SDL_helloworld does not have it.](https://github.com/libsdl-org/SDL_helloworld/blob/501e31c4d9deac135b63b7535c2220d7d71d4d92/android-project/app/build.gradle#L25-L29)

Covkie: default SDK cmake is too old for this project

CMakeLists.txt
madebr: prefab generates config files that do not use the cache, so I'd be really surprised if those were used.
Can you please try to reproduce the problem?

And if they appear, can you share the prefab-generated `SDL3Config.cmake` files?

android-project/app/build.gradle
madebr: I understand :)

CMakeLists.txt
Covkie: I nuked the gradle cache in my home directory and it works now soo

AJenbo:
> Support minimum GLES2 with runtime GLES3

If what you mean by this is that you want to make it run on devices that do not have GLES3 then I don't think there is a need to do so since all Android versions supported by SDL3 have GLES3.

Covkie:
> If what you mean by this is that you want to make it run on devices that do not have GLES3 then I don't think there is a need to do so since all Android versions supported by SDL3 have GLES3.

> https://developer.android.com/develop/ui/views/graphics/opengl/about-opengl
> Caution: Regardless of the Android platform version, a device cannot support the OpenGL ES 3.0 API unless the device manufacturer provides an implementation of this graphics pipeline

There can be cases of platform 21+ devices only having 2.0 support. https://developer.android.com/about/dashboards#OpenGL has 4% of in use devices on 2.0

Either way I won't be doing it in this PR but it could be useful for that 4% and other ports to older devices

VoxelTek:
> There can be cases of platform 21+ devices only having 2.0 support. https://developer.android.com/about/dashboards#OpenGL has 4% of in use devices on 2.0
> 
> Either way I won't be doing it in this PR but it could be useful for that 4% and other ports to older devices

Should be stupid simple to do, considering we used to be using 2.0 before moving to 3.0, so I'd guess 90% of the work is done, just in past commits.

AJenbo:
Yeah we can just dig up the 2.0 renderer, I'm happy to help once this is merged

AJenbo:
@Covkie I added back the OpenGL ES 2.0 renderer https://github.com/isledecomp/isle-portable/pull/678

PR #660, opened 2025-08-06
Wii U Port

Lyelye150:
CMake'd [Done]
Build isle.rpx and LEGO1.rpl [Still in progress]
Test it (tiramisu environment only) [Not There]
maybe Test it in cemu (Wii U emulator may work?) [Maybe?]
make isle.rpx and LEGO1.rpl into isle.wuhb (mainly for Aroma environment compatibility) [Not There]
Test it again (in Aroma environment) [Not There]
send it for review [when im done]

Roadmap:

Add it to homebrew store
Add wup/.app support so it can be in the home menu (mainly for tiramisu environment or adding it to NAND so you can have it forever)
maybe add a minimap for the wii u gamepad only
Lego Island launcher

more info when its done


CMakeLists.txt
VoxelTek, line 1: I'm not too big of a fan of these changes to the `CMakeLists.txt`. There are quite a few questionable or straight-up unwanted changes made here. I'd suggest trying to closer mimic the way other ports have done this.

ISLE/wiiu/config.cpp
VoxelTek, line 1: What does `NO_SDL3` mean? Where does it come from? If the Wii U doesn't have SDL3, why bother checking to see if it does?

Additionally, there are better ways to change the default path values, and these paths that you've provided don't seem to be correct, compared to the standard structure of the LEGO Island disc.

VoxelTek (changes requested):
At this point, there are several issues I have with the changes made, especially those made to `CMakeLists.txt`. I would heavily recommend that this be changed to better mimic how other ports have modified this file.

There have not been any changes made yet to accommodate actually compiling and running the game on Wii U. As the Wii U does not support SDL3, either SDL3 must be ported to the Wii U, or `isle-portable` must be backported to SDL2, both of which would be a fairly-difficult undertaking that would have to be completed *before* any progress can begin on a Wii U port.

AJenbo:
This breaks all other versions, a change like this should probably come in several PRs that rework the needed parts before landing the port it self.

jpeisach:
IMO since we are using SDL3, which is primarily for newer systems, maybe a fork of this repo for SDL2 might be nice so then systems that would benefit them (which are primarily lower memory systems) could have a dedicated more lightweight version of isle

AJenbo:
It should be pretty straight forward to make a small wrapper lib that can abstract SDL2/3 and keep it all in the same repo

Issue #583, opened 2025-07-11
Background not cleared in building


foxtacles:
i.e. gas station:

<img width="1619" height="744" alt="Image" src="https://github.com/user-attachments/assets/37a77e57-b562-44a2-bee6-207e6cdd4711" />

foxtacles:
@AJenbo what would be the cleanest way to solve this in your opinion? I know where in the code base we are in a vehicle vs. building, what would be the best approach to tell the renderer to not draw anything outside the 2D overlay?  I'm not sure where the responsible code path is in miniwin

AJenbo:
Yeah you will have to make the game aware and render something at a negative position. For DevilutionX what we did was extend the load screens, it's a bit of work but can look pretty decent.

Step one:

<img width="1110" height="480" alt="Image" src="https://github.com/user-attachments/assets/ee1c1e7a-a001-4627-8972-71955500dbf4" />

AJenbo:
A couple of hours later and this is about as well as I can do:

<img width="1120" height="480" alt="Image" src="https://github.com/user-attachments/assets/63f65820-9feb-4dd0-89f1-5c40ea854226" />

AJenbo:
Final-2.zip
<img width="1120" height="480" alt="Image" src="https://github.com/user-attachments/assets/b0fa3625-6bed-4b30-8886-37af5b9727c7" />

AJenbo:
Usually what we do is cut out the 640x480 part and have it be transparent so that it only adds sides, you might want to bleed it a bit since this isn't always rendered 1:1

foxtacles:
This is the place where we can override bitmaps if desired:

https://github.com/isledecomp/isle-portable/blob/master/LEGO1/omni/src/video/mxstillpresenter.cpp#L73

I'm not really sure how best to progress though. Replacing the data like the texture loader does seems easy enough, no idea about the rest going from there.

foxtacles:
It is worth exploring how we can enable modding through loading our own, newly created `.SI` asset packs. This would definitely be the desirable long term solution because it allows us to modify / extend most data in the game, and not have to handcraft solutions for every aspect. Existing tooling (`SiEdit`) is already out there:

<img width="1920" height="1034" alt="Image" src="https://github.com/user-attachments/assets/ea33e7aa-dd0b-452d-a7f2-fb692b18d152" />

[...]

PR #664, opened 2025-08-10
Add SI Loader extension

foxtacles:
This adds a new extension that allows adding custom SI files to the game.

Using `libweaver`, assets that are provided by this repository can be built by enabling `ISLE_BUILD_ASSETS` in CMake. For this first iteration, an initial version of `widescreen.si` is built (see #583) which is meant to add widescreen background bitmaps to building areas.

In order for custom SIs to interact with existing assets, two new SI "extra" directives have been added (`StartWith` and `RemoveWith`). These allow new assets to be launched / removed alongside existing assets. In the future, further directives will be added (i.e. ones that allow completely replacing an existing asset).

`widescreen.si` adds a single widescreen background to the Garage door area for now:

image

The extension can be configured similarly to the texture loader:

```
[si loader]
# Comma-separated list of paths relative to disk or CD path
files=\lego\scripts\extra\widescreen.si

[extensions]
SI Loader=true
```

`libweaver` doesn't currently compile in `WINDOWS_STORE` environments so I had to disable extensions for the Xbox build for now.


3rdparty/CMakeLists.txt
```diff
@@ -72,3 +73,23 @@ target_include_directories(imgui PUBLIC ${imgui_SOURCE_DIR})
...
+
+if(DOWNLOAD_DEPENDENCIES)
+    include(FetchContent)
+    FetchContent_Populate(
+        libweaver
+        URL https://github.com/isledecomp/SIEdit/archive/6da93b2072c41c41d526b8b9df7d4292be1f0f55.tar.gz
+        URL_MD5 ae59007fcb9efadc06c67621e1e107cb
+    )
+else()
+    set(libsmacker_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/libweaver")
```
madebr: Ideally, you'd do `find_package`, or `find_path`/`find_library` but [it looks SIEdit currently does not do install()``](https://github.com/isledecomp/SIEdit/blob/master/lib/CMakeLists.txt)
```suggestion
    set(libweaver_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/libweaver")
```

CMakeLists.txt
```diff
@@ -644,6 +647,36 @@ if (ISLE_BUILD_CONFIG)
...
+  add_custom_command(
+    OUTPUT  ${GENERATED_ASSETS_DIR}/.stamp
+    COMMAND ${CMAKE_COMMAND} -E make_directory ${GENERATED_ASSETS_DIR}
+    COMMAND $<TARGET_FILE:asset_generator> ${GENERATED_ASSETS_DIR}
+    COMMAND ${CMAKE_COMMAND} -E touch ${GENERATED_ASSETS_DIR}/.stamp
+    WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/assets
+    DEPENDS asset_generator
+    COMMENT "Generating assets into ${GENERATED_ASSETS_DIR}/"
+    VERBATIM
+  )
```
madebr: If you can let `assets/main.cpp` generate a depfile, you can have [dependency tracking with (most) generators](https://cmake.org/cmake/help/latest/command/add_custom_command.html#generating-files).
It should write a `.d` file like:
```
absolute/path/to/widescreen.si: absolute/path/to/garadoor.bmp
``

**Review by AJenbo — COMMENTED** [review]
I don't have anything to add

**Comment by foxtacles** [comment]
> I don't have anything to add

Could use a few more widescreen backgrounds :D

**Comment by AJenbo** [comment]
lets see what happens over the next few days

File: CMakeLists.txt (hunk 2 above)
Comment (foxtacles) [reply CMakeLists.txt]: Thanks, I added the depfile. Does this look good?

```
/home/foxtacles/isle-portable/build/assets/widescreen.si: /home/foxtacles/isle-portable/assets/widescreen/garadoor.bmp
```

File: CMakeLists.txt [hunk 3]
```diff
@@ -644,6 +647,38 @@ if (ISLE_BUILD_CONFIG)
   endif()
 endif()
 
+if(ISLE_BUILD_ASSETS)
+  message(STATUS "Asset building is enabled")
+  set(GENERATED_ASSETS_DIR "${CMAKE_BINARY_DIR}/assets")
+  set(GENERATED_ASSETS_DEPFILE "${GENERATED_ASSETS_DIR}/.d")
```
madebr: I meant a file with a `.d` extension. But this works too :)

madebr (approved):

extensions/src/siloader.cpp
```diff
@@ -0,0 +1,110 @@
...
+void SiLoader::Initialize()
+{
+}
+
+bool SiLoader::Load()
+{
+	char* files = SDL_strdup(options["si loader:files"].c_str());
+	char* saveptr;
+
+	for (char* file = SDL_strtok_r(files, ",\n", &saveptr); file; file = SDL_strtok_r(NULL, ",\n", &saveptr)) {
```
VoxelTek, line 26: Does this imply that each separate file needs to be separated not only by a `,` but also a newline? Or am I misunderstanding?

extensions/src/siloader.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+	return true;
+}
+
+bool SiLoader::LoadFile(const char* p_file)
+{
+	si::Interleaf si;
+	MxStreamController* controller;
+
+	MxString path = MxString(MxOmni::GetHD()) + p_file;
```
VoxelTek, line 64: Due to the way this is structured, this requires that each file path given begins with `/`, which is somewhat difficult to accomplish cleanly and safely from `isle-config`. Might it be possible to rework this somewhat, ideally using a cleaner method of obtaining the full absolute path? 

~~I feel like SDL has functions to do something like this, although I may be wrong.~~
While SDL3 has filesystem functions, it does not appear to be particularly extensive. I was hoping for something similar to that in Qt, but that doesn't seem to be the case.

VoxelTek: Alternatively, switching to an absolute path would solve this issue, however I understand that may not be ideal in all situations.

VoxelTek: One (somewhat hacky) solution could be to somehow split `p_file` into an array by `/` or `\` characters, and then stitch them back together, adding `/` and `\` wherever necessary.

Covkie: currently the config does this to the paths and fails to parse
image
```
[si loader]
files                          = ",../isle-portable/build/assets/widescreen.si,../../../../../home/kylie/Desktop/widescreen.si"```

Comment (VoxelTek) [reply extensions/src/siloader.cpp:64]: > currently the config does this to the paths and fails to parse 
> image
> 
> ```
> [si loader]
> files                          = ",../isle-portable/build/assets/widescreen.si,../../../../../home/kylie/Desktop/widescreen.si"```
> ```

Yeah, I only just discovered that, hence my coming here to see if there was something I could do to fix it. Not sure why you've got a `,` before any of the paths though, that shouldn't be happening.

Comment (VoxelTek) [reply extensions/src/siloader.cpp:64]: @Covkie the intent is to place the `assets` folder in the game data path (`cdpath` or `diskpath`), and then the directory provided should be relative to that. If you remove the `,` at the start, replace it with `/`, and probably remove the second path, it should work.

i.e.
```
[si loader]
files                          = "/../isle-portable/build/assets/widescreen.si"
```

Covkie: > Not sure why you've got a , before any of the paths though, that shouldn't be happening.

Looks to be from whatever is causing the `../`. Doesn't happen when it's in the data dir

VoxelTek: > > Not sure why you've got a , before any of the paths though, that shouldn't be happening.
> 
> Looks to be from whatever is causing the `../`.

No, not really. The `../` is expected, the `,` before all that is not.

> Doesn't happen when it's in the data dir

What do you mean? The `,` doesn't occur when it's in the data dir, but it does when it's outside of it? That shouldn't be happening, and in fact doesn't happen for me.

EDIT: Figured out why, I think. My code expected the user to actually select a file when the `+` button is pressed. However, if the file dialog is cancelled and no file is selected, it's possible it was appending an empty string to the path list. `isle-config` would then add a comma to separate it from the other paths...even though there's nothing there. It would likely fix itself after the next change you make in `isle-config`. Anyway, I've now fixed up the code. Haven't been able to fix the requirement for `/`, so it still doesn't yet work properly, but I'll see what I can do.

Covkie: oh there's an invisible irremovable entry so if i simply use the + the leading comma gets added.
image

VoxelTek: Should hopefully be fixed with #669. May not remove your "invisible" entry, but it should prevent it from happening again. You may simply have to remove that line from your `isle.ini`.

PR #667, opened 2025-08-11
Add options for `si loader` extension to `isle-config`.

VoxelTek:
image
image


VoxelTek:
Damn, you just missed an extra commit I added to make things a tiny bit nicer looking
image

PR #671, opened 2025-08-13
Add `hdmusic.si` asset pack, extend SiLoader with `Replace` directive

### LINKED ISSUE #346: Severe Memory Leakage?
For context: I am working on a mod to this source code that aims to provide a new story, HQ music, etc.

I recently noticed (after playing the game for longer than a few minutes) that it would crash soon before the portion of the Brickster Hunt after you've chased him but before you rebuild the helicopter.

For kicks, I compiled under debug, and attached the Visual Studio 2022 debugger, and watched in horror as the Memory usage very quickly skyrocketed (as in, an exponential amount of mb of memory used up every minute), until the program reached 1.7GB of memory usage and crashed when trying to load into the music stream buffer. The error? Bad memory allocation.

After further testing, I have concluded that the SI streamer is at fault. The game idles at just about 104MB of RAM usage from my testing, and skyrockets whenever audio is being streamed. The stock music usually raises this to about 140MB of RAM usage, and some of the higher quality tracks idle around here too for whatever reason, such as The Jail and Police Station themes. However, other higher quality tracks, such as The Lego Island Theme (if that's the track's real name) and the Information center theme bring this amount up way higher, and in the case of the Chasing the Brickster, the memory usage makes it's way to a point where the game can no longer allocate any memory at all (in my testing, this is just under 2GB, which makes sense since that's the memory limit for 32bit applications).

Edit: I have figured out that the Higher Quality tracks that weren't causing memory leaks was because those particular tracks I was instead using the unused 22050Hz versions instead of the masters. Replacing these versions with the 44100Hz masters reveals the same memory leak once again.

This isn't really something to fix for the decomp, since the aim is accuracy, but this is something to keep in mind if you plan on making a mod of Lego Island.

foxtacles:
This adds high quality versions of the music through a new asset pack `hdmusic.si`. The music was sourced from cassette tapes here: https://projectisland.org/music/  as well as this recreation of the Hospital theme: https://www.youtube.com/watch?v=BuW7Yf3Qx6E

As far as I know I've added all possible HQ versions that exist. I've not substituted unused music such as `MusicTheme1`, `MusicTheme3` or `Act2Cave`. What's missing is `CentralNorthRoad_Music`, `GarageArea_Music`, `Elevator_Music` and `PizzaMission_Music`, since there don't seem to be recovered versions of these.

image

I've had to make minor changes to `MxBackgroundAudioManager` to support streaming audio from arbitrary SI files. The class was built with the implicit assumption that only `JUKEBOX.SI` is used, at least in some places.

~While this PR is functional it's not quite production ready. Since we rely on meta information in the SI asset packs, I'm using `libweaver` for parsing, however currently the library requires loading the entire files into memory which is expensive in this case since `hdmusic.si` is about 450MB in size.~  Addressed: https://github.com/isledecomp/SIEdit/pull/37

Close #346


assets/main.cpp
```diff
@@ -16,30 +16,182 @@ si::MemoryBuffer mxHd;
...
-	GaraDoor_Wide.up_.y = 1.0;
-	GaraDoor_Wide.ReplaceWithFile("widescreen/garadoor.bmp");
-	si.AppendChild(&GaraDoor_Wide);
-	depfile << file << ": " << (std::filesystem::current_path() / "widescreen/garadoor.bmp").string() << std::endl;
-
-	si.Write(file.c_str());
+	int i = 0;
+	for (const AssetView& asset : widescreenBitmaps) {
+		si::Object* object = new si::Object;
+		std::string file = std::string("widescreen/") + asset.name + ".bmp";
```
madebr, line 36: Does and should this program verify whether the bmp's and wav's are actually there, and are no LFS text files? Perhaps it should check the file magic?

foxtacles: This should take care of it:

```
		if (!object->ReplaceWithFile(file.c_str())) {
			abort();
		}
```

If it's not a valid file according to the specified type this should always abort

madebr: My only remark is to print something to stderr as feedback.

I made this comment first thinking an ordinary `git fetch` or `git pull` would not download lfs files, but apparently they are cloned automatically.

PR #678, opened 2025-08-15
Add OpenGL ES 2.0 renderer

AJenbo:
This recoveres the OpenGL ES 2.0 renderer which was rewritten into an OpenGL ES 3.0 renderer, so that both are now available in order to support older Android devices which may not be able to run OpenGL ES 3.0.


foxtacles (approved):

AJenbo:
Updated the code with some fixes from the ES 3.0 renderer and adhere to the configured AF settings.

VoxelTek:
I noticed that the OpenGL ES 2.0 renderer code doesn't have anything for MSAA. Does it not support it, or is there some other reason for not including MSAA?

AJenbo:
It's not supported on webgl at least, but I think it's possible to support it via extensions on Opengl es 2.0. but doing so is even more complicated then for Opengl es 3.0 and it's uncertain how much hardware would support it, probably most that could would also support 3.0. so I consider it not worth the effort.

PR #680, opened 2025-08-16
Use 8bit surfaces for 2D content

AJenbo:
This reduces memory usage by at least 1.3MB by keeping 2D surfaces.

It also aligns some of the very similar functions in this class.


foxtacles:
dx5: intro isn't visible; backgrounds not visible:

image

AJenbo:
Buu

AJenbo:
@foxtacles please give it another try

foxtacles:
> @foxtacles please give it another try

Same issue

VoxelTek:
Seems like an issue exclusive to DX5, I get no issues when running it.

Definitely worth fixing, given that this would potentially open the door for more platforms to port LEGO Island to. I'd lend a hand, however I don't use Windows and would be unable to test if any fixes were actually successful.

VoxelTek:
hm, ok, so this seems to break the cursor on the PSP version.

image

Note the black box near the top-right corner. That's the in-game cursor.

There are also several other fairly-minor graphical glitches.
image
However, aside from these, everything appears to function fine.

I'll revert to an earlier commit to fully determine if this is the 8-bit surface change.

EDIT: Yep, seems that this is the cause of it.

AJenbo:
> There are also several other fairly-minor graphical glitches.

I'm guessing your refering to the broken clipping and transparency which are things I'm still waiting for feedback on from a hardware test before trying to apply fixes for.

AJenbo:
> hm, ok, so this seems to break the cursor

Fixed (it's not PSP specific)

AJenbo:
@foxtacles this should be ready now

foxtacles:
The background in the Infocenter shows now, but the intro isn't visible
(dx5)

AJenbo:
Hm sorry for that, I added some more guess work, try seeing if it works now.

PR #685, opened 2025-08-20
Add keymap rebinding

VoxelTek:
Allows input keymaps to be rebinded.

- [ ] Detect input keys
  - [ ] Make popup SDL window look better
- [ ] Store keymap to `.ini`
- [ ] Load keymap from `.ini`
- [ ] Apply keymap to game

Screenshot_20250820_093422


VoxelTek:
I'm not sure how to resolve the CI errors, annoyingly. It compiles fine and without complaint for me on Linux, so...

As for the incomplete task:
In order to get key presses, and in the correct format for SDL, I need to create a temporary SDL window, and have that window be focused. However, I don't really have anything to display on that window. Displaying text would require SDL_TTF, and displaying an image would require finding a suitable image and including it as a header file. Neither is particularly ideal. Currently, the temporary SDL window ends up looking like this:
image
It ends up just showing whatever was behind it when it spawned in. I'm not sure what the best solution to this would be. Hell, I'm still concerned that the solution I've implemented for input key mapping is terrible.

PR #692, opened 2025-08-20
De-duplicate surface blitting

AJenbo:


LEGO1/omni/src/video/mxdisplaysurface.cpp
```diff
@@ -434,7 +434,7 @@ void MxDisplaySurface::VTable0x28(
 	MxBITMAPINFO* bmi = p_bitmap->GetBitmapInfo();
 	if (bmi) {
 		PALETTEENTRY pe[256];
-		for (int i = 0; i < 256; i++) {
+		for (int i = 1; i < 256; i++) {
```
foxtacles, line 437: What does this change do?

foxtacles:
DX5 (glitches on intro, no backgrounds)

https://github.com/user-attachments/assets/658815e7-1c48-4f04-b3cc-c76c3ab7d182

LEGO1/omni/src/video/mxdisplaysurface.cpp
AJenbo: Makes the edges of scaled transparent surfaces black instead of purple, it only affects devices that have resolutions less then 640x480

AJenbo:
You have to admit it looks pretty cool in the intro :)

AJenbo:
While nice this isn't really an important change so I don't think I will pursuit this one further.

PR #695, opened 2025-08-21
Remove obsolete `mediapath` stuff

### LINKED ISSUE #662: Remove `media path`?
To the best of my understanding, the `media path` config value is obsolete and unused. The code appears to back this up, and while it loads `media path` it doesn't actually seem to do anything with it. As such, could it maybe just be removed entirely? i.e. `isle` no longer cares for it or generates it in the `.ini`, and `isle-config` dropped it ages ago.

VoxelTek:
Removes all obsolete references to `mediapath`. 

Closes #662

### DIFF
--- a/ISLE/android/config.cpp
+++ b/ISLE/android/config.cpp
@@ -20,7 +20,6 @@ void Android_SetupDefaultConfigOverrides(dictionary* p_dictionary)
 
 	iniparser_set(p_dictionary, "isle:diskpath", data);
 	iniparser_set(p_dictionary, "isle:cdpath", data);
-	iniparser_set(p_dictionary, "isle:mediapath", data);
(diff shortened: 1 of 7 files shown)


foxtacles:
This does not improve portability or functionality and only increases potential friction by further increasing the delta between portable and the original decomp.

VoxelTek:
To be fair, the presence of `mediapath` resulted in some issues when attempting the port to Android, but yeah ok fair enough.

foxtacles:
> To be fair, the presence of `mediapath` resulted in some issues when attempting the port to Android, but yeah ok fair enough.

I've seen no definitive proof of that other than someone's statement. I don't see any reason why the presence of this would cause any issues. If there's documented proof we can look at it and address it appropriately.

PR #698, opened 2025-08-21
Use system stack size for MxDiskStreamProviderThread

### LINKED ISSUE #687: Black Screen on Launch (Android 10)
After setting up LEGO Island on my Android 10 device and launching the app, it black screens. No error message or anything. I don't know if it's something on my end or a bug or something else. Even relaunching the app doesn't work. Any help with this issue is much appreciated.

Thanks!

Covkie:
fixes #687

### DIFF
--- a/LEGO1/omni/src/stream/mxdiskstreamprovider.cpp
+++ b/LEGO1/omni/src/stream/mxdiskstreamprovider.cpp
@@ -32,7 +32,7 @@ MxResult MxDiskStreamProviderThread::Run()
 MxResult MxDiskStreamProviderThread::StartWithTarget(MxDiskStreamProvider* p_target)
 {
 	m_target = p_target;
-	return Start(0x1000, 0);
+	return Start(0, 0);
 }
 
 // FUNCTION: LEGO1 0x100d0f70


LEGO1/omni/src/audio/mxsoundmanager.cpp
```diff
@@ -119,7 +119,7 @@ MxResult MxSoundManager::Create(MxU32 p_frequencyMS, MxBool p_createThread)
 	if (p_createThread) {
 		m_thread = new MxTickleThread(this, p_frequencyMS);
 
-		if (!m_thread || m_thread->Start(0, 0) != SUCCESS) {
+		if (!m_thread || m_thread->Start() != SUCCESS) {
```
foxtacles: We shouldn't make unnecessary "cleanup" changes to the code base. This is a fork based on the original decompilation that keeps receiving updates, it will increase merge conflicts.

Leave everything untouched except the implementation of `MxThread::Start`. You can simply ignore the parameters. I'd also suggest using `SDL_CreateThread` directly (using `NULL` for the name parameter) since we no longer need the property based approach when we are not specifying stack size.

Covkie: just dropped the useless arg commit

foxtacles:
lgtm - will merge once confirmed by @HyperHaxStudios that it fixes issue #687

Covkie:
it does 0 * 4 for every Start() call but it doesn't really matter

PR #703, opened 2025-09-01
Add new widescreen backgrounds

foxtacles:
See #583 - thanks @AJenbo 🎉


AJenbo:
Good timing, it will be about a week before I can make any more

PR #707, opened 2025-09-05
feat: SDL2 shim for ports without SDL3 support

Covkie:
Everything works as far as I can tell.

I would appreciate some help with the build script refactoring.

Uses system SDL2.

### DIFF
--- a/LEGO1/lego/legoomni/src/common/legoutils.cpp
+++ b/LEGO1/lego/legoomni/src/common/legoutils.cpp
@@ -31,8 +31,10 @@
 #include "realtime/realtime.h"
 #include "scripts.h"
 
-#include <SDL3/SDL_events.h>
+#if SDL_MAJOR_VERSION >= 3
 #include <SDL3/SDL_process.h>
+#endif
+#include <SDL3/SDL_events.h>
 #include <SDL3/SDL_stdinc.h>
 #include <SDL3/SDL_version.h>
 #include <stdio.h>


Covkie:
~~Does anyone know why SDL_MAJOR_VERSION doesn't seem to be defined in the runner? It works locally.~~

Covkie:
Could I get some help with this? The game starts and loads the intro animation then freezes and gets stuck in a loop at https://github.com/isledecomp/isle-portable/blob/57e918904cd30e112829a0b366606ac2ecb94bd0/LEGO1/lego/legoomni/src/video/legovideomanager.cpp#L304-L307

I assume theres some subtle SDL2 difference that i missed which is causing this..

3rdparty/sdl3-shim/SDL3/SDL_mutex.h
```diff
@@ -0,0 +1,12 @@
...
+#include <SDL2/SDL_mutex.h>
+
+// https://wiki.libsdl.org/SDL3/README-migration#sdl_mutexh
+
+#define SDL_Mutex SDL_mutex
+
+#define SDL_Semaphore SDL_sem
+
+#define SDL_WaitSemaphore SDL_SemWait
+#define SDL_SignalSemaphore SDL_SemPost
```
AJenbo, line 12: These may fail in SDL2 with != 0, but is void in SDL3 so probably they should handle errors as failure.

3rdparty/sdl3-shim/SDL3/SDL_iostream.h
```diff
@@ -0,0 +1,49 @@
...
+typedef int SDL_IOWhence;
+
+#define SDL_IO_SEEK_SET RW_SEEK_SET
+#define SDL_IO_SEEK_CUR RW_SEEK_CUR
+#define SDL_IO_SEEK_END RW_SEEK_END
+
+// #define SDL_IOFromConstMem SDL_RWFromConstMem
+#define SDL_IOFromFile SDL_RWFromFile
+#define SDL_IOFromMem SDL_RWFromMem
+#define SDL_CloseIO SDL_RWclose
```
AJenbo, line 18: This need to convert != 0 to bool, they could likely be the issue you are having.

Covkie: All of these either have the same return values or the return value isn't checked (SDL_CloseIO)

3rdparty/sdl3-shim/SDL3/SDL_timer.h
```diff
@@ -0,0 +1,70 @@
...
+typedef Uint32 SDL3_TimerCallback (void *userdata, SDL_TimerID timerID, Uint32 interval);
+
+struct SDL2TimerShimData {
+	SDL3_TimerCallback *callback2;
+	void *userdata2;
+	SDL_TimerID id;
+};
+
+static std::map<SDL_TimerID, void*> g_timers;
+static SDL_mutex *g_timerMutex = NULL;
```
AJenbo: Should this not be initialized with SDL_CreateMutex() somewhere?

3rdparty/sdl3-shim/SDL3/SDL_audio.h
```diff
@@ -0,0 +1,101 @@
...
+
+struct ShimAudioCtx {
+	SDL_AudioStream* stream;
+	SDL_AudioStreamCallback* callback;
+	void *userdata;
+	SDL_AudioSpec obtained;
+};
+
+static std::map<SDL_AudioDeviceID, ShimAudioCtx> g_audioCtxs;
+static SDL_mutex *g_audioMutex = NULL;
```
AJenbo: Where is the init?

AJenbo:
Changing `UseSurface()` it always `return true` "fixes" the issue. Maybe SDL2 and SDL3 aren't firing the same events leading to the state not being the same for loading and rendering.

Covkie:
Works if you comment out 
https://github.com/isledecomp/isle-portable/blob/6ea82ae144b91874bd41fe32e9ffb82507a87ef0/LEGO1/omni/src/video/mxstillpresenter.cpp#L96-L97

foxtacles:
We should have at least one CI build that tests this.

Are there any ports lined up that will make use of this? I wouldn't want to add this complexity if it won't actually be used.

CMakeLists.txt
```diff
@@ -74,6 +81,7 @@ if (DOWNLOAD_DEPENDENCIES)
   message(STATUS "Fetching SDL3 and iniparser. This might take a while...")
   include(FetchContent)
 
+  if(NOT USE_SDL2)
   if(ANDROID)
     # Built by Gradle
     find_package(SDL3 REQUIRED CONFIG COMPONENTS Shared)
```
madebr: I'd just do:
```cmake
if(ANDROID)
  if(NOT USE_SDL3)
    message(FATAL_ERROR "SDL2 port is unavailable for Android")
  endif()
```
Same for `WINDOWS_STORE`. @Helloyunho can fix this later. Upstream SDL2 has uwp support.

CONFIG/MainDlg.cpp
```diff
@@ -194,6 +194,7 @@ bool CMainDialog::OnInitDialog()
 			m_ui->exFullResComboBox->setCurrentIndex(i);
 		}
 	}
+	SDL_free(displayModes);
```
madebr: Where is `displayModes` stored? If it's an object member, you'd also need to set it to `nullptr`.
If it's unused elsewhere, then can it be removed as a member?

CMakeLists.txt
```diff
@@ -55,10 +55,17 @@ cmake_dependent_option(ISLE_COMPILE_SHADERS "Compile shaders" ON "SDL_SHADERCROS
...
+option(USE_SDL2 "Use SDL2 instead of SDL3 via a shim" OFF)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}" CACHE PATH "Directory where to put executables and dll")
 set(ISLE_EMSCRIPTEN_HOST "" CACHE STRING "Host URL for Emscripten streaming (e.g., https://test.com)")
 cmake_dependent_option(BUILD_SHARED_LIBS "Build lego1 as a shared library" ON "NOT EMSCRIPTEN" OFF)
 
+if(USE_SDL2)
+  add_compile_definitions(SDL_MAJOR_VERSION=2)
+else()
+  add_compile_definitions(SDL_MAJOR_VERSION=3)
+endif()
```
madebr: I don't like this defines macro's in SDL's namespace (=`SDL_` prefix).
Something like `ISLE_SDL(2|3)` is safer imho.

CMakeLists.txt
```diff
@@ -74,6 +81,7 @@ if (DOWNLOAD_DEPENDENCIES)
   message(STATUS "Fetching SDL3 and iniparser. This might take a while...")
   include(FetchContent)
 
+  if(NOT USE_SDL2)
```
madebr, line 79: @Helloyunho SDL2 support is useful for UWP/xbox.

3rdparty/CMakeLists.txt
```diff
@@ -93,3 +105,40 @@ add_library(libweaver STATIC
...
+#    FetchContent_Declare(
+#            SDL2
+#            GIT_REPOSITORY "https://github.com/libsdl-org/SDL.git"
+#            GIT_TAG 5d249570393f7a37e037abf22cd6012a4cc56a71
+#            EXCLUDE_FROM_ALL
+#    )
+#    FetchContent_MakeAvailable(SDL2)
+    # Find SDL2 normally
+    find_package(SDL2 REQUIRED)
+
```
madebr: Can this be enabled again? And use the `SDL2` branch. We're fine with using the latest, potentially unstable.

3rdparty/CMakeLists.txt
```diff
@@ -93,3 +105,40 @@ add_library(libweaver STATIC
...
+
+  add_library(SDL3_shim STATIC
+          sdl3-shim/main.cpp
+  )
+  target_link_libraries(SDL3_shim PRIVATE SDL2::SDL2)
+  target_include_directories(SDL3_shim PUBLIC
+          ${SDL2_INCLUDE_DIRS}
+          "${CMAKE_CURRENT_SOURCE_DIR}/sdl3-shim"
+          "${CMAKE_CURRENT_BINARY_DIR}/sdl3-shim"
```
madebr: the SDL3 shim headers must be included with a (very) high priority, even higher then the SDL2 headers.
Something like the following I think:
```cmake
target_include_directories(<target> PUBLIC BEFORE "${CMAKE_CURRENT_SOURCE_DIR}/sdl3-shim" "${CMAKE_CURRENT_BINARY_DIR}/sdl3-shim")
```

3rdparty/CMakeLists.txt
```diff
@@ -93,3 +105,40 @@ add_library(libweaver STATIC
...
+    configure_file(${header} ${CMAKE_CURRENT_BINARY_DIR}/sdl3-shim/SDL3/${fname} COPYONLY)
+  endforeach()
+
+
+  add_library(SDL3_shim STATIC
+          sdl3-shim/main.cpp
+  )
+  target_link_libraries(SDL3_shim PRIVATE SDL2::SDL2)
+  target_include_directories(SDL3_shim PUBLIC
+          ${SDL2_INCLUDE_DIRS}
```
madebr: This is not needed: you're already linking to `SDL2::SDL2`.

3rdparty/CMakeLists.txt
```diff
@@ -48,31 +48,43 @@ add_library(libsmacker STATIC
...
-    ${imgui_SOURCE_DIR}/imgui_demo.cpp
-    ${imgui_SOURCE_DIR}/backends/imgui_impl_sdl3.cpp
-    ${imgui_SOURCE_DIR}/backends/imgui_impl_sdlrenderer3.cpp
-)
-target_include_directories(imgui PUBLIC ${imgui_SOURCE_DIR})
-target_link_libraries(imgui PUBLIC SDL3::Headers)
-target_link_libraries(imgui PRIVATE SDL3::SDL3)
-set_property(TARGET imgui PROPERTY CXX_CLANG_TIDY "")
+  target_include_directories(imgui PUBLIC ${imgui_SOURCE_DIR})
+  target_link_libraries(imgui PUBLIC SDL3::Headers)
```
madebr: I don't think this is needed. `SDL3::SDL3` already provides it.
```suggestion
```

3rdparty/CMakeLists.txt
madebr: This is probably why you're doing `configure_file(COPYONLY)`.

Covkie:
> Are there any ports lined up that will make use of this?

I've made this with the intention of using devkitpro's ogc-sdl-2 branch for a wii port.

3rdparty/CMakeLists.txt
Covkie: SDL3_shim or ${SDL2_INCLUDE_DIRS}. The latter is needed to satisfy imgui - something to do with how they prefix SDL2 includes

CMakeLists.txt
Covkie: When building CONFIG SDL_MAJOR_VERSION isn't set which causes issues so this is just to ensure

3rdparty/CMakeLists.txt
Covkie: I've done it this way so the SDL2 headers are copied under SDL3/...h then the shim headers are on top. Otherwise `#include <SDL3/...h> wouldn't find the SDL2 header.

CONFIG/MainDlg.cpp
Covkie: Unsure, asan was complaining so I threw it in there to move on.

CMakeLists.txt
Covkie: This if(NOT USE_SDL2) is for stopping SDL3 fetching entirely, not for Android specifically. 
SDL2 has Android/UWP support but I haven't tested what happens when building those with USE_SDL2 (probably errors from missing shims)

CMakeLists.txt
AJenbo: Are you sure that isn't just because your not including SDL_version.h in the given file? SDL_VERSION_ATLEAST(X, Y, Z) is a standard macro and relies on SDL_MAJOR_VERSION internally.

https://wiki.libsdl.org/SDL2/SDL_VERSION_ATLEAST

Covkie: It was only failing in the runner so I'm not exactly certain. I can look later

CMakeLists.txt
Helloyunho: Sounds good but I'm not sure if SDL2 has GPU acceleration for WinRT(UWP)

AJenbo: It would require a new renderer backend, currently only the software renderer would work under UWP on SDL2. We would need to write a custom D3D 11/12 render for it.

CONFIG/MainDlg.cpp
Covkie: I've dropped this as it's not really related to this PR

Covkie:
looks like -Werror=dev causes issues with the SDL2 configuration

CMakeLists.txt
Covkie: NVRHI/NRI might be an interesting SDL GPU alternative for SDL2. Both support Vulkan and D3D 11/12

LEGO1/mxdirectx/mxdirectxinfo.cpp
```diff
@@ -228,11 +229,16 @@ BOOL MxDeviceEnumerate::EnumDirectDrawCallback(LPGUID p_guid, LPSTR p_driverDesc
 
 	result = lpDD->QueryInterface(IID_IDirect3DMiniwin, (void**) &miniwind3d);
 	if (result == DD_OK) {
+#if SDL_MAJOR_VERSION >= 3
```
madebr, line 232: Use `SDL_VERSION_ATLEAST` macro

madebr:
This patch is slightly smaller then [the wii patch for SDL2](https://github.com/devkitPro/pacman-packages/blob/master/wii/SDL2/SDL2-2.28.5.patch).
I think it would be better use of your and our time to add proper wii support to SDL3.

Since wii is not supported anymore by Nintendo, the SDL project should be welcoming to maintain it as part of SDL3 (unless the port is doing unholy illegal things).

Covkie:
> This patch is slightly smaller then [the wii patch for SDL2](https://github.com/devkitPro/pacman-packages/blob/master/wii/SDL2/SDL2-2.28.5.patch).
I think it would be better use of your and our time to add proper wii support to SDL3.

Most of the lines in this are simple #defines (or commented copy pastes from the migration wiki) I really doubt migrating devkitpro's SDL2 patches to SDL3 will be less work.

madebr:
You already did a big job by making it SDL2 compatible, so moving SDL completely out, should be a refactoring job. (Don't worry about supporting multiple SDL versions by one binary)

foxtacles:
> I give this approach a NAK. This gives us a big technical debt. Platforms should either all get sdl3 support, or isle-portable should get an intermediate layer.
> 
> I helped adding such a layer in [dethrace](https://github.com/dethrace-labs/dethrace/tree/main/src/harness/platforms): it supports SDL 1, 2 and 3. (All in a single binary)

I agree with this take.

AJenbo:
I would suggest using C++ objects rather then C structs and pointers like [dethrace](https://github.com/dethrace-labs/dethrace/tree/main/src/harness/platforms), but other then that I agree and abstraction with multiple implementations would be more maintainable then making SDL2 look like SDL3

Covkie:
How would you want such an intermediate layer to be laid out? Prefixing everything in INTERMEDIATE_ over SDL_ and ifdefing its definition? I feel like that's way way more work than whats currently happening.

madebr:
> How would you want such an intermediate layer to be laid out? Prefixing everything in INTERMEDIATE_ over SDL_ and ifdefing its definition? I feel like that's way way more work than whats currently happening.

Let's call this new layer `mortar` (as it connects blocks/bricks), so prefix with `MORTAR_`? (we can always change it)

There should be no ifdefs in isle and lego1: everything is a proper `MORTAR_` definition/function where SDL is an implementation detail.
This probably also means the new layer will have to translate the various SDL events to something generic.

It should be possible to add an extra mortar backend for a completely different backend (e.g. glfw3, raylib, sokol, win32, ...) without changing anything in isle and lego1.

A first step would be to move all SDL3 things from isle and lego1 to a mortar library. (the SDL3 headers *should* only be included by its private cpp/h files, and *must not* be present in its public h headers.)

madebr:
Here's an extra motivation for adding SDL3 support for the Wii (and attempting to upstream it):

The main reason upstream SDL does not have a Nintendo Switch support is because Icculus (one of the SDL main devs) signed a Nintendo NDA and does not want to get in trouble.

Nintendo Wii is a legacy platform, so this limitation does not longer apply there.

I'm not active in the Wii homebrew scene, so perhaps you can motivate people in doing this (and also sticking around to fix and advise on various Wii issues).

leap0x7b:
I disagree. I don't know how would that be any different than an making an SDL2 shim. After all, many platforms still only have SDL2 (or even SDL1), not just GameCube and Wii. Even if the PR author intended the shim to be used for a Wii port, it doesn't mean it couldn't be used for future ports. You don't need to port SDL3 first, just use the SDL2 shim, make some tweaks for the platform, and et voila. 

IMO, isle-portable shouldn't use SDL3 in the first place. I think it would do just fine with SDL2, even if isle-portable targeted modern computers first and legacy platforms second. You know what, I actually agree with the intermediate layer thing. ClassicCube and Scratch Everywhere already does it (even though the 99% of the ports on latter are literally just SDL2). It would take a very long time to decouple isle-portable from the SDL3isms though since it's extremely dependent on SDL3.

AJenbo:
Sdl3 made both 3d rendering and web port easier, it was the right choice in many ways

Helloyunho:
me personally don't know if it should've written in sdl2 from the beginning, but i do agree that for future ports having sdl2 layer is good

it's good if you port sdl3 to your desire console but that's too much work than just using existing sdl2 lib
and since (afaik) we're aiming for easy portability i think it's a 'why not' question to have a sdl2 layer

leap0x7b:
> me personally don't know if it should've written in sdl2 from the beginning, but i do agree that for future ports having sdl2 layer is good
> 
> it's good if you port sdl3 to your desire console but that's too much work than just using existing sdl2 lib and since (afaik) we're aiming for easy portability i think it's a 'why not' question to have a sdl2 layer

yeah that's what i'm saying

Covkie:
> me personally don't know if it should've written in sdl2 from the beginning, but i do agree that for future ports having sdl2 layer is good
> 
> it's good if you port sdl3 to your desire console but that's too much work than just using existing sdl2 lib and since (afaik) we're aiming for easy portability i think it's a 'why not' question to have a sdl2 layer

Honestly my exact thought process and why I chose this path. Both porting the SDL2 wii patches to SDL3 and replacing SDL3 with another layer of abstraction seems like wayy more work than what this shim is.

(I already have a working wii port with this shim.)

foxtacles:
The team has decided that a SDL2 shim should not be part of this project. The preferred direction to go would be to add SDL3 support for desired platforms - or further pursue this effort: https://github.com/isledecomp/isle-portable/pull/738

PR #711, opened 2025-09-12
Remove SDL3 workaround for missing mouse add events

foxtacles:
Fixed here: https://github.com/libsdl-org/SDL/issues/12815


foxtacles:
@Helloyunho do you know how to fix the iOS build?

Helloyunho:
> error: No simulator runtime version from ["22E238", "22F77", "22G86"] available to use with iphonesimulator SDK version 23A337

Huh, I thought I fixed that error

One thing you could try is running `xcodebuild -downloadPlatform iOS` right after when `select Xcode` is finished
You're probably fine to do this for only iOS build

PR #728, opened 2025-11-02
Fix build errors on macOS 26.

calmsacibis995:
This pull request affects the OpenGL ES 2.x and 3.x rendering code.

- Adds some `#if` statements for macOS.
- Adds macOS-specific macros that point to macOS-equivalent OpenGL functions.
- Adds macOS-specific macros that point to two OpenGL macros, since OpenGL ES is not supported.

### DIFF
--- a/miniwin/src/d3drm/backends/opengles2/renderer.cpp
+++ b/miniwin/src/d3drm/backends/opengles2/renderer.cpp
@@ -3,7 +3,7 @@
 
 #if defined(__APPLE__)
 #include <TargetConditionals.h>
-#if defined(TARGET_OS_IOS)
+#if TARGET_OS_IOS
 #include <OpenGLES/ES2/gl.h>
 #include <OpenGLES/ES2/glext.h>
 #else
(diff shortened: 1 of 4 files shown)


miniwin/src/d3drm/backends/opengles2/renderer.cpp
```diff
@@ -1,8 +1,15 @@
 #include "d3drmrenderer_opengles2.h"
 #include "meshutils.h"
 
+#if defined(__APPLE__)
```
AJenbo, line 4: should this be made macOS specific since iOS does support ES?

https://developer.apple.com/library/archive/documentation/3DDrawing/Conceptual/OpenGLES_ProgrammingGuide/AdoptingOpenGLES3/AdoptingOpenGLES3.html

calmsacibis995: Sure, but I haven't tested this code on iOS, only on macOS.

calmsacibis995: OK, now it is solved. macOS and iOS builds should build correctly now.

PR #735, opened 2025-11-21
libweaver: support system library and use latest wip read flags

madebr:
- Update SI loader extension to use the latest WIP libweaver flags PR (https://github.com/isledecomp/SIEdit/pull/37)
- Support using a system libweaver library (https://github.com/isledecomp/SIEdit/pull/38):this allows me to build/install libweaver once and use it in isle-portable. It also makes it easier to experiment with changes in 3rd party libraries: you edit/build it in a separate IDE instance, and it is automatically used by isle-portable without going through git etc.

### DIFF
--- a/3rdparty/CMakeLists.txt
+++ b/3rdparty/CMakeLists.txt
@@ -52,8 +52,8 @@ if(DOWNLOAD_DEPENDENCIES)
     include(FetchContent)
     FetchContent_Populate(
         libweaver
-        URL https://github.com/isledecomp/SIEdit/archive/ae447259389f3bf8273c7e7a4844743faf7cbdb8.tar.gz
-        URL_MD5 dee68424fde8db6d5cef3b9034a8151f
+        URL https://github.com/isledecomp/SIEdit/archive/afd4933844b95ef739a7e77b097deb7efe4ec576.tar.gz
+        URL_MD5 59fd3c36f4f380f730cd9bedfc846397
     )
     add_library(libweaver STATIC
         ${libweaver_SOURCE_DIR}/lib/core.cpp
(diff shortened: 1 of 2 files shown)


madebr:
(ci is failing because of missing https://github.com/isledecomp/SIEdit/pull/38)

The git hash of SIEdit needs to be updated after a merge, so marking this PR as draft.

gwuen:
This seems to break building with `libweaver` as a Git submodule, unfortunately:

```
CMake Error at CMakeLists.txt:139 (find_package):
  By not providing "Findlibweaver.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "libweaver", but CMake did not find one.

  Could not find a package configuration file provided by "libweaver" with
  any of the following names:

    libweaverConfig.cmake
    libweaver-config.cmake

  Add the installation prefix of "libweaver" to CMAKE_PREFIX_PATH or set
  "libweaver_DIR" to a directory containing one of the above files.  If
  "libweaver" provides a separate development package or SDK, be sure it has
  been installed.


-- Configuring incomplete, errors occurred!
```

Was this intended?

madebr:
We currently have a `DOWNLOAD_DEPENDENCIES` variable.
For SDL3, it means use `FetchContent_Declare` when true, else use `find_package`
For miniaudio, use `FetchContent_Declare` when true, else use the git submodule

For libweaver, this used to mean use `FetchContent_Declare` when true else use the git submodule.

Now, it means use `FetchContent_Declare` when true, else use `find_package`.

Unless we make `DOWNLOAD_DEPENDENCIES` a tristate variable, the new state makes more sense imho and libweaver as a git submodule can be removed.

(This is the same behavior as iniparser)

@foxtacles  thoughts?

foxtacles:
> the new state makes more sense imho and libweaver as a git submodule can be removed.

yes

Link4Electronics:
It's not possible to package anymore because of this `DOWNLOAD_DEPENDENCIES` change or install `isle-portable` form AUR. It tries to replace `SDL3` libs from system. I was trying to make an appimage for `aarch64` (more specifically postmarketOS which uses `musl` instead of `glibc`)
```
Packages (1) isle-portable-git-r2444.0f96ee20-1

Total Installed Size:  7.07 MiB

:: Proceed with installation? [Y/n] 
checking keyring...
checking package integrity...
loading package files...
checking for file conflicts...
error: failed to commit transaction (conflicting files)
isle-portable-git: /usr/lib/libSDL3.so exists in filesystem (owned by sdl3)
isle-portable-git: /usr/lib/libSDL3.so.0 exists in filesystem (owned by sdl3)
Errors occurred, no packages were upgraded.
Error: Process completed with exit code 1.
```
Reverting `libweaver` as submodule commits can compile and package it.

PR #738, opened 2025-11-27
Add mortar layer

madebr:


madebr:
This pr is ready for testing.
I only need to change one thing before this is ready for merging: limit the number of SDL3 symbols that are dynamically loaded.

madebr:
I'm marking this ready for review.
It needs a thorough review + testing before considered ready for merging.

CompSciDude1:
What's mortar?

leap0x7b:
the graphics backend layer that sits between the windowing library (miniwin) and the renderer (opengl, directx, etc)

leap0x7b:
we're using sdl3 as of right now (which is a bad idea imo should've used sdl2) though the mortar thing will allow you to use other backends

madebr:
> What's mortar?

The motivation of an extra layer is platforms that do not (yet) have SDL3 support, or don't even have any SDL support.
Please read the thread starting from https://github.com/isledecomp/isle-portable/pull/707#pullrequestreview-3378315904
This extra layer (with some, but minimal overhead)  allows isle-portabl to be built on top of any platform.


> we're using sdl3 as of right now (which is a bad idea imo should've used sdl2) though the mortar thing will allow you to use other backends

sdl2 is in maintenance mode and will never ever get additional features.
Perhaps [sdl2-compat](https://github.com/libsdl-org/sdl2-compat) provides too good compatibility.

leap0x7b:
> sdl2 is in maintenance mode and will never ever get additional features. Perhaps [sdl2-compat](https://github.com/libsdl-org/sdl2-compat) provides too good compatibility.

Does it really matter if it's in maintenance mode and doesn't get new features? A bunch of projects hasn't switched to SDL3 yet, some even still using SDL1. And when they do add SDL3 support, SDL2 support is usually kept for older platforms. I don't think using SDL2 over SDL3 would make a big of a difference other than the lack of a native GPU API.

All that rant and my remarks of SDL3 on my original comment is just my opinion. Can we please not turn this into a whole drama?

madebr:
> All that rant and my remarks of SDL3 on my original comment is just my opinion. Can we please not turn this into a whole drama?

Opinions do not belong in GitHub discussions. Only technical things are discussed here.
You can vent your opinion on discord or in the matrix channel.

Rhlp-Engineering:
what is a mortar layer?

Rhlp-Engineering:
> > What's mortar?
> 
> The motivation of an extra layer is platforms that do not (yet) have SDL3 support, or don't even have any SDL support. Please read the thread starting from [#707 (review)](https://github.com/isledecomp/isle-portable/pull/707#pullrequestreview-3378315904) This extra layer (with some, but minimal overhead) allows isle-portabl to be built on top of any platform.
> 
> > we're using sdl3 as of right now (which is a bad idea imo should've used sdl2) though the mortar thing will allow you to use other backends
> 
> sdl2 is in maintenance mode and will never ever get additional features. Perhaps [sdl2-compat](https://github.com/libsdl-org/sdl2-compat) provides too good compatibility.

oh sorry that i didnt see this.

foxtacles:
@madebr could you rebase this one more time? I'll do some testing this weekend

madebr:
> @madebr could you rebase this one more time? I'll do some testing this weekend

Done.

When testing and reviewing the code, make sure to pay attention to the mortar surfaces and palettes in `sdl3_surface.cpp`, palettes in particular as these depend on reference counting.  I think I got it right but it's always good to have extra eyes on it.

Please make sure your texture loader extension still functions properly, as it was present in the traces of https://github.com/isledecomp/isle-portable/pull/775.

This mortar thing is missing proper testing that you intend to do coming weekend.

extensions/src/textureloader.cpp
```diff
@@ -46,26 +46,26 @@ bool TextureLoader::PatchTexture(LegoTextureInfo* p_textureInfo)
...
 
 	MxU8* dst = (MxU8*) desc.lpSurface;
-	Uint8* srcPixels = (Uint8*) surface->pixels;
+	uint8_t* srcPixels = (uint8_t*) surface->pixels;
 
 	if (details->bits_per_pixel == 8) {
-		SDL_Palette* sdlPalette = SDL_GetSurfacePalette(surface);
+		MORTAR_Palette* sdlPalette = MORTAR_GetSurfacePalette(surface);
 		if (!sdlPalette) {
 			p_textureInfo->m_surface->Unlock(desc.lpSurface);
```
foxtacles, line 67: The texture loader fails for some textures here because `sdlPalette` (`mortar_sdl_surface->mortar_palette`) is `NULL`. Couldn't figure out yet why that is.

foxtacles: https://github.com/madebr/isle-portable/pull/1

foxtacles:
Viable proof of concept but needs further testing and validation.

PR #740, opened 2025-12-11
implement WASD control option

Ramen2X:
This PR implements a toggleable setting to use WASD for navigation instead of the arrow keys, like in Rebuilder (and modern PC games in general). This setting is disabled by default.

I've also implemented the toggle in the Qt CONFIG application, and I intend to add it to the isle.pizza frontend post-merge as well.

Naming, formatting and implementation feedback is welcome as always!
image


Ramen2X:
The Linux (Debug) CI pass seems to be failing due to a 504 when trying to get the AppImage runtime, doesn't seem related to this PR.

AJenbo:
Nice :)

PR #758, opened 2025-12-31
MxSoundManager: use buffer size and don't rely on buffer capacity

madebr:
This fixes a detect_container_overflow address sanitizer error.

### DIFF
--- a/LEGO1/omni/src/audio/mxsoundmanager.cpp
+++ b/LEGO1/omni/src/audio/mxsoundmanager.cpp
@@ -149,7 +149,6 @@ void MxSoundManager::AudioStreamCallback(
 {
 	static vector<MxU8> g_buffer;
 	if (p_additionalAmount > g_buffer.size()) {
-		// We resize and don't reserve memory to avoid an AddressSanitizerContainerOverflow asan error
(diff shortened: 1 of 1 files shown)


LEGO1/omni/src/audio/mxsoundmanager.cpp
```diff
@@ -148,7 +148,10 @@ void MxSoundManager::AudioStreamCallback(
 )
 {
 	static vector<MxU8> g_buffer;
-	g_buffer.reserve(p_additionalAmount);
+	if (p_additionalAmount > g_buffer.size()) {
+		// We resize and don't reserve memory to avoid an AddressSanitizerContainerOverflow asan error
```
foxtacles: I think we can do without the comment - it's just a bug we are fixing

foxtacles (approved):

LEGO1/omni/src/audio/mxsoundmanager.cpp
madebr:
```suggestion
```
Aight, I added it to avoid the next hacker re-adding the `.reserve()` optimization to avoid memory clearing.

PR #779, opened 2026-03-01
Fix AppRun script for AppImage execution

Damglador:
Right now when path to AppImage has spaces in it (for example if it gets installed in ~/Applications by a manager I'll be named "Isle Portable") I'll fail to execute with an error:
```
/tmp/.mount_Isle PCAHbep/AppRun: line 16: [: /home/damglador/Applications/Isle: binary operator expected
/tmp/.mount_Isle PCAHbep/AppRun: line 24: /tmp/.mount_Isle PCAHbep/usr/bin/isle-config: Input/output error
```

This patch should fix it.


Damglador:
```
/tmp/.mount_Isle PCAHbep/AppRun: line 24: /tmp/.mount_Isle PCAHbep/usr/bin/isle-config: Input/output error
```
Is apparently caused by isle-config somehow being 0 bytes

PR #799, opened 2026-05-01
DOS port

AJenbo:
Glorious mode 13h, 8bit color at 320x200, this is the same as Quake!!!
Screenshot From 2026-05-01 23-15-02

With some help from @madebr we now DOS support. I added an 8bit renderer which might also be helpful for other low spec devices like the PSP.

- fix a use after free issue
- turn off software mouse during credits to avoid HOM


LEGO1/lego/legoomni/src/worlds/infocenter.cpp
```diff
@@ -1499,7 +1499,7 @@ void Infocenter::StartCredits()
 	GetViewManager()->RemoveAll(NULL);
 
 	InvokeAction(Extra::e_opendisk, *g_creditsScript, CreditsScript::c_LegoCredits, NULL);
-	SetAppCursor(e_cursorArrow);
+	SetAppCursor(e_cursorNone);
```
foxtacles: Is this needed for DOS support?

foxtacles:
Seeing this glitch on emscripten - something to do with removal of flags in `MxDisplaySurface`?

https://github.com/user-attachments/assets/9d278746-c8b2-49d1-be8d-49db6f3e8011

LEGO1/lego/legoomni/src/worlds/infocenter.cpp
AJenbo: No, it's needed for all systems that use the build in cursor.

foxtacles: This changes the active cursor from the regular arrow to the "loading" spinner though while the credits are playing. I'm not sure I follow why this is needed.
(see the video below)

AJenbo: image

foxtacles: In that case the fix should be something more targeted because this changes the cursor for everyone, even those who use the system cursor - and they now permanently see the "loading" cursor (see below) on the credits screen which is awkward:
 
image

AJenbo: Fixed

AJenbo:
Fixed

AJenbo:
The Vita build error appears to be unrelated.

foxtacles:
Seem to be getting this crash on startup when running with DX5

image

AJenbo:
Hm give it another try now

foxtacles:
> Hm give it another try now

Same error
(confirmed it's a regression in this PR, `master` works fine)

AJenbo:
Really hard to address since I can't build or run it, if you are able to bisect the code that would be very helpful :/

foxtacles:
This patch brings back working dx5 (needs 32bit palette allocated, and the source color key)

[surface_dx5.patch](https://github.com/user-attachments/files/27307330/surface_dx5.patch)

This is mostly reverting the changes you applied to `mxdisplaysurface.cpp`. How to square this with DOS support I don't know, but this is needed for dx5

AJenbo:
Oh I see it now, I flipped the case when merging  an `#ifdef`, it probably would be fine with all that stuff enabled on miniwin, but I'll just fix it and leave both sides minimally changed. Miniwin just needs the palettes always set since it allows for paletted rendering to and RGB window surface.

The color key at the bottom is stuff I introduced in https://github.com/isledecomp/isle-portable/commit/77c832c237fe90c5bc2793c6349d2f52c7502bc6 it's just luck that it hasn't been causing bad renderings as it's for the solid video overlays that shouldn't actually be keyed.

Please give it another try.

foxtacles:
Still seeing these glitches

https://github.com/user-attachments/assets/27bf930f-6bb0-4604-a9ac-04c7150baa75

To fix that I needed to put the source color key stuff back as well

AJenbo:
Got it, videos uses a different palette :)

foxtacles (approved):
Nice work!

madebr:
1. Using this ini:
<details>

<summary>isle.ini</summary>

```ini
[isle]
diskpath                       = "D:"
cdpath                         = "D:"
mediapath                      = "D:"
savepath                       = "C:"
flip surfaces                  = "false"
full screen                    = "false"
wide view angle                = "true"
3dsound                        = "true"
music                          = "true"
usejoystick                    = "true"
joystickindex                  = "0"
draw cursor                    = "false"
back buffers in video ram      = "-1"
island quality                 = "2"
island texture                 = "1"
max lod                        = "3.600000"
max allowed extras             = "20"
transition type                = "3"
3D Device ID                   = "0 0x682656F3 0x00000000 0x00000000 0x07000000"
Active in Background           = "true"
```
</details>

2. Starting dosbox as `dosbox -c "mount D path/to/lego/island/game/data" .` (from the build directory

3. And starting the game as `C:\isle.exe --ini C:\isle.ini`


I took this screenshot in the information center:
image
It looks like the window has 320x200, but all is rendered unscaled as 640x480.
Because I cannot click on the book, I cannot enter the island.

madebr (commented):
Got it working!

The problem was `full screen = "false"` in my ini.

Image

madebr (approved):

AJenbo:
I tried running it on real hardware but my machine is being stubborn and only wanting to see 528MB of my disk which isn't enough to get things running.

denilsonsa:
@AJenbo, please don't forget to add a mention to DOS at the`README.md`, and at anywhere else where there is a list of ports. (e.g. maybe somewhere in the wiki)

AJenbo:
Feel free to make a PR to Fix that

PR #800, opened 2026-05-02
Voodoo 3dfx support

AJenbo:
For now the targeted card has to be specified at build time:
```
cmake -DGLIDE_HW=sst1
```

| id |  cards |
|-----|--------|
| sst1 |  Voodoo Graphics |
| sst96 | Voodoo Rush |
| cvg |   Voodoo 2 |
| h3 |   Voodoo Banshee/Voodoo 3 |
| h5 |    Voodoo Voodoo 4/5 |

Much nicer then the software renderer:
image

(tested with Voodoo 1 and Voodoo 3 emulation)


foxtacles (approved):

AJenbo:
Just noticed I forgot:
- Detect if Voodoo is present else fall back to software
- Evict textures to avoid running out of memory on the 4MB VRAM.

AJenbo:
I don't have my Voodoo 2 anymore but this is from the Voodoo enabled build with auto detection:

https://github.com/user-attachments/assets/3cdd5d23-8812-4ee6-bb21-8b47dc07eceb

From testing on PCem it looks like building it for Voodoo 1 lets it run on all cards, probably the other options are just optimizations that exclude some of the older ones.

foxtacles (approved):

PR #808, opened 2026-05-17
ci: suffix debug-build artifacts to avoid release filename collision

### LINKED ISSUE #646: Corrupted Release zip for Windows x64
I'm not sure exactly when it started, unfortuantly, but for at least a week or so every isle-0.1-Windows-x64.zip I have downloaded from the releases has had CRC errors and header errors.

Here's a couple of examples from the most recent release and the one just before. There doesn't seem to be much of pattern as to what part has errors.

Current build:
Image

Penultimate build:
Image

foxtacles:
`actions/download-artifact@v8` with `merge-multiple: true` extracts every artifact into the same Release/ directory in parallel. Release and Debug builds of the same platform produced identical CPack/AppImage filenames (isle-0.1-Windows-x64.zip, isle-0.1-Linux-x64.tar.gz, Isle_Portable-x86_64.AppImage), so the parallel writes raced on the same path and intermittently produced corrupt archives in the release.

Append `-debug` to CPACK_PACKAGE_FILE_NAME when ISLE_DEBUG is set, and rename the AppImage in the same case, so each matrix entry contributes a unique filename to the release.

Fixes #646


madebr:
Is `ISLE_DEBUG` still needed as  a CMake option/item in the ci array?
`ISLE_DEBUG` is only used for selecting another citro3d library on nintendo 3ds.

I'd get rid of it, and perhaps add a `ISLE_CITRO3D_DEBUG` option.

madebr (approved):

PR #812, opened 2026-07-05
Fix softlocked cops stuck on donut

### LINKED ISSUE #586: Police don't behave correctly with doughnuts
They often will not see the first one but need a second one before they see it which feels like an off by one error.

They can also easily get stuck on top of a doughnut, like they simultaneously think they picked it up but then realize they didn't spinning around so fast that they just flicker on screen. They never seem to break out of this.

I have tired but where unable to reproduce this on the original or decomp.

AJenbo:
This fixes a bug that can cause the cops or The Brikster to get stuck in act 3 and often leaves me to loose the game.

It's a bug in the original game and seems pretty clearly a bug and not intentional.

I did extensive searching for other issues so with this fixed i think we can close https://github.com/isledecomp/isle-portable/issues/586

### DIFF
--- a/LEGO1/lego/legoomni/src/actors/act3actors.cpp
+++ b/LEGO1/lego/legoomni/src/actors/act3actors.cpp
@@ -378,6 +378,19 @@ MxResult Act3Cop::FUN_10040360()
 				Mx3DPointFloat localec(local88);
 				localec -= local2c;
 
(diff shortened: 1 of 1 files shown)


foxtacles (approved):
Thanks, that looks correct to me!

Please close the original issue after the merge (assuming you verified that it fixed it)

PR #817, opened 2026-07-25
Add alternate-take voice masters found by the final audit

foxtacles:
Follow-up to #816: a completeness audit that checks every unreplaced object whose embedded source filename exists in the dump at higher quality — independent of acoustic correlation, which is what let these slip past the original pipeline (different takes correlate poorly by waveform).

**21 alternate-take voice lines**: Act3 Brickster taunts, the build instructor introductions, and several Infomaniac/character lines where the 22kHz master is a different read of the same line. Masters that outrun the shipped scope are cut to it (e.g. the 54s IIC017IN master cut to the shipped 25.2s window) so paired animations stay in sync; standalone lines keep the master's full length.

**Evaluated and rejected — Papa's bio narration**: the only one of the five Infocenter bios that shipped below master quality (11kHz/8-bit; the other four ship the 22kHz masters byte-for-byte). The dump's `Papafin.wav` is 22kHz and exactly movie-length, but phrase-level analysis shows it is a different performance whose timing drifts up to 3.7s from the take the Papa_Smk animation was authored against — unfixable without voice/music separation, since both takes sit on a continuous music bed. Sync wins; the retail track stays.

Pack grows to 981 replacements; full parity, duration and boundary verification pass. Remaining unreplaced audio is confirmed to be per-cutscene Foley mixdowns with no dump master, same-quality matches with no gain, and the Build/Race UI sounds whose 22kHz dump versions are old-generation designs rather than quality upgrades (deliberately skipped).

### DIFF
--- a/assets/hdaudio/index.txt
+++ b/assets/hdaudio/index.txt
@@ -220,6 +220,7 @@ Act3/ACT3.SI	566	wav/offscreen/cl/sns63xcl.wav
 Act3/ACT3.SI	567	wav/offscreen/en/sns64xen.wav
 Act3/ACT3.SI	568	wav/offscreen/re/sns65xre.wav
 Act3/ACT3.SI	569	wav/offscreen/sl/sns66xsl.wav
+Act3/ACT3.SI	570	wav/offscreen/ml/sns67xml.wav
 Act3/ACT3.SI	571	wav/offscreen/bu/sns68xbu.wav
 Act3/ACT3.SI	572	wav/offscreen/sn/sns69xsn.wav
 Act3/ACT3.SI	573	wav/offscreen/ni/sns70xni.wav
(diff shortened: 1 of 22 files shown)


GordonFreakman:
Where did the HD audio dump come from?

foxtacles:
> Where did the HD audio dump come from?

I don't know the exact source, I believe it's been passed down from one of the original developers.
Here's a copy: https://drive.google.com/file/d/1hqjWFONDE9cdNWF69D7S6fUOZFmd8SXv/view?usp=drive_link
Maybe this among some other things will be pushed to archive.org / similar at some point.

GordonFreakman:
> > Where did the HD audio dump come from?
> 
> I don't know the exact source, I believe it's been passed down from one of the original developers. Here's a copy: https://drive.google.com/file/d/1hqjWFONDE9cdNWF69D7S6fUOZFmd8SXv/view?usp=drive_link Maybe this among some other things will be pushed to archive.org / similar at some point.

Awesome! Thank you so much!

PR #834, opened 2026-08-27
Fix list corruption in MxDSObjectList::FindInternal

southcitycapture:
Hardening, not a crash fix: with `unknown24 == -3`, `FindInternal` deliberately scans the whole list for the last match, so the loop iterator sits at `end()` after the loop — and the erase would erase `end()`, which is UB on `std::list`. Unreachable with current callers (the only `-3` writer wraps a non-deleting `Find` and restores the value immediately); the change tracks the found node's iterator and erases that instead, so behavior is identical on every reachable path.

### DIFF
--- a/LEGO1/omni/src/action/mxdsobject.cpp
+++ b/LEGO1/omni/src/action/mxdsobject.cpp
@@ -180,6 +180,7 @@ MxDSObject* MxDSObjectList::FindInternal(MxDSObject* p_action, MxBool p_delete)
 	// DECOMP ALPHA 0x1008b99d ?
 
 	MxDSObject* found = NULL;
+	iterator foundIt;
 
 #ifdef COMPAT_MODE
 	iterator it;
@@ -191,6 +192,7 @@ MxDSObject* MxDSObjectList::FindInternal(MxDSObject* p_action, MxBool p_delete)
 			if (p_action->GetUnknown24() == -2 || p_action->GetUnknown24() == -3 ||
 				p_action->GetUnknown24() == (*it)->GetUnknown24()) {
 				found = *it;
+				foundIt = it;
 				if (p_action->GetUnknown24() != -3) {
 					break;
 				}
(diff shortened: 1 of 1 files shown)


foxtacles:
The code change is correct and behavior-identical on every reachable path, so I'm fine taking it — but it's hardening, not a crash fix, and the description doesn't hold up.

`erase(end())` needs `p_action->GetUnknown24() == -3` **and** `p_delete == TRUE`. The only writer of `-3` in the tree is `mxramstreamcontroller.cpp:45`, and it wraps a `Find()` (`p_delete == FALSE`), restoring the value four lines later. Every `FindAndErase` call site passes `-1`, `-2`, or `0..N`, and `-2` breaks out of the loop on first match. So the buggy line is dead code here.

The trace also can't be from this path: `MxDSStreamingAction : MxDSAction`, while `MxDSParallelAction : MxDSMultiAction` — `~MxDSStreamingAction` can't call `~MxDSParallelAction`, they're sibling branches.

Two notes on the framing, since this looks like it came out of your PowerPC port:

- the "vendored HP STL" isn't what's building here — `mxstl/stlcompat.h` only pulls `mxstl.h` for `_MSC_VER <= 4.20`, and that header just wraps MSVC's own `<list>`. Everywhere else it's `std::list`. `erase(end())` is UB in both, so the change is still fine, just not for that reason
- if your port genuinely reaches the `-3` path, that means it has a `FindAndErase` caller this tree doesn't — worth pointing at it, because that caller is the actual bug and it'd be in your fork, not here

Please rewrite the description to say what this is ("unreachable with current callers, hardening against `erase(end())`") — it becomes the squash-merge message — and drop the comment.

southcitycapture:
Reworked then. The description now calls it hardening, the code comment is gone, and the commit message matches (it's the squash message).

On the trace: the nesting itself is coherent (`~MxDSStreamingAction` deletes its owned internal action, which can be a `MxDSParallelAction`, so the destructors appear nested without being an inheritance chain) but your conclusion stands anyway. Rechecking the timeline, that crash predates a mesh-indexing fix in my port that had been reading garbage heap all over; the corruption almost certainly came from there, not from this path. The `-3` audit and the stlcompat guard are both as you say, I'd pattern-matched `mxstl.h` without reading the `_MSC_VER` gate. Thanks for the thorough read.

PR #835, opened 2026-08-27
Guard PutFrame against a presenter with no surface and no bitmap

southcitycapture:
`MxDisplaySurface::VTable0x44` can return NULL (GetDisplayMode, Lock, or allocation failure), and `MxStillPresenter::LoadFrame` frees `m_frameBitmap` regardless of that outcome — so a presenter can end up with neither a surface nor a decoded frame, and every branch of `MxVideoPresenter::PutFrame` then dereferences one of them. Skip the frame in that state, and log the reason on `VTable0x44`'s three failure paths, which previously failed silently into exactly that crash.

(Originally this PR also cleared the display-mode masks before the 8-bit surface requests and logged rejections in `DirectDrawImpl::CreateSurface` — dropped per review: SDL3 maps any bpp=8 mask combination to `INDEX8`, so those paths are unreachable in this tree. The failure I chased was my SDL2-subset shim keeping SDL2's stricter mask mapping; fixing the shim to match SDL3's semantics is the real cure on my side.)

### DIFF
--- a/LEGO1/omni/src/video/mxdisplaysurface.cpp
+++ b/LEGO1/omni/src/video/mxdisplaysurface.cpp
@@ -8,6 +8,7 @@
 #include "mxutilities.h"
 #include "mxvideomanager.h"
 
+#include <SDL3/SDL_error.h>
 #include <SDL3/SDL_log.h>
 #include <assert.h>
 #ifdef MINIWIN
@@ -716,6 +717,7 @@ LPDIRECTDRAWSURFACE MxDisplaySurface::VTable0x44(
 	ddsd.dwSize = sizeof(ddsd);
 
 	if (draw->GetDisplayMode(&ddsd)) {
+		SDL_Log("MxDisplaySurface::VTable0x44: GetDisplayMode failed");
 		return NULL;
 	}
 
@@ -748,6 +750,9 @@ LPDIRECTDRAWSURFACE MxDisplaySurface::VTable0x44(
 		else {
 			surface = NULL;
 		}
+		if (!surface) {
+			SDL_Log("MxDisplaySurface::VTable0x44: CreateSurface failed: %s", SDL_GetError());
+		}
 	}
 
 	if (surface) {
@@ -778,6 +783,7 @@ LPDIRECTDRAWSURFACE MxDisplaySurface::VTable0x44(
 		ddsd.dwSize = sizeof(ddsd);
 
 		if (surface->Lock(NULL, &ddsd, DDLOCK_WAIT | DDLOCK_WRITEONLY, 0) != DD_OK) {
+			SDL_Log("MxDisplaySurface::VTable0x44: Lock failed: %s", SDL_GetError());
 			surface->Release();
 			surface = NULL;
 		}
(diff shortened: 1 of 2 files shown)


foxtacles:
The mask leak is real, but the failure it's blamed for can't happen on SDL3 — and that difference looks like it's coming from your fork.

SDL3 (`SDL_pixels.c:481-488`):

```c
case 8:
    if (Rmask == 0xE0 && Gmask == 0x1C && Bmask == 0x03 && Amask == 0x00)
        return SDL_PIXELFORMAT_RGB332;
    return SDL_PIXELFORMAT_INDEX8;      // anything else
```

SDL2 (`SDL_MasksToPixelFormatEnum`, `SDL_pixels.c:310-320`):

```c
case 8:
    if (Rmask == 0) return SDL_PIXELFORMAT_INDEX8;
    if (Rmask == 0xE0 && ...) return SDL_PIXELFORMAT_RGB332;
    break;                              // -> SDL_PIXELFORMAT_UNKNOWN
```

So the leaked `00ff0000/0000ff00/000000ff` masks give `UNKNOWN` on SDL2 and `INDEX8` on SDL3 — I ran it against our SDL3 to confirm. `DDERR_INVALIDPIXELFORMAT` is unreachable at bpp=8 here, and the log line you quoted is the text this PR itself adds, so it had never printed.

That points at the SDL2-subset shim in your PowerPC port: the fix that actually unblocks you is making it match SDL3's semantics (8bpp → `INDEX8` unless exact RGB332), which is the contract the rest of this tree is written against. Worth auditing the shim for other places it kept SDL2 behavior that SDL3 changed.

For this PR, we'd rather not carry changes that don't fix or improve something here, so:

- **drop both `memset`s** — under SDL3 the masks are read and then ignored, so they change nothing
- **drop the `bpp=%d masks=...` log** in `DirectDrawImpl::CreateSurface`, and the WIDTH/HEIGHT one — I can't find a caller that reaches either: every site that sets `DDSD_PIXELFORMAT` either hardcodes bpp=8 (always `INDEX8` on SDL3) or propagates a real surface's format, which round-trips
- **drop the `SDL_Log` inside `PutFrame`** — it runs per frame, so a stuck presenter spams every frame
- **keep the `PutFrame` guard** — that one earns its place for a different reason than the description gives: `VTable0x44` can still return NULL (GetDisplayMode / Lock / allocation failure) and `mxstillpresenter.cpp:86-96` deletes `m_frameBitmap` regardless, so both-NULL is genuinely reachable
- the three failure logs in `VTable0x44` are the diagnostics worth keeping, if any: they sit on paths that can actually fail and currently fail silently into exactly that crash

And please rewrite the description — the CreateSurface rejection isn't the cause here.

southcitycapture:
Reworked as suggested: the memsets, the CreateSurface logs, and the per-frame PutFrame log are gone; what's left is the guard (with your justification, which is the correct one, thanks for spelling out the reachable path) and the three VTable0x44 failure logs. Title, description, and commit message updated to match.

You called the shim thing pretty spot on there mate: my `SDL_GetPixelFormatForMasks` was forwarding to SDL 2.0.3's `SDL_MasksToPixelFormatEnum` semantics, where bpp=8 with nonzero masks is UNKNOWN. Fixing the shim to return INDEX8 there (SDL3's contract) and auditing it for other SDL2-isms it should be papering over.

PR #836, opened 2026-08-27
Bounds-check triangle indices in FlattenSurfaces

southcitycapture:
Skip (with a log) any triangle that references a vertex outside the group instead of reading out of bounds. Corrupt or truncated mesh data previously read past the vertex array; on some heap layouts that crashed inside the flatten loop. The short-circuit ordering also covers an index count that isn't a multiple of three.

(The mesh-builder half-unpacking hunk is dropped per review — it's only correct downstream of a byteswapping storage layer, which this tree doesn't have yet. It moves into the endianness series, where the swap and the unpacking land together.)

### DIFF
--- a/miniwin/src/internal/meshutils.cpp
+++ b/miniwin/src/internal/meshutils.cpp
@@ -2,6 +2,7 @@
 
 #include "mathutils.h"
 
+#include <SDL3/SDL_log.h>
 #include <tuple>
 #include <unordered_map>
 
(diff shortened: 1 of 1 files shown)


foxtacles:
The endian-dependence of the short-pair cast is real, but this change doesn't fix big-endian in *this* tree, because nothing byteswaps the data on the way in. `legolod.cpp:226` reads `polyIndices` raw, `LegoStorage::Read` is a raw byte copy, and there is no endian handling anywhere in the repo (`docs/*.ksy` all declare `endian: le`).

With file bytes `b0 b1 b2 b3`, the true values are `vertex = b1<<8|b0` and `normal|flag = b3<<8|b2`. On big-endian:

| | low | high |
|---|---|---|
| original | `b0<<8\|b1` — right half, bytes swapped | `b2<<8\|b3` — right half, bytes swapped |
| this PR | `b2<<8\|b3` — **wrong half** | `b0<<8\|b1` — **wrong half** |

Neither is correct, and this one also swaps the roles, so the bit-15 "new vertex" test ends up reading a bit of the vertex index.

The shift version is only right *downstream of a byteswap* — so I suspect your port already swaps `LegoU32` on read somewhere in its storage layer, and this hunk is correct against that, not against master. If so, the byteswapping layer is the thing to upstream (or keep in the fork), and this change can't land ahead of it: on its own it's a bit-identical no-op on LE and still broken on BE, with no BE target in CI to show either way. If your port *doesn't* byteswap, then it's still reading garbage and this doesn't fix that either.

Suggest splitting: drop the meshbuilder hunk, keep the `FlattenSurfaces` bounds check — that stands on its own and the short-circuit ordering is correct.

southcitycapture:
Split as you suggested, this is now just the FlattenSurfaces bounds check, and the meshbuilder hunk is out. Your table is right: my fork's storage layer byteswaps on read (ReadLE accessors over LegoStorage, LE-declared per the .ksy docs), so the shift version is correct there and meaningless here ahead of that layer. It'll come back as part of the endianness series so the swap and the unpacking land together.

PR #838, opened 2026-08-27
Make the SDL GPU and software renderer backends optional

southcitycapture:
Two new CMake options, `ISLE_RENDERER_SDLGPU` and `ISLE_RENDERER_SOFTWARE` (both ON by default, so existing builds are unchanged), let a platform without SDL_GPU or SDL_Renderer support build the remaining backends. Alongside that, the app only requests an OpenGL-capable window when a GL backend (opengl1, GLES2, or GLES3) is actually compiled in, so GL-less and headless builds (dummy video driver) can create their window.

Behavior note: for an existing MINIWIN build where `find_package(OpenGL)` and both GLES probes fail, this does change behavior — the window is no longer requested as OpenGL-capable where it previously was. That's the intent: such a build has no backend that could use the GL context.

Context: we're porting to big-endian PowerPC Mac OS X Tiger through an SDL3-subset compatibility layer over SDL 2.0.3, which has no SDL_GPU and no usable SDL_Renderer path — these options are what let the opengl1 + paletted backends build there, and they also enable headless runs under the dummy video driver.

### DIFF
--- a/ISLE/isleapp.cpp
+++ b/ISLE/isleapp.cpp
@@ -1006,7 +1006,8 @@ MxResult IsleApp::SetupWindow()
 	SDL_SetNumberProperty(props, SDL_PROP_WINDOW_CREATE_HEIGHT_NUMBER, g_targetHeight);
 	SDL_SetBooleanProperty(props, SDL_PROP_WINDOW_CREATE_FULLSCREEN_BOOLEAN, m_fullScreen);
 	SDL_SetStringProperty(props, SDL_PROP_WINDOW_CREATE_TITLE_STRING, WINDOW_TITLE);
-#if defined(MINIWIN) && !defined(__3DS__) && !defined(WINDOWS_STORE) && !defined(__vita__) && !defined(__DJGPP__)
+#if defined(MINIWIN) && (defined(USE_OPENGL1) || defined(USE_OPENGLES2) || defined(USE_OPENGLES3)) &&                  \
+	!defined(__3DS__) && !defined(WINDOWS_STORE) && !defined(__vita__) && !defined(__DJGPP__)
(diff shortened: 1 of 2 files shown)


Rhlp-Engineering:
woa your making such cool stuff!

foxtacles:
Unlike the others in this batch, this one isn't a fork-shim issue — making the backends optional is generally useful and the option itself is fine to land. One real gap though: the new condition omits `USE_OPENGLES3`.

`miniwin/CMakeLists.txt:63` registers it independently, and the probes differ — ES2 is `find_library(OPENGL_ES2_LIBRARY NAMES GLESv2)` while ES3 is `find_library(OPENGL_ES3_LIBRARY NAMES GLESv3 GLESv2)`. A system with GLESv3 and no GLESv2 gets ES3 without ES2, and `opengles3/renderer.cpp:77` calls `SDL_GL_CreateContext` directly, so it needs `SDL_WINDOW_OPENGL`. That build would silently lose its window.

```
#if defined(MINIWIN) && (defined(USE_OPENGL1) || defined(USE_OPENGLES2) || defined(USE_OPENGLES3)) && ...
```

Otherwise this looks right. The software backend is safe without the property — SDL3's GL renderer reconfigures the window itself (`SDL_render_gl.c:1855-1863`) — and both options default ON, so existing builds are unchanged.

Worth stating in the description: for an existing MINIWIN build where `find_package(OpenGL)` fails, this does change behavior (no GL window where there was one). That's the intent, just say so.

southcitycapture:
Good catch on ES3  a GLESv3-without-GLESv2 system would indeed have lost its window like a blind bat in a box bro. The condition now includes `USE_OPENGLES3`, and the description states the behavior change for GL-less MINIWIN builds for real .

PR #853, opened 2026-08-29
Render at native resolution on high-DPI displays

### LINKED ISSUE #846: Recognize when user is using HiDPI scaling
I have a 2560x1600@120Hz display, but the game thinks it can only run at 1280x800? It's not obvious if the game is running at full resolution, so it would be nice if the Linux port could make that clear.

foxtacles:
Fixes #846.

On Wayland (and macOS) with display scaling, SDL3 window sizes are in logical coordinates: a fullscreen window on a 2560x1600 display at 200% scaling gets a 1280x800 backing store that the compositor upscales, so the game really does render at a quarter of the native resolution. X11, Windows, and Android always deal in physical pixels, which is why this was never visible there.

The fix requests `SDL_WINDOW_HIGH_PIXEL_DENSITY` at window creation. The render pipeline already sizes itself from `SDL_GetWindowSizeInPixels()` and `SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED`, so it picks up the full-resolution backing store with no further changes. Mouse events, however, arrive in window coordinates while the viewport transform lives in pixel space, so `ConvertEventToRenderCoordinates` now scales event coordinates by `SDL_GetWindowPixelDensity()` (and `ConvertRenderToWindowCoordinates` divides by it for the joystick-driven cursor warp). Finger events are normalized and multiplied by the pixel size, so they were already correct. `Resize()` also logs the effective render resolution now, since the issue notes it's not obvious whether the game runs at full resolution.

Per SDL's high-DPI docs the pixel density is always 1.0 on Windows, X11, and Android, making this a no-op there. The flag is not set for Emscripten: the web port manages canvas resolution itself through `emscripten_enter_soft_fullscreen` with `EMSCRIPTEN_FULLSCREEN_CANVAS_SCALE_HIDEF`, which already multiplies by `devicePixelRatio`.

Tested:
- macOS built-in Retina display (2x): a 640x480 window now logs `Render resolution: 1280x960` (previously 640x480) and the game plays normally; on a 1x external monitor behavior is unchanged (density 1.0).
- Web build in headless Chrome with `--force-device-scale-factor=2` against the isle.pizza frontend, confirming Emscripten needs no change: "Original (640x480)" renders 640x480 as designed, and "Maximum" already renders at full native resolution (`Render resolution: 2168x1626` for a 1084x813 CSS canvas at DPR 2, and 1084x813 at DPR 1).

https://claude.ai/code/session_017CVAycW6CZioD9CXSb1AS4

### DIFF
--- a/ISLE/isleapp.cpp
+++ b/ISLE/isleapp.cpp
@@ -1006,6 +1006,9 @@ MxResult IsleApp::SetupWindow()
 	SDL_SetNumberProperty(props, SDL_PROP_WINDOW_CREATE_HEIGHT_NUMBER, g_targetHeight);
 	SDL_SetBooleanProperty(props, SDL_PROP_WINDOW_CREATE_FULLSCREEN_BOOLEAN, m_fullScreen);
 	SDL_SetStringProperty(props, SDL_PROP_WINDOW_CREATE_TITLE_STRING, WINDOW_TITLE);
+#ifndef __EMSCRIPTEN__
+	SDL_SetBooleanProperty(props, SDL_PROP_WINDOW_CREATE_HIGH_PIXEL_DENSITY_BOOLEAN, true);
+#endif
 #if defined(MINIWIN) && (defined(USE_OPENGL1) || defined(USE_OPENGLES2) || defined(USE_OPENGLES3)) &&                  \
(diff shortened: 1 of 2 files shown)


Rhlp-Engineering:
does this make joystick movements work better?

PR #878, opened 2026-08-31
Fix 3DS cursor transparency, distant texture corruption, and d-pad movement

### LINKED ISSUE #877: 3ds port : messed up cursor, strange colors all over, and broken controlls.
[2026-08-30_19-57-05.927_bot.bmp](https://github.com/user-attachments/files/31623365/2026-08-30_19-57-05.927_bot.bmp)

Image

heya everyone! rose here!
uh, im not even shure what is going on here.
but i booted up the newest version of lego island, and the 3ds circlepad now works wich is nice!
however exept for touchscreen, there is no way to move. maby having the dpad be wasd could be nice?

anywho, as you can see, something is wrong. the cursor is wrong and not transparent.
it's also a bit odd to move, but with the fps thing it works much better!

and lastly,.. uh, the colors. i dont..
someone else can work on that part because i dont know where to even start on that one gheebers.

thank you for your time, and have a lovley day!
i love to see this thing get better and better! :) 

~~ Rose!Image

foxtacles:
Closes #877. All three problems reproduced and verified in Azahar (with the game dumping its rendered bottom-screen framebuffer to the SD card, same mechanism as the reporter's screenshot).

**Opaque white box around the cursor**: `Draw2DImage` blends with premultiplied-alpha factors (`GPU_ONE`, `GPU_ONE_MINUS_SRC_ALPHA`), but the textures carry straight alpha. Instrumentation showed the converted cursor texture is correct in memory — color-keyed texels are palette gray `c3c3c3` with alpha 0 — yet with a source factor of `GPU_ONE` those texels still add their full RGB to the framebuffer, which produces exactly the observed artifact: a light-gray box that saturates to white over the sky, with the background bleeding through additively. Switched to straight-alpha blending; the cursor now renders with a fully transparent background.

**Rainbow colors in the distance**: every world texture was created with `maxLevel = 4`, but citro3d's `C3D_TexGenerateMipmap` cannot generate levels smaller than 8x8 — its inner loops simply don't execute (`dst_height/8 == 0`) and the levels are left as uninitialized `linearAlloc` memory. A 64x64 texture therefore samples garbage in its 4x4 level at high minification, which is why the corruption appears specifically on distant geometry (in the emulator the unwritten levels read back as zeros; on hardware they contain heap garbage, hence the rainbow noise). Verified live: pre-fix, the deepest level of a 64x64 texture was unwritten; post-fix the chain stops at a properly generated 8x8 level and distant geometry renders clean.

**No way to move outside the touch schemes**: #874 turned the circle pad into the cursor stick, and the d-pad only duplicated cursor movement. The d-pad now feeds the joystick navigation path as full-deflection digital axes (the same path the circle pad drove before #874), so d-pad up/down walks and left/right turns, as requested in the issue. Verified end-to-end via a virtual gamepad: holding d-pad up ramps the nav controller's linear velocity and advances the position; holding right produces the full rotational velocity.

### DIFF
--- a/ISLE/isleapp.cpp
+++ b/ISLE/isleapp.cpp
@@ -1730,6 +1730,7 @@ void IsleApp::MoveVirtualMouseViaJoystick()
 	float dpadX = 0.0f;
 	float dpadY = 0.0f;
 
+#ifndef __3DS__
 	if (g_dpadLeft) {
 		dpadX -= m_cursorSensitivity;
 	}
@@ -1742,6 +1743,7 @@ void IsleApp::MoveVirtualMouseViaJoystick()
 	if (g_dpadDown) {
 		dpadY += m_cursorSensitivity;
 	}
+#endif
 
 	// Use joystick axis if non-zero, else fall back to dpad
 	float moveX = ((g_lastJoystickMouseX != 0) ? g_lastJoystickMouseX : dpadX) * frames;
--- a/LEGO1/lego/legoomni/src/input/legoinputmanager.cpp
+++ b/LEGO1/lego/legoomni/src/input/legoinputmanager.cpp
@@ -208,6 +208,28 @@ MxResult LegoInputManager::GetJoystickState(MxU32* p_joystickX, MxU32* p_joystic
 			break;
 		}
 	}
+#else
+	// The circle pad doubles as the cursor stick, so the d-pad is used for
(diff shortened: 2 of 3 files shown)


Rhlp-Engineering:
wow wow wow! yay! thank you!

PR #892, opened 2026-09-09
Point PTATCAM animations at the player in third-person camera

foxtacles:
Animations that carry a `PTATCAM` directive turn the listed ROIs toward the camera every frame. With the third-person camera extension active, the camera orbits the player, so those characters looked past the player at the orbit position instead of at them.

## Change

- `ThirdPersonCameraExt` gains a `HandlePtAtCamTarget(Vector3&)` hook. It returns TRUE and writes the player ROI's world position while the third-person camera is active and has a player ROI, and FALSE otherwise.
- `LegoAnimPresenter::PutFrame` and `LegoLoopingAnimPresenter::PutFrame` resolve the facing target once per frame through the hook and fall back to the camera controller's world location when it declines. The facing math is unchanged.

## Unaffected

Multiplayer reenactments apply their own PTATCAM pass in `ScenePlayer::ApplyPtAtCam` (targeting the spectator participant) and never go through these presenters.

## CI: macOS (Intel) brew step

Homebrew's `llvm` formula moved to 23.1.1 without an x86_64 macOS bottle, so the macOS (Intel) job began compiling LLVM from source in the brew step (two attempts cancelled after 25 and 13 minutes on `cmake --build .` inside the llvm formula; the step took 2 to 4 minutes on every earlier master run). `LLVM_ROOT` is only consumed by the clang-tidy lookup, and no macOS job enables clang-tidy, so the second commit gates the llvm install on the matrix `clang-tidy` flag, the same way `qt6` is already gated.

## Verification

- Web (Emscripten) Release build and native Debug build compile and link cleanly.
- The local target copy was checked to hold its own storage rather than alias the temporary returned by `GetWorldLocation`.

https://claude.ai/code/session_013ED5f7A2zaZxVBimhxXFok

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,7 +100,7 @@ jobs:
         run: |
           brew update
           brew uninstall cmake || true
-          brew install cmake ninja llvm ${{ !matrix.qt-action && 'qt6' || '' }}
+          brew install cmake ninja ${{ matrix.clang-tidy && 'llvm' || '' }} ${{ !matrix.qt-action && 'qt6' || '' }}
           echo "LLVM_ROOT=$(brew --prefix llvm)/bin" >> $GITHUB_ENV
       
       - name: Use latest Xcode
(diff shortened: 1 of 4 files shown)


Rhlp-Engineering:
hey! does the game use delta time?

PR #893, opened 2026-09-10
Merge upstream isle/master (original source layout series)

foxtacles:
Merges `isle/master` 05b8c54f (28 commits, 11e6e2f7..05b8c54f) into portable. Merge base was 31bd20de.

## What comes in from upstream

- The "match the original source layout" series: `tgl/d3drm/*.cpp` + `impl.h` merged into `tglrl40.cpp`/`tglrl40.h`/`tglimpl.h`; `carrace.cpp` + `jetskirace.cpp` into `legorace.cpp`; `legoanim.cpp` split into `legoanimscene.cpp`; `legoedge`/`legoorientededge` into `legoweedge`/`legowegedge`; `viewlod.cpp`/`viewroi.cpp` into `viewmanager.cpp` + `viewlod.inl.h`; `realtime.cpp`/`roi.h` into `realtime.h` + `vectorlength.inl.h`; `mxomnicreateflags/param.cpp` into `mxmain.cpp`; `mxscheduler.cpp` into `mxthread.cpp`; `legoactors.cpp`/`legoplants.cpp`/`legolocations.cpp` tables into their managers; `act2genactor.cpp`, `buildingentity.cpp`, `main.cpp` removed.
- BETA10 annotations and spellings, `LegoActor::e_*` actor constants, struct field renames (`ModelInfo::modelName`, `AnimInfo::animName`/`models`, `AnimState`, `LegoBuildingInfo::entity`, `LegoLocation::Boundary::src`, `ViewROI` token API, `ModelDbWorld::m_partlist`/`m_modarr`, `MXIOINFO` deriving from the `MMIOINFO` layout, `RegistrationBook::m_letters`/`m_cursorPos`, `InfocenterState::m_step`, ...), include-order restoration (`SortIncludes: Never` in `.clang-format`), matrix/vector inline relocation.

## What is deliberately not carried over

Decomp-only artifacts: `reprobit/` + `reprobit.toml`, reccmp CSVs and `cmake/reccmp.cmake`, the `CONFIG/` dialog sources (portable keeps `CONFIG/qt`), `3rdparty/smacker` + `smartheap` libraries, upstream `.github` workflows/actions/scripts, `tools/*.py` + `tools/README.md`, `LegoOmni.def`, the `capstone`/`pefile` additions to `tools/requirements.txt`, upstream README/CONTRIBUTING/.gitignore/.gitattributes wording.

Facilities portable had already dropped stay dropped rather than coming back as guarded copies: DirectSound cache sounds, DirectInput/winmm joystick, registry config, GDI FPS text, `CreateCursorSurface`, `FILE*`-based model DB `Dump`/`Write`/`WriteModelDbParts`, `conio`/`io.h` helpers.

## Portable adjustments re-applied to the moved code

- `tglrl40.h`/`tglimpl.h`/`tglrl40.cpp`: `unsigned int` Tgl signatures, signed `Pick` coordinates, `MINIWIN` include blocks and `LPD3DRM_APPDATA`, `D3DRMMAP_NONE`, `LightSetColor` releases, `GetPosition(NULL, ...)`, no `refCount == 0` asserts in `~ViewportAppData`, `DeviceGetTrianglesDrawn` returns 0 under `MINIWIN` (no `GetStats`/`D3DSTATS`).
- `legorace.cpp`: `SDL_rand`, initialised `presenter`, `EmitGameEvent(e_raceFinished)` for both races, `<stdio.h>`.
- `mxio.h`/`.cpp`: `MXIOINFO : public ISLE_MMIOINFO` with the member prefix dropped like upstream; SDL_IOStream body unchanged.
- `modeldb.cpp` rebuilt on the upstream layout with `SDL_IOStream` reads and the buffer-overrun fix.
- `viewmanager.cpp`: new `FlushBuffers`/`FlushFrameBuffers` bodies under `#ifndef MINIWIN` (D3DRM device/visual arrays are not part of miniwin); `c_tokenUnset`/`GetToken` names with portable's 0.002 radius threshold and user-max-LOD check kept.
- `isleapp.cpp` no longer includes `tglimpl.h`; the miniwin device accessor reads the D3DRM device through the abstract `Tgl::Device::ImplementationDataPtr()`. With the include below in place, GCC emitted the Tgl implementation vtables inside the executable, which then referenced DLL-internal symbols and broke the mingw64 link (lego1 is a DLL there).
- `tglimpl.h` includes `tglrl40.h` at its end. Upstream's `tglimpl.h` defines the `Impl` class methods inline and declares the D3DRM helper functions they call, but those helpers are only defined (inline) in `tglrl40.h`. Clang and MSVC happened to emit the bodies from `tglrl40.cpp`; GCC inlines them away, so every GCC job (Linux, Flatpak, 3DS, Switch, Vita, DOS) failed at link time with hundreds of undefined `TglImpl::*` references. Reproduced and verified with the devkitPro 3DS toolchain in docker.
- `legoanim.h`/`.cpp`: `LegoAnimNodeData::Interpolate` is no longer `inline`. After the split it is called from `legoanimscene.cpp`, and the Release Emscripten link failed with an undefined symbol.
- `legoanimactor.cpp`: upstream's branch order with the `numROIs != 0` guard; the second loop redeclares `i` (upstream relies on MSVC for-scope leakage).
- `mxvideopresenter.cpp`: no 640x480 clamping (portable resolutions); upstream's `assert(r == DD_OK)` after `Blt` is kept and flagged below.
- `legovideomanager.cpp`: upstream's light-support condition written as `dcmColorModel == (D3DCOLORMODEL) 0` (miniwin's `D3DCOLORMODEL` is an enum class).
- `legotesttimer.cpp`: the reconstructed console control (`_kbhit`/`_getch`) under `#ifndef MINIWIN`.
- `CMakeLists.txt` source lists, `lego1_pch.h`, and `extensions/` updated for the removed/added files and the renames (`e_*` actors, `modelName`/`animName`/`models`, `Extra::roi`, `LegoBuildingInfo::entity`, `GetToken`, `m_partlist`/`m_modarr`, `tglimpl.h`).
- `mxdirectdraw.cpp`: `~DeviceModesInfo` without `m_modeArray` (portable enumerates modes through SDL).
- `legoutils.cpp` `WriteDefaultTexture`: the row copy stays `memcpy(bits, surface, ...)`. Upstream flipped it to reproduce the retail binary, where the `lPitch != dwWidth` branch is dead. In portable it is live for texture-loader replacements (32-bpp surfaces, `lPitch = 4 * width`), and the retail direction would write the uninitialised image over the locked texture on every save.
- `isleapp.cpp`: portable's SDL app kept; only upstream's named constants merged.

## Asserts that fire in normal play (flagged, unchanged)

These come from upstream and are left as they are so they can be made BETA10-only in `isledecomp/isle` if desired. Release/NDEBUG builds (the shipped ones) are unaffected; Debug builds abort at the first one.

| Assert | Where | Why it fires |
| --- | --- | --- |
| `assert("Tickle Manager client is not registered!" == NULL)` | `mxticklemanager.cpp` `GetClientTickleInterval` | `RegisterClient` calls it to test whether the client is already registered, so every first registration hits the not-found path (36 times during startup + island tour). |
| `assert(p_numAnims == m_numAnims)` | `legoanimationmanager.cpp` `AnimState::CopyToAnims` | `LoadWorldInfo` calls it right after creating the `AnimState`; `m_numAnims` is still 0 and `m_animSaveInfo` NULL, which the function body already tolerates. |
| `assert(refCount == 0)` | `tglrl40.h` `RendererDestroy` | New in this range. Under miniwin the renderer holds extra references (`QueryInterface` for `IDirect3DRM2`, the never-released `handle`, the second `CreateRenderer` from `Lego3DManager`), so `Release` returns 2 and then 1 at shutdown. |
| `assert(newworld)` | `modeldb.cpp` `ReadModelDbWorlds` | The only caller (`LegoWorldPresenter::LoadWorld`) passes a NULL out-pointer, and the function sets it to NULL two lines later; fires on every world load. |
| `assert(act2State)` / `assert(act3State)` | `infocenter.cpp` `ReadyWorld` | Reasoned, not reproduced: loading a save written while the current act was 2 or 3. `LegoAct2State`/`Act3State` are not serializable, so `LegoGameState::Load` never recreates them, and the retained `if (act2State && ...)` guards show retail expected NULL. |

| `assert(r == DD_OK)` | `mxvideopresenter.cpp` `PutFrame` | Reasoned: miniwin's framebuffer `Blt` returns `DDERR_GENERIC` once the renderer is gone (lost WebGL context, teardown), so this fires only on those abnormal paths, not in normal play. |

The first two were observed with a soft-assert Debug run (asserts turned into log lines) that boots to the island and teleports through 15 island locations; nothing else fired. The `RendererDestroy` one is reached on a clean quit. The review agents' static passes also list plausible-but-unproven ones: `legogamestate.cpp` `SetActor` `assert(p_id != LegoActor::e_none)`, reached from the Infocenter map's own Info Center target when no character has been picked yet (the sibling map cases guard against `e_none`, this one does not); `mxdiskstreamprovider.cpp` "Queueing request after provider stop" on quit ordering; `legoextraactor.cpp` `SwitchDirection` (`assert(m_destEdge && m_boundary)` right before the `if (!m_boundary)` fallback), `legoanimmmpresenter.cpp` `assert(m_roiMapSize)` for an MM presenter whose child resolved no ROI, and asserts that extension SI content can trip (`assert(chunk->GetChunkFlags() & DS_CHUNK_END_OF_STREAM)` in `legoanimpresenter.cpp` for loop-flagged animations, `assert(m_roi)` on `DB_CREATE`, `assert(texture)` in the character manager, `assert(0)` on `CreatePresenter` failure in `mxdsbuffer.cpp`, the `legolod.cpp`/`legoroi.cpp` read-failure and missing-texture `assert(0)` that used to be `assertIfBeta10` and now abort Debug builds for custom models with unregistered textures, `realtime.h` `assert(lods && i < lods->Size())` turning a silent out-of-range LOD read into an abort, `legopathstruct.cpp` `assert(p_data <= 24)` on path-trigger data).

## Independent reviews

Three independent agents reviewed the whole merge result file by file (upstream changes present, portable adjustments preserved, CMake/pch lists, renamed identifiers in non-locally-built code, DX5 non-miniwin branches, for-scope leaks, every upstream-added assert, clang-format and ncc). Their findings were applied: the dropped `assert(newworld)` and `assert(r == DD_OK)` restored (flag, not change), the FPS-overlay clear back to `lPitch` (#308), the `~MxDSBuffer` refcount assert removed like upstream, the `WriteDefaultTexture` copy direction kept, `GetAttenuation` parameter name and range assert taken, `g_target*` defaults on the named constants, the `Advance` declaration order and comment, `.gitattributes` rebuilt on upstream's file plus portable's LFS/SVG rules (batch files and the GXM `.perf.txt` outputs keep their CRLF bytes), `.gitignore` duplicates removed.

A diff-of-diffs check (upstream's base→05b8c54f change against portable's master→merge change, compared line by line per file) leaves only SDL/miniwin replacements, the renames propagated into `extensions/`, `MINIWIN` guards, the tglrl40 type adjustments and the facilities portable had already dropped.

## Verification

- macOS native Debug and Release builds with extensions and the Qt `isle-config` target: clean; the Release binary boots and runs.
- Emscripten Release build (`emcmake`, `ISLE_COMPILE_SHADERS=OFF`, embedded data): clean, `isle.wasm` links.
- Nintendo 3DS build (devkitARM GCC) in docker: clean after the `tglimpl.h` include fix.
- CI-equivalent `clang-format` 17 pass over `CONFIG LEGO1 ISLE miniwin`: no diff.
- Soft-assert Debug run: boots, intro, infocenter, Act 1 island, 15 teleports; only the tickle-manager and `CopyToAnims` asserts fired.
- Not built locally: MSVC with the real DX5 SDK (`ISLE_USE_DX5=ON`, non-`MINIWIN`), mingw, iOS, Switch, Vita, DOS, Android. The first CI run on this branch passed MSVC x86 (DX5), x64, arm64, Xbox One, Android, Emscripten, macOS, FreeBSD, clang-format and the ncc check; only the GCC jobs failed with the link errors fixed above (static-library platforms in the first run, the mingw64 DLL split in the second).
- Not run locally: the `ncc` naming check (needs libclang 16).

https://claude.ai/code/session_013ED5f7A2zaZxVBimhxXFok


CST1229:
~~did you replace yourself with an ai or something?~~

PR #895, opened 2026-09-18
Added always on mouse look (toggleable) and mouse sensitivity.

oliveryasuna:
Thank you for this wonderful project!

I wanted to modernize the third-person extension a bit more:

1. Added an option to have mouse look enabled by default, with a configurable key/button that toggles it.
2. Added options for mouse sensitivity.

I haven't written C++ in like 10 years, and of course this project is new to me, so all feedback is welcome.


oliveryasuna:
It doesn't look like all those workflow failures are due to my changes, are they?

foxtacles:
It looks like upstream iniparser broke. I'll push a fix that pins to a previous version

oliveryasuna:
> It looks like upstream iniparser broke. I'll push a fix that pins to a previous version

Thank you!

